PR #7, opened 2021-10-26
Add app deploy scenario

qiujian16:


mikeshng:
@qiujian16 we have implemented the placement API integration. There might be more changes needed in the future but for now I tried the scenario in this PR and it works fine. Thanks for creating this scenario.

/unhold

/approve

/lgtm

PR #16, opened 2021-12-13
Adopt FOSSA scan.

mikeshng:
Signed-off-by: Mike Ng <ming@redhat.com>


mikeshng:
Looks like we can't add the fossa scan to this project because 
`No analysis targets found in directory.`

PR #71, opened 2022-08-30
add a solution to deploy argocd apps

elgnay:
Signed-off-by: Yang Le <yangle@redhat.com>

### DIFF
--- a/solutions/deploy-argocd-apps/README.md
+++ b/solutions/deploy-argocd-apps/README.md
@@ -1,0 +1,137 @@
+# Deploy applications with Argo CD
+
+The script and instructions provided in this doc help you to setup an Open Cluster Management (OCM) environment with Kind clusters and integrate it with Argo CD. And then you can deploy Argo CD applications to OCM managed clusters.
+
+## Prerequisite
+
+- [kind](https://kind.sigs.k8s.io) must be installed on your local machine. The Kubernetes version must be >= 1.19, see [kind user guide](https://kind.sigs.k8s.io/docs/user/quick-start/#creating-a-cluster) for more details.
+
+- Download and install [clusteradm](https://github.com/open-cluster-management-io/clusteradm/releases). For Linux OS, run the following commands:
+
+    ```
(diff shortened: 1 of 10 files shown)


solutions/deploy-argocd-apps/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+    ```bash
+    export ADMIN_PASS=$(kubectl -n argocd get secret argocd-initial-admin-secret -o=jsonpath='{.data.password}' | base64 -d)
+    argocd login localhost:8080 --username=admin --password="${ADMIN_PASS}" --insecure
+    ```
+
+    Register the two OCM managed clusters to Argo CD
+
+    ```bash
+    argocd cluster add kind-cluster1 --name=cluster1
```
qiujian16, line 77: will we be able to use a script to do this? or use managed-sa?

@mikeshng @xiangjingli when we have the ability to run argo with ManifestWork, this steps will be eased, right?

elgnay: Done. A separated script , `setup-argocd.sh`, is create to integrate Argo CD with OCM.

xiangjingli: @qiujian16   The solution is actually the push model of the argocd integration. Basically there is only one argocd server running on the hub cluster.  The argocd api server is exposed to localhost , then the argocd CLI is able to log into the argocd server, then uses argocd CLI to import the managed cluster secrets to the argocd api server NS, then create a applicationSet to directly deploy resources to the  remote managed clusters. 

In the new pull model of argocd integration we are working on, each managed cluster will run a local argocd  server. user needs to create a argocd application template on hub, and the argocd application template will be delivered to each managed cluster via manifestwork. Then the argocd server running on each managed cluster is triggered to deploy resources locally.  There is no need to import all managed cluster secrets to the ArgoCD api server. and we don't reply on argocd cli either.

With that being said, the push model and pull model don't conflict each other.  I think we can add another new solution to show how to deploy argocd app with pull model once it is available in upstream.

PR #73, opened 2022-09-01
Add Kubefed to Open Cluster Management migration guide

mikeshng:
Signed-off-by: Mike Ng <ming@redhat.com>


qiujian16:
I wonder whether we should have a subcommand in clusteradm to help migrate from kubefed.

mikeshng:
> I wonder whether we should have a subcommand in clusteradm to help migrate from kubefed.

That's a good idea. What are you thinking in terms of what that command does? Is it to help with the federated resources adopt to using manifestwork?

qiujian16:
I think the first step could be we have a command in clusteradm that register all clusters which are joined in kubefed.

mikeshng:
> I think the first step could be we have a command in clusteradm that register all clusters which are joined in kubefed.

Yes, I think that will simplify things. I will give it a try. I think we can also have this guide and then adjust it later once we have the clusteradm changes merged.

PR #90, opened 2022-11-10
add deploy-ocm-controlplane script

ycyaoxdu:
also modified the local start controlplane script

### DIFF
--- a/controlplane/Dockerfile
+++ b/controlplane/Dockerfile
@@ -1,0 +1,22 @@
+FROM golang:1.18 AS builder
+
+ARG OS=linux
+ARG ARCH=amd64
+ENV DIRPATH /go/src/open-cluster-management.io/OCM/controlplane
+WORKDIR ${DIRPATH}
+
+COPY . .
+
+RUN apt-get update && apt-get install net-tools && make vendor 
+RUN GOOS=${OS} \
+    GOARCH=${ARCH} \
+    make build
+
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+ENV USER_UID=10001
+
+COPY --from=builder /go/src/open-cluster-management.io/OCM/controlplane/bin/ocm-controlplane /
+RUN mkdir -p /.ocmconfig
+
+USER ${USER_UID}
--- a/controlplane/Dockerfile
+++ b/controlplane/Dockerfile
@@ -17,6 +17,5 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 ENV USER_UID=10001
 
 COPY --from=builder /go/src/open-cluster-management.io/OCM/controlplane/bin/ocm-controlplane /
-RUN mkdir -p /.ocmconfig
 
 USER ${USER_UID}
--- a/controlplane/hack/deploy/deployment.yaml
+++ b/controlplane/hack/deploy/deployment.yaml
@@ -16,16 +16,35 @@ spec:
     spec:
       containers:
       - name: controlplane
-        image: quay.io/open-cluster-management/controlplane:latest
-        imagePullPolicy: IfNotPresent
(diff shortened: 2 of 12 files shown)


ycyaoxdu:
should add `--use-bootstrap-token` flag while get token

controlplane/hack/deploy/kustomization.yaml
```diff
@@ -39,5 +40,13 @@ kind: Kustomization
...
 - name: controlplane-cert
   files:
-  - kube-serviceaccount.key
-  - client-ca.crt
+  - cert/kube-serviceaccount.key
+  - cert/client-ca.crt
+  - cert/client-ca.key
+  - cert/serving-kube-apiserver.crt
+  - cert/serving-kube-apiserver.key
+  - cert/client-kube-aggregator.crt
```
qiujian16, line 48: we do not need cert/client-kube-aggregator.crt and key, right?

controlplane/hack/deploy/deployment.yaml
```diff
@@ -16,16 +16,35 @@ spec:
     spec:
       containers:
       - name: controlplane
-        image: quay.io/open-cluster-management/controlplane:latest
-        imagePullPolicy: IfNotPresent
+        image: docker.io/yaoyuchen/controlplane:latest
```
qiujian16, line 19: @clyang82 we could later use the repo on quay.io?

qiujian16 (commented):
would you add a readme also?

controlplane/hack/deploy/kustomization.yaml
ycyaoxdu: we need it in [kubeconfig](https://github.com/open-cluster-management-io/OCM/blob/standalone-controlplane/controlplane/hack/lib/util.sh#L302) which is needed while creating comfigmap  cluster-info, maybe should write kubeconfig with crt and key data rather than file first, then remove the files from secret?

controlplane/hack/deploy/ns.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ocm-controlplane
```
zhiweiyin318, line 4: the namespace should be configured by user because we need to deploy multi controlplane in a single OCP.

clyang82: Yes. we do not need to create namespace.

controlplane/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN GOOS=${OS} \
+    GOARCH=${ARCH} \
+    make build
+
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+ENV USER_UID=10001
+
+COPY --from=builder /go/src/open-cluster-management.io/OCM/controlplane/bin/ocm-controlplane /
+RUN mkdir -p /.ocmconfig
```
clyang82, line 20: do we need to create this folder since you have mounted in deployment?

controlplane/hack/deploy/ns.yaml
ycyaoxdu: done

controlplane/hack/deploy/kustomization.yaml
clyang82: We may need to create a separated secret to store kubeconfig. The kubeconfig is consumed by addons

qiujian16:
/approve

looks good in general.

controlplane/hack/deploy/kustomization.yaml
clyang82: As we discussed offline, `kube-aggregator.kubeconfig` will contain ca and token. we create another kubeconfig secret to be used by addons.

PR #92, opened 2022-11-14
Create kubeconfig secret

clyang82:


controlplane/hack/deploy/kustomization.yaml
```diff
@@ -44,8 +41,9 @@ secretGenerator:
   - cert/client-ca.key
   - cert/serving-kube-apiserver.crt
   - cert/serving-kube-apiserver.key
-  - cert/client-kube-aggregator.crt
-  - cert/client-kube-aggregator.key
   - cert/kube-aggregator.kubeconfig
+- name: controlplane-kubeconfig
```
qiujian16, line 45: this may have issue when generating cluster-info cm?

ycyaoxdu: looks like the cert/kube-aggregator.kubeconfig stiil exists in secret controlplane-cert @qiujian16

clyang82: yes. cert/kube-aggregator.kubeconfig is still using for generating cluster-info cm.

qiujian16: but this one is not useful because it has no related cert/key, right?

PR #98, opened 2022-11-15
fix some deploy issues and update registration pkg version to latest

zhiweiyin318:
Signed-off-by: Zhiwei Yin <zyin@redhat.com>
1. fix the work-manager addon miss agent clusterrole.
2. fix the readme typo.
3. fix the demo script issue. 
4. update the registration pkg to the latest which includes some fixes.


controlplane/hack/demo/demo-magic.sh
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env bash
```
clyang82, line 1: for third-party lib, it is better to reference it

zhiweiyin318: ok, removed.

PR #104, opened 2022-11-17
Fix wrong type error in admission

qiujian16:


morvencao (commented):
/LGTM

controlplane/hack/lib/util.sh
```diff
@@ -283,13 +283,13 @@ EOF
 function kube::util::write_client_kubeconfig {
     local sudo=$1
     local dest_dir=$2
-    local ca=`cat $3 | base64 -w0`
+    local ca=`cat ${dest_dir}/$3 | base64`
```
morvencao, line 286: is this intent? removing `-w0`?

clyang82, line 286: why remove `-w0`? I think we need to disable the line wrapping.

PR #107, opened 2022-11-17
Disable openapiV3 since it will print error

qiujian16:


controlplane/hack/deploy/controlplane/deployment.yaml
```diff
@@ -38,7 +38,7 @@ spec:
           - "--secure-port=9443"
           - "--tls-cert-file=/controlplane/cert/serving-kube-apiserver.crt"
           - "--tls-private-key-file=/controlplane/cert/serving-kube-apiserver.key"
-          - "--feature-gates=DefaultClusterSet=true"
+          - "--feature-gates=DefaultClusterSet=true,OpenAPIV3=false"
```
clyang82, line 41: there is a duplicated one in line 24.

PR #147, opened 2023-05-30
Support building images with podman

zhujian7:
With this PR, we can use podman to build images by:
```
RUNTIME=podman make images
```


xuezhaojun:
Had another PR fix this: https://github.com/open-cluster-management-io/ocm/pull/154

PR #148, opened 2023-05-30
Refactor version/feature/cmd packages

qiujian16:


xuezhaojun:
The e2e of registration-operator is failing, but it's seems like a flaky error.

PR #149, opened 2023-05-30
client certificate expiration seconds must greater or qual to 3600

youhangwang:
In PR: https://github.com/open-cluster-management-io/registration/pull/312, a option `--client-cert-expiration-seconds` was exposed so that a user can specify a expiration seconds for the client cert. And we use the minimum value of expiration time in CSR - 10min to check if the expiration time is validated.

However, for the clientCertificateController in registration, 600 is too short:

1. clientCertificateController to check if cert is expired every 5min.
2. kubeclient resync the client cert every 5min.
3. clientCertificateController will refresh the cert if the cert has less than a random percentage range from 20% to 25% of its life remaining.

so the minimum expiration second would be (5 * 60)/(1/5) = 1500s for clientCertificateController. otherwise, the kube client will use a expired client cert to send request to the remote cluster.

Considering to add some margin. 1h(3600s) could be a good choice.


qiujian16:
hrm, seems there is some issue with operator e2e...I am looking at it.

qiujian16:
@youhangwang would you rebase the PR, there was a bug in the code that is fixed in https://github.com/open-cluster-management-io/ocm/pull/150

youhangwang:
@qiujian16 branch was rebased on the main branch

PR #150, opened 2023-05-30
Refactore integration test

qiujian16:


xuezhaojun:
Still have the flaky integration test on work component, it also happened in the origin work repo, should be fixed later.

PR #156, opened 2023-05-31
Move operator e2e

qiujian16:


xuezhaojun:
The e2e-work fail is because download kustiomize error, but the command works fine on my local environment, guess it's because temp network error.

Issue #162, opened 2023-06-05
[It] Should schedule placement successfully once spec.Predicates LabelSelector changes is flaky


qiujian16:
**Describe the bug**
https://github.com/open-cluster-management-io/ocm/actions/runs/5154964640/jobs/9284041470?pr=161

**To Reproduce**
run the integration test

/assign @haoqing0110

PR #165, opened 2023-06-05
stabilize placement integration test

haoqing0110:
Fix: https://github.com/open-cluster-management-io/ocm/issues/162

### DIFF
--- a/test/integration/placement/assertion_test.go
+++ b/test/integration/placement/assertion_test.go
@@ -3,9 +3,10 @@ package placement
 import (
 	"context"
 	"fmt"
-	"open-cluster-management.io/ocm/test/integration/util"
 	"sort"
 	"time"
+
+	"open-cluster-management.io/ocm/test/integration/util"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


test/integration/placement/assertion_test.go
```diff
@@ -410,16 +411,22 @@ func assertCreatingPlacementWithDecision(name, namespace string, noc *int32, nod
...
 	gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
 	placement.Spec.NumberOfClusters = noc
-	placement.Spec.Tolerations = tolerations
+	placement.Spec.Predicates = predicate
 	placement.Spec.PrioritizerPolicy = prioritizerPolicy
-	placement, err = clusterClient.ClusterV1beta1().Placements(namespace).Update(context.Background(), placement, metav1.UpdateOptions{})
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	placement.Spec.Tolerations = tolerations
+	gomega.Eventually(func() error {
```
qiujian16, line 423: I think you need to put get in the eventually also

qiujian16 (commented):
@haoqing0110 I am fine to merge this to fix the flaky issue, but we should consider using patch. the assertUpdatingPlacement cannot update partial of the spec which may cause some complexity in the future.

/approve
/lgtm

PR #177, opened 2023-06-08
:seedling: add a verify rule for golang files import order

zhujian7:
## Summary

This PR uses the [gci tool](https://github.com/daixiang0/gci) to make all go files' import section with a specific order, it will organize import with group with order：
1. standard library modules
2. 3rd party modules
3. modules in OCM org, like the `open-cluster-management.io/api`
4. current project `open-cluster-management.io/ocm` modules

developers can use the `make format-gofiles-import` to format the import automatically and the `make verify-gofiles-import-format` to check for any violation.

## Related issue(s)

Fixes #


xuezhaojun:
Do this mean developers need to depend on [gci](https://github.com/daixiang0/gci) in the future? And what's the relationship between gci and [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports)?

zhujian7:
> Maybe we can import this indirectly by depending on golangci-lint https://golangci-lint.run/usage/linters/#gci, it's like a more commonly used linter tool integrated with more features including gci.

Thanks @xuezhaojun yes, I have considered using golangci-lint, but once enable the lint check, there are so many errors that need to fix, which makes the PR hard to review, so I will add the lint check in another PR.

Makefile
```diff
@@ -68,6 +68,21 @@ verify-gocilint:
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.45.2
 	golangci-lint run --timeout=3m --modules-download-mode vendor ./...
 
+install-golang-gci:
+	go install github.com/daixiang0/gci@v0.10.1
+
+format-gofiles-import: install-golang-gci
```
qiujian16, line 74: the target name could be simpler like `imports` or `fmt-imports`

Makefile
```diff
@@ -68,6 +68,21 @@ verify-gocilint:
...
+
+verify-gofiles-import-format: install-golang-gci
+	@output=$$(gci diff --skip-generated -s standard -s default -s "prefix(open-cluster-management.io)" -s "prefix(open-cluster-management.io/ocm)" cmd pkg test dependencymagnet); \
+	if [ -n "$$output" ]; then \
+	    echo "Diff output is not empty: $$output"; \
+	    exit 1; \
+	else \
+	    echo "Diff output is empty"; \
+	fi
+
```
qiujian16, line 85: this could be put under verify target, so you do not need create another action step

zhujian7: after this PR is merged, I will integrate the verification into golangci-lint:  https://golangci-lint.run/usage/linters/#gci by a follow up PR

PR #185, opened 2023-06-14
:bug: filter secret informer to reduce cache size

zhiweiyin318:
## Summary
the klusterlet watches the secrets and deployments. when there are large numbers of secrets  or deployments in the managed cluster, the pod of klusterlet will be OOM because the cache size of secret/deployment informer is increased beyond the limit. So this PR is to filter the secret and deployment informer to reduce the case size.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go
+++ b/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go
@@ -46,7 +46,7 @@ func NewKlusterletCleanupController(
 	apiExtensionClient apiextensionsclient.Interface,
 	klusterletClient operatorv1client.KlusterletInterface,
 	klusterletInformer operatorinformer.KlusterletInformer,
-	secretInformer coreinformer.SecretInformer,
+	secretInformers map[string]coreinformer.SecretInformer,
 	deploymentInformer appsinformer.DeploymentInformer,
 	appliedManifestWorkClient workv1client.AppliedManifestWorkInterface,
 	kubeVersion *version.Version,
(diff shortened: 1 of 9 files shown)


zhiweiyin318:
> we can reduce the permission, right?

@qiujian16  I am afraid not, the registration-agent also need secret permission to create addon hub-kubeconfig secrets which have different names.

pkg/operator/operators/klusterlet/options.go
```diff
@@ -52,6 +56,24 @@ func (o *Options) RunKlusterletOperator(ctx context.Context, controllerContext *
 
 	kubeInformer := informers.NewSharedInformerFactory(kubeClient, 5*time.Minute)
 
+	// to reduce cache size if there are larges number of secrets
+	newOnTermInformer := func(name string) informers.SharedInformerFactory {
+		return informers.NewSharedInformerFactoryWithOptions(kubeClient, 5*time.Minute,
+			informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+				options.FieldSelector = fields.OneTermEqualSelector("metadata.name", name).String()
```
zhujian7, line 63: Is there any expression of the selector to make it possible to be logic OR for the `metadata.name`?

pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go
```diff
@@ -62,7 +60,6 @@ func NewKlusterletCleanupController(
 	}
 
 	return factory.New().WithSync(controller.sync).
-		WithInformersQueueKeyFunc(helpers.KlusterletSecretQueueKeyFunc(controller.klusterletLister), secretInformer.Informer()).
```
zhujian7, line 65: in hosted mode, we need the external-managed-kubecnfig to build a client to clean up resources on the managed cluster. so I think it is necessary to watch this secret to trigger a cleanup reconcile.

zhiweiyin318: added back.

pkg/operator/operators/klusterlet/options.go
zhiweiyin318: have not found that expression.

pkg/operator/operators/klusterlet/options.go
```diff
@@ -52,6 +56,29 @@ func (o *Options) RunKlusterletOperator(ctx context.Context, controllerContext *
 
 	kubeInformer := informers.NewSharedInformerFactory(kubeClient, 5*time.Minute)
 
+	// to reduce cache size if there are larges number of secrets
+	newOnTermInformer := func(name string) informers.SharedInformerFactory {
```
zhujian7, line 60:
```suggestion
	newOneTermInformer := func(name string) informers.SharedInformerFactory {
```

zhiweiyin318: done

zhujian7:
LGTM

PR #193, opened 2023-06-21
 🌱 Refactor cluster permission

qiujian16:
## Summary

This is to use lister for permission to avoid too frequent get calls. 
For all legacy k8s resource managed by registration, we add a cluster label and informer will use label selector to filter.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/common/apply/rbac.go
+++ b/pkg/common/apply/rbac.go
@@ -3,6 +3,7 @@ package apply
 import (
 	"context"
 	"fmt"
+
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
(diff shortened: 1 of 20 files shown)


skeeey:
LGTM

pkg/common/apply/rbac.go
```diff
@@ -0,0 +1,147 @@
...
+			result.Result, result.Changed, result.Error = Apply[*rbacv1.ClusterRole](
+				ctx, m.clusterRoleLister, m.client.RbacV1().ClusterRoles(), compareClusterRole, t, recorder)
+		case *rbacv1.ClusterRoleBinding:
+			result.Result, result.Changed, result.Error = Apply[*rbacv1.ClusterRoleBinding](
+				ctx, m.clusterRoleBindingLister, m.client.RbacV1().ClusterRoleBindings(), compareClusterRoleBinding, t, recorder)
+		case *rbacv1.Role:
+			result.Result, result.Changed, result.Error = Apply[*rbacv1.Role](
+				ctx, m.roleLister.Roles(t.Namespace), m.client.RbacV1().Roles(t.Namespace), compareRole, t, recorder)
+		case *rbacv1.RoleBinding:
+			result.Result, result.Changed, result.Error = Apply[*rbacv1.RoleBinding](
```
zhujian7, line 72: if the "roleRef" field changes, the apply will fail, right?

qiujian16: maybe not for now. it is implicit

Issue #128, opened 2023-03-01
Consolidate code of core component into one repo


qiujian16:
We currently have different component repos:
- registration
- work
- placement
- registration-operator
- clusteradm
- addon-framework

It is hard to do e2e tests with many components and hard to do dependency check. I propose,we consolidate the code in these repos to one single repo:

- code in the work/registration/placement/operator/clusteradm component is merged into https://github.com/open-cluster-management-io/OCM
- split addon-framework to pure library code and addon-manager code. The addon-manager code is also put into https://github.com/open-cluster-management-io/OCM
- archive the the independent repo

After that we will have
- https://github.com/open-cluster-management-io/api
- https://github.com/open-cluster-management-io/OCM
- https://github.com/open-cluster-management-io/addon-framework (just library, remove unnecessary deps)

With this consolidation, we will have the benefit that
1. we do not need webhook per component, we can have one single webhook server
2. we can build hub and agent as one single binary
3. easier e2e testing
4. better understanding and dev experience for the contributor

note from @deads2k we should validate gates for PR that we do not have cross deps among different components.

clyang82:
Can we rename `https://github.com/open-cluster-management-io/OCM` to `https://github.com/open-cluster-management-io/ocm`? It seems the lower-case repo name is a common practices.

xuezhaojun:
https://github.com/xuezhaojun/ocm-core: An example merged repo, keeping the same commit history.

xuezhaojun:
> I'm on board with consolidating the repos, but I'm not sure having `clusteradm` in the same repo with the controllers makes sense. It seems like the CLI should have its own repo since it's not a controller/API?

But it seems including CLI in the main repo is a common pattern, is there particular advantages we can gain from not include it?

dhaiducek:
I wasn't aware including the CLI in the main repo was a common pattern (`kubectl` was the only one I checked). I was thinking having it separate might make it more discoverable and would keep the `ocm` repo focused on controllers.

xuezhaojun:
I checked a few repos like [argo-cd](https://github.com/argoproj/argo-cd/blob/5bdd42ceed96de92667262cf066aafffa2891be6/util/cli/cli.go), [istio](https://github.com/istio/istio/tree/master/istioctl) and one similar usage repo [karmada](https://github.com/karmada-io/karmada/blob/master/cmd/karmadactl/karmadactl.go). 

[...]

PR #196, opened 2023-06-25
🌱 Move addon manager from addon-framework to ocm repo

zhujian7:
## Summary
Move addon manager from repo [addon-framework](https://github.com/open-cluster-management-io/addon-framework/tree/0056b148e0a0621c716f1ceffaa034e0f5caaaf0/pkg/manager) to ocm repo.

- [x] move go files
- [x] move integration test
- [ ] move e2e tests: a bit complicated and will be tracked in a follow-up PR

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/128


pkg/addon/controllers/addonmanagement/controller.go
```diff
@@ -0,0 +1,108 @@
...
+				addonClient:                addonClient,
+				placementDecisionLister:    placementDecisionInformer.Lister(),
+				placementLister:            placementInformer.Lister(),
+				managedClusterAddonIndexer: addonInformers.Informer().GetIndexer(),
+				addonFilterFunc:            addonFilterFunc,
+			},
+		},
+	}
+
+	return factory.New().WithInformersQueueKeysFunc(
```
qiujian16, line 65: I think we can use library-go factory for all these controllers

qiujian16:
@zhujian7  also need to stop the gh action on addon-framework repo at first.

zhujian7:
@qiujian16 PTAL

PR #197, opened 2023-06-27
🌱 Refactor to use common queue/filter funcs

qiujian16:
## Summary
put common queue key/filter function in a single place, so we could have a better test coverage.

## Related issue(s)

Fixes #


pkg/common/queue/queuekey.go
```diff
@@ -0,0 +1,63 @@
...
+	return []string{accessor.GetName()}
+}
+
+func QueueKeyByMetaNamespace(obj runtime.Object) []string {
+	accessor, _ := meta.Accessor(obj)
+	return []string{accessor.GetNamespace()}
+}
+
+func QueueKeyByMetaNamespaceName(obj runtime.Object) []string {
+	key, _ := cache.MetaNamespaceKeyFunc(obj)
```
zhujian7, line 61: not sure if it is better to use `DeletionHandlingMetaNamespaceKeyFunc` here?

pkg/operator/operators/klusterlet/controllers/addonsecretcontroller/controller.go
```diff
@@ -40,14 +40,13 @@ func NewAddonPullImageSecretController(kubeClient kubernetes.Interface, operator
...
-		namespace := o.(*corev1.Namespace)
-		return namespace.GetName()
-	}, func(obj interface{}) bool {
-		// if obj has the label, return true
-		namespace := obj.(*corev1.Namespace)
-		return namespace.Labels[addonInstallNamespaceLabelKey] == "true"
-	}, namespaceInformer.Informer()).WithSync(ac.sync).ToController("AddonPullImageSecretController", recorder)
+	return factory.New().WithFilteredEventsInformersQueueKeysFunc(
+		queue.QueueKeyByMetaName,
+		func(obj interface{}) bool {
```
zhujian7, line 45: should we add a common func like `FileterByLabelKeyValue`?

pkg/placement/controllers/scheduling/scheduling_controller.go
```diff
@@ -152,26 +153,17 @@ func NewSchedulingController(
...
-			return key
-		}, placementInformer.Informer()).
+		WithInformersQueueKeysFunc(
+			queue.QueueKeyByMetaNamespaceName,
+			placementInformer.Informer()).
 		WithFilteredEventsInformersQueueKeyFunc(func(obj runtime.Object) string {
 			accessor, _ := meta.Accessor(obj)
 			labels := accessor.GetLabels()
-			placementName := labels[placementLabel]
+			placementName := labels[clusterapiv1beta1.PlacementLabel]
```
zhujian7, line 162: delete the const `placementLabel` in Line 46?

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_index.go
```diff
@@ -71,7 +72,11 @@ func (m *ManifestWorkReplicaSetController) manifestWorkQueueKeyFunc(obj runtime.
 	if !ok {
 		return ""
 	}
-	return key
+	keys := strings.Split(key, ".")
```
zhujian7, line 75: why do we need this?

pkg/common/queue/queuekey.go
qiujian16: this is handled by library-go already.

zhujian7:
/lgtm
/hold
just some nit comments, @qiujian16 feel free to unhold

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_index.go
qiujian16: the manifestwork has a reference to manifestworkreplicaset by a label value of <namespace>.<name>
we cannot use "/" here since it is invalid for label value.

PR #205, opened 2023-06-29
:bug: make the custom sign ca secret namespace fixed

zhujian7:
## Summary

- Restrict the custom sign ca secret must be in the same namespace as addon-manager, and the user needs to be responsible for granting addon manager permission to get the secrt
- Add an e2e case for the addon manager to sign a ca with a customized ca secret.

## Related issue(s)

Fixes #


pkg/addon/templateagent/registration.go
```diff
@@ -25,13 +27,29 @@ import (
...
 	// AddonTemplateLabelKey is the label key to set addon template name. It is to set the resources on the hub relating
 	// to an addon template
 	AddonTemplateLabelKey = "open-cluster-management.io/addon-template-name"
 )
 
+var (
+	podNamespace = ""
+)
+
+func AddonManagerNamespace() string {
```
qiujian16, line 39: I think this should be able to be passed from outside? The controllerCtx have this namespace

pkg/addon/templateagent/registration.go
```diff
@@ -25,13 +27,29 @@ import (
 )
 
 const (
-	TLSCACert = "ca.crt"
-	TLSCAKey  = "ca.key"
 	// AddonTemplateLabelKey is the label key to set addon template name. It is to set the resources on the hub relating
 	// to an addon template
 	AddonTemplateLabelKey = "open-cluster-management.io/addon-template-name"
```
qiujian16, line 32: TODO move this to the api repo.

test/e2e/manifests/addon/signca_secret_role.yaml
```diff
@@ -0,0 +1,16 @@
...
+- apiGroups:
+    - ""
+  resources:
+    - secrets
+  verbs:
+    - get
+    # - list
+    # - watch
+  resourceNames:
+    - << CustomSignerSecretName >>
```
qiujian16, line 16: eof

test/e2e/manifests/addon/signca_secret_rolebinding.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: get-customer-ca
+  namespace: << AddonManagerNamespace >>
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: get-customer-ca
+subjects:
+- kind: ServiceAccount
+  name: addon-manager-controller-sa
+  namespace: << AddonManagerNamespace >>
```
qiujian16, line 14: eof

deploy/cluster-manager/config/rbac/cluster_role.yaml
```diff
@@ -20,10 +20,9 @@ rules:
     - "placement-controller-sa-kubeconfig"
     - "work-controller-sa-kubeconfig"
     - "external-hub-kubeconfig"
-# addon manager needs this to sign the customized type csr
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["create", "get"]
+  verbs: ["create"]
```
qiujian16, line 25: we do not need create I think?

deploy/cluster-manager/config/rbac/cluster_role.yaml
```diff
@@ -9,7 +9,7 @@ rules:
   verbs: ["create", "get", "list", "update", "watch", "patch", "delete", "deletecollection"]
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["list", "watch", "update", "patch", "delete"]
+  verbs: ["get", "list", "watch", "update", "patch", "delete"]
```
qiujian16, line 12: get is not needed, right?

pkg/addon/templateagent/registration.go
zhujian7: This will be used by the `TemplateCSRSignFunc` which does have the controllerCtx . And I think using downward api to get the pod namespace is easier.

pkg/addon/templateagent/registration.go
zhujian7: Because we restricted the sign ca secret type to `kubernetes.io/tls`, these are not needed.

deploy/cluster-manager/config/rbac/cluster_role.yaml
zhujian7: the cluster role here is configured for cluster-manager, I think get for the following secrets is still needed, I just changed it back to several days ago.

deploy/cluster-manager/config/rbac/cluster_role.yaml
zhiweiyin318: we need create secret if clustermanager is in hosted mode. https://github.com/open-cluster-management-io/ocm/blob/9cae24c7fe2df16f7d796fa408e91b80facb0b3e/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go#L304

PR #209, opened 2023-07-03
:sparkles: Trigger clusteradm e2e after post images.

xuezhaojun:
## Related PR
https://github.com/open-cluster-management-io/clusteradm/pull/352

### DIFF
--- a/.github/workflows/post.yml
+++ b/.github/workflows/post.yml
@@ -38,7 +38,6 @@ jobs:
           name: unit
           verbose: true
           fail_ci_if_error: true
-
   images:
     name: images
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/post.yml
```diff
@@ -133,3 +132,13 @@ jobs:
...
+  trigger-clusteradm-e2e:
+    needs: [ images, image-manifest ]
+    name: trigger clusteradm e2e
+    runs-on: ubuntu-latest
+    steps:
+      - uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.OCM_BOT_PAT }}
+          repository: open-cluster-management-io/clusteradm
+          event-type: ocm_changes
```
zhujian7, line 144: can we also send the commit sha, so clusteradm will know which commit triggers?

xuezhaojun: Sure, the sha and ref added.

.github/workflows/post.yml
```diff
@@ -133,3 +132,14 @@ jobs:
...
           docker manifest push quay.io/open-cluster-management/placement:latest
           docker manifest push quay.io/open-cluster-management/addon-manager:latest
+  trigger-clusteradm-e2e:
+    needs: [ images, image-manifest ]
+    name: trigger clusteradm e2e
+    runs-on: ubuntu-latest
+    steps:
+      - uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.OCM_BOT_PAT }}
```
xuezhaojun, line 142: Hi, @mikeshng does OCM_BOT_PAT has write permission on `Contents`? The permission is needed to send a repository-dispatch event.
image

PR #210, opened 2023-07-04
🌱 Reduce e2e process time

qiujian16:
## Summary

for work and addon e2e, we do not need to
restart klusterlet for each case

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/addonmanagement_test.go
+++ b/test/e2e/addonmanagement_test.go
@@ -44,9 +44,9 @@ var (
 	}
 )
 
-var _ = ginkgo.Describe("Enable addon management feature gate", ginkgo.Label("addon-manager"), func() {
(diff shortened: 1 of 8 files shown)


test/e2e/addonmanagement_test.go
```diff
@@ -44,9 +44,9 @@ var (
 	}
 )
 
-var _ = ginkgo.Describe("Enable addon management feature gate", ginkgo.Label("addon-manager"), func() {
+var _ = ginkgo.Describe("Enable addon management feature gate", ginkgo.Ordered, ginkgo.Label("addon-manager"), func() {
```
zhujian7, line 47: why do we need the `ginkgo.Ordered` for this test?

qiujian16: since we use the same addon name in each test case, so the two cannot run concurrently. It also help us to add a BeforeAll and AfterAll to enable/disable the feature gates, so we do not need to enable/disable for each case.

PR #212, opened 2023-07-05
:seedling: use DeploymentAvailability to check addon health

zhujian7:
## Related issue(s)

Fixes #


zhujian7:
@qiujian16 @zhiweiyin318

qiujian16:
Could we add some unit test to try not reduce coverage?

zhujian7:
@qiujian16 PTAL

PR #214, opened 2023-07-06
🐛 remove klusterlet finalizer forcely

zhujian7:
## Summary
Now when we delete a klusterlet CR, the klusterlet-operator will remove the `klusterlet.spec.namespace` namespace, before removing the finalize `operator.open-cluster-management.io/klusterlet-cleanup` for the klusterlet CR.
so there is a case, if the klusterlet-operator is running in the `klusterlet.spec.namespace` namespace, removing the namespace will trigger deleting the klusterlet-operator itself, and if the klusterlet-operator is deleted before the finalize is removed(eg, remove finalize got conflict), the klusterlet will be left and can not be deleted automatically.

this PR is trying to resolve the conflict error when removing the finalizer.

BTW, this PR will fix the warning log:
```
W0710 06:36:59.764545       1 warnings.go:70] unknown field "metadata.UID"
```
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/common/patcher/patcher.go
+++ b/pkg/common/patcher/patcher.go
@@ -24,12 +24,15 @@ type PatchClient[R runtime.Object] interface {
 
 type Patcher[R runtime.Object, Sp any, St any] interface {
 	AddFinalizer(context.Context, R, ...string) (bool, error)
-	RemoveFinalizer(context.Context, R, ...string) error
-	// RemoveFinalizerForcely removes finalizers forcely without checking resource version.
-	RemoveFinalizerForcely(context.Context, R, ...string) error
+	RemoveFinalizer(context.Context, R, PatchOptions, ...string) error
 	PatchStatus(context.Context, R, St, St) (bool, error)
 	PatchSpec(context.Context, R, Sp, Sp) (bool, error)
 	PatchLabelAnnotations(context.Context, R, metav1.ObjectMeta, metav1.ObjectMeta) (bool, error)
+}
+
+type PatchOptions struct {
+	// IgnoreResourceVersion will ignore the resource version matching when patching.
+	IgnoreResourceVersion bool
 }
 
 // Resource is a generic wrapper around resources so we can generate patches.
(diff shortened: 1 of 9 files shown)


pkg/common/patcher/patcher.go
```diff
@@ -131,6 +140,16 @@ func (p *patcher[R, Sp, St]) RemoveFinalizer(ctx context.Context, object R, fina
 		if errors.IsNotFound(err) {
 			return nil
 		}
+
+		// If it's a conflict, we need to get the latest version of the object and try again.
+		if errors.IsConflict(err) {
```
qiujian16, line 145: it seems too greedy, you have to set a backoff time. Also instead of jsonMergePatch, should we use jsonPatch here?

zhujian7: but setting a backoff time will make the retry slow, which is not good for deleting the klusterlet CR.
can we set a retry times(eg.3 times) into the ctx, and if it exceeds, we stop?

zhujian7:
@qiujian16 PTAL

pkg/common/patcher/patcher.go
```diff
@@ -25,6 +25,8 @@ type PatchClient[R runtime.Object] interface {
 type Patcher[R runtime.Object, Sp any, St any] interface {
 	AddFinalizer(context.Context, R, ...string) (bool, error)
 	RemoveFinalizer(context.Context, R, ...string) error
```
qiujian16, line 27: actually, could we add opts var in the RemoveFinalizer func

zhujian7: you mean the patch option for the `p.client.Patch(ctx, accessor.GetName(), types.MergePatchType, patchBytes, metav1.PatchOptions{})`？ 

we will not set `force` to `true` for the patch option, it is used for `re-acquire conflicting fields owned by other people.`
in our case, we just need to keep the `resourceVersion` empty.

qiujian16: not necessary to be a mevat.PatchOptions, but a options field that you can know whether it is force patch or not.

zhujian7: added a PatchOptions param for the RemoveFinalizers func.

pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go
```diff
@@ -191,7 +191,7 @@ func (n *klusterletCleanupController) sync(ctx context.Context, controllerContex
 		return utilerrors.NewAggregate(errs)
 	}
 
-	return n.patcher.RemoveFinalizerForcely(ctx, klusterlet, klusterletFinalizer, klusterletHostedFinalizer)
+	return n.patcher.RemoveFinalizer(ctx, klusterlet, patcher.PatchOptions{IgnoreResourceVersion: true}, klusterletFinalizer, klusterletHostedFinalizer)
```
qiujian16, line 194: will it be better if it is like

n.patcher.WithOptions(opts).RemoveFinalizer? ....

zhujian7: Does this work for all funcs? Or only for RemoveFinalizers?

qiujian16: yes, I think so

zhujian7: fixed

zhujian7:
@qiujian16 PTAL

PR #218, opened 2023-07-13
🌱 Refactor code to fix lint warning

qiujian16:
## Summary
no code behavior change, just to fix lint warnings.
1. handle unhandled error
2. message format in error msg
3. empty slice declaration

## Related issue(s)

Fixes #


zhujian7:
I think we should enable these check rules/linters in the .golangci.yaml

PR #222, opened 2023-07-14
:warning: enable addon management feature gate by default

zhujian7:
## Summary
Enable the addon management feature gate by default for:
- clusterManager, will install the addon-manager component, if you want disable:
```yaml
apiVersion: operator.open-cluster-management.io/v1
kind: ClusterManager
metadata:
  name: cluster-manager
spec:
...
  addOnManagerConfiguration:
    featureGates:
    - feature: AddonManagement
      mode: Disable
```
- klusterlet, will enable the addon management feature for the registration-agent, if you want to disable:
```yaml
apiVersion: operator.open-cluster-management.io/v1
kind: Klusterlet
metadata:
  name: klusterlet
spec:
...
  registrationConfiguration:
    featureGates:
    - feature: AddonManagement
      mode: Disable
```

## Related issue(s)

Fixes #

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	k8s.io/kube-aggregator v0.27.2
 	k8s.io/utils v0.0.0-20230313181309-38a27ef9d749
 	open-cluster-management.io/addon-framework v0.7.1-0.20230705031704-6a328fa5cd63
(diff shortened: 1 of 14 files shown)


qiujian16:
I think we also enable this feature gate manually in some tests, should we also remove those codes?

PR #229, opened 2023-07-24
:bug: refactor gc controller

zhiweiyin318:
## Summary
refactor rbacfinalizerdeletion  to gccontroller, and delete work clustrerole after there is no cluster and manifestwork.

## Related issue(s)

Fixes #


pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,80 @@
...
+
+	informerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	worklister "open-cluster-management.io/api/client/work/listers/work/v1"
+
+	"open-cluster-management.io/ocm/pkg/common/queue"
+)
+
+// gcReconcile is an interface for reconcile cleanup logic.
+type gcReconcile interface {
+	reconcile(ctx context.Context, clusterName string) (bool, error)
```
qiujian16, line 23: we had a change to refactor the gc part. I would think the return should also have a reconcileStop/Continue, so we can know further gc is needed. In the future I think we could build a reconcile chain with metaObjectClient.

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,108 @@
...
+
+	informerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterv1listers "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	worklister "open-cluster-management.io/api/client/work/listers/work/v1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+type gcReconcileOp string
+
+const (
```
qiujian16, line 26: use enum

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+	}
+	return factory.New().
+		WithInformersQueueKeysFunc(func(obj runtime.Object) []string {
+			accessor, _ := meta.Accessor(obj)
+			if accessor.GetDeletionTimestamp() != nil && !accessor.GetDeletionTimestamp().IsZero() {
+				return []string{accessor.GetName()}
+			}
+			return []string{}
+		}, clusterInformer.Informer(), namespaceInformer.Informer()).
```
qiujian16, line 67: why namespaceInformer here?

zhiweiyin318: there is a case that the cluster ns is deleted by user, the work rolebinding in the cluster ns cannot be deleted since there is a finalizer on it. so need to remove the finalizer to make the cluster ns deleted and the cluster ns will be re-created. 
I think we can add a TODO here , will remove namespaceInformer here and to prohibit deleting cluster ns in registration webhook

qiujian16:
need some more ut for coverage.

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,110 @@
...
+			newGCRoleBindingController(roleBindingLister, namespaceInformer, manifestWorkLister, rbacClient, eventRecorder),
+			newGCClusterRoleController(clusterRoleLister, clusterInformer, manifestWorkLister, rbacClient, eventRecorder),
+		},
+	}
+
+	// TODO: remove namespaceInformer and to prohibit delete cluster ns if the cluster is not deleting.
+	return factory.New().
+		WithInformersQueueKeysFunc(func(obj runtime.Object) []string {
+			accessor, _ := meta.Accessor(obj)
+			if accessor.GetDeletionTimestamp() != nil && !accessor.GetDeletionTimestamp().IsZero() {
```
qiujian16, line 65: commonly we check the deleting in reconcile rather than in queuekey func

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,110 @@
...
+				return []string{accessor.GetName()}
+			}
+			return []string{}
+		}, clusterInformer.Informer(), namespaceInformer.Informer()).
+		WithSync(controller.sync).ToController("GCController", eventRecorder)
+}
+
+func (r *GCController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	key := controllerContext.QueueKey()
+	if key == "" {
```
qiujian16, line 75: I remember there is also a default queuekey

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil
+	}
+
+	_, clusterName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return nil
+	}
+
+	cluster, err := r.clusterLister.Get(clusterName)
+	if err != nil && !errors.IsNotFound(err) {
```
qiujian16, line 85: should we remove finalizer on the cluster as the last step? if it is the gc controller of cluster.

pkg/registration/hub/gc/controller.go
zhiweiyin318: factory.DefaultQueueKey is "key". 
replaced with `if key == "" || key == factory.DefaultQueueKey`

pkg/registration/hub/gc/controller.go
zhiweiyin318: moved to reconciles

pkg/registration/hub/gc/controller.go
zhiweiyin318: I think each finalizer should be removed by the the controller who adds it on the mcl in general. 
if the finalizers are not removed after a period of time for example 10 minutes, gc controller should remove all finalizers to force delete mcl.  
the force deletion strategy including the resources in cluster ns and mcl needs to be discussed more

qiujian16: I do not think we should force remove the finalizer here. The only thing we can do is adding a condition stating why cluster is still deleting. Since the registration adds only one finalizer, I am also fine if we move adding finalizer to this controller.

zhiweiyin318: moved to this controller.

pkg/registration/hub/manager.go
```diff
@@ -182,12 +190,16 @@ func (m *HubManagerOptions) RunControllerManagerWithInformers(
 		controllerContext.EventRecorder,
 	)
 
-	rbacFinalizerController := rbacfinalizerdeletion.NewFinalizeController(
+	gcController := gc.NewGCController(
```
qiujian16, line 193: I think we might need a flag to let user specify a list gvr they would like to gc upon cluster deletion. This could help also to handle the addon's CRs in the cluster namespace

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	if cluster == nil {
+		return utilerrors.NewAggregate(errs)
+	}
+
+	newCluster := cluster.DeepCopy()
+	applyErrors := operatorhelpers.NewMultiLineAggregate(errs)
+	meta.SetStatusCondition(&newCluster.Status.Conditions, metav1.Condition{
+		Type:    ManagedClusterConditionDeleting,
```
qiujian16, line 152: this type should be in api repo I think?

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,176 @@
...
+		Status:  metav1.ConditionFalse,
+		Reason:  "Error",
```
qiujian16, line 154: followup no in necessarily in this PR but need a TODO: we should have more details, resourceRemaning？finalizerPending?

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,176 @@
...
+	if clusterName == "" || clusterName == factory.DefaultQueueKey {
+		return nil
+	}
+
+	cluster, err := r.clusterLister.Get(clusterName)
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	if cluster == nil {
```
qiujian16, line 114: do we really need to check the namespace?

pkg/registration/hub/gc/gc_resources.go
```diff
@@ -0,0 +1,62 @@
...
+	opts := metav1.DeleteOptions{PropagationPolicy: &background}
+	for _, resource := range r.deletedResourceList {
+		listOptions := metav1.ListOptions{
+			LabelSelector: resource.labelSelector,
+		}
+		err := r.metadataClient.Resource(resource.gvr).Namespace(clusterName).DeleteCollection(ctx, opts, listOptions)
+		if err != nil && !errors.IsNotFound(err) {
+			errs = append(errs, err)
+		}
+	}
```
qiujian16, line 55: I think you need to get after deletecollection, so we can know how much is left.

pkg/registration/hub/gc/gc_cluster_rbac.go
```diff
@@ -0,0 +1,193 @@
...
+
+	_, err = r.kubeClient.RbacV1().RoleBindings(clusterName).Update(ctx, workRoleBinding, metav1.UpdateOptions{})
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	return err
+}
+
+// removeFinalizer removes a finalizer from the list. It mutates its input.
+func removeFinalizer(obj runtime.Object, finalizerName string) bool {
```
qiujian16, line 158: I think we patcher to remove finalizer.

pkg/registration/hub/gc/gc_resources.go
```diff
@@ -0,0 +1,62 @@
...
+	var errs []error
+	if cluster != nil && cluster.DeletionTimestamp.IsZero() {
+		return gcReconcileContinue, nil
+	}
+
+	background := metav1.DeletePropagationBackground
+	opts := metav1.DeleteOptions{PropagationPolicy: &background}
+	for _, resource := range r.deletedResourceList {
+		listOptions := metav1.ListOptions{
+			LabelSelector: resource.labelSelector,
```
qiujian16, line 49: why do we need labelSelector? We should not delete manifestwork before addon is all deleted.

pkg/registration/hub/gc/gc_resources.go
```diff
@@ -0,0 +1,62 @@
...
+		err := r.metadataClient.Resource(resource.gvr).Namespace(clusterName).DeleteCollection(ctx, opts, listOptions)
+		if err != nil && !errors.IsNotFound(err) {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) != 0 {
+		return gcReconcileContinue, operatorhelpers.NewMultiLineAggregate(errs)
+	}
+	r.eventRecorder.Eventf("ManagedClusterGC", "managedCluster or ns %s is deleting and all addons and manifestworks are deleted", clusterName)
+
```
qiujian16, line 60: we should return how much is left as the error message and reflect it on the condition, and it should requeue if anything is left.

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,176 @@
...
+			return nil
+		}
+	}
+
+	var errs []error
+	for index := 0; index < len(r.gcReconcilers); index++ {
+		op, err := r.gcReconcilers[index].reconcile(ctx, clusterName, cluster)
+		switch op {
+		case gcReconcileRequeue:
+			controllerContext.Queue().AddAfter(clusterName, 10*time.Second)
```
qiujian16, line 131: 10 seconds seems quite long. I would start with 1

pkg/registration/hub/manager.go
zhiweiyin318: added a flag named `gc-resource-list`

pkg/registration/hub/gc/controller.go
zhiweiyin318: added to the deleting condition

pkg/registration/hub/gc/controller.go
zhiweiyin318: removed

pkg/registration/hub/gc/gc_resources.go
zhiweiyin318: added count the remaining resources

pkg/registration/hub/gc/gc_cluster_rbac.go
zhiweiyin318: done

pkg/registration/hub/gc/gc_resources.go
zhiweiyin318: removed

pkg/registration/hub/gc/gc_resources.go
zhiweiyin318: added

pkg/registration/hub/gc/controller.go
zhiweiyin318: done

pkg/registration/hub/gc/gc_resources.go
```diff
@@ -0,0 +1,228 @@
...
+		DeleteCollection(ctx, opts, metav1.ListOptions{LabelSelector: labelSelector})
+	switch {
+	case errors.IsNotFound(err):
+		return nil, true
+	case err != nil:
+		return err, false
+	}
+
+	notAllDeletedErr := fmt.Errorf("not all resources are deleted")
+	err = retry.OnError(retry.DefaultRetry,
```
qiujian16, line 162: why need retry here?

pkg/registration/hub/gc/gc_resources.go
```diff
@@ -0,0 +1,228 @@
...
+		if !deleted {
+			return nil
+		}
+	}
+	return nil
+}
+
+func (r *gcResourcesController) generateLabelSelectors(ctx context.Context,
+	resource schema.GroupVersionResource, clusterName string) ([]string, error) {
+	allResources, err := r.metadataClient.Resource(resource).Namespace(clusterName).List(ctx, metav1.ListOptions{})
```
qiujian16, line 95: if we also need a list here, should we list all at first so we do not need list again when calculating remaining.

pkg/registration/hub/gc/gc_resources.go
zhiweiyin318: removed

pkg/registration/hub/gc/gc_resources.go
zhiweiyin318: changed to list only once in each reconcile, calculate remanings, sort the resources by priority, then delete the resources with the lowest priority, and requeue.

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,161 @@
...
+	cluster := originalCluster.DeepCopy()
+	var errs []error
+	for _, reconciler := range r.gcReconcilers {
+		op, err := reconciler.reconcile(ctx, cluster)
+		if err != nil {
+			errs = append(errs, err)
+		}
+
+		if op == gcReconcileRequeue {
+			controllerContext.Queue().AddAfter(clusterName, 1*time.Second)
```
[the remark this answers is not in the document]
zhiweiyin318: moved to the end after process all reconcilers.  first patch the status to update the condition, and then check if requeue.

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		gcResources = append(gcResources, schema.GroupVersionResource{
+			Group: subStrings[0], Version: subStrings[1], Resource: subStrings[2]})
+	}
+
+	controller := &GCController{
+		clusterLister:  clusterInformer.Lister(),
+		clusterPatcher: clusterPatcher,
+		gcReconcilers: []gcReconciler{
+			newGCResourcesController(metadataClient, gcResources, eventRecorder),
```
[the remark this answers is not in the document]
zhiweiyin318: added a flag `gc-resource-list`,  the default value is to gc addon and manifestwork. should we disable it by default for backward compatible?
https://github.com/open-cluster-management-io/ocm/pull/229/files#diff-17bf0f8b650c4dbf7c59747fbe7131789d2aa6348f2c299e63c0a9db27bee4feR60

pkg/registration/hub/gc/controller.go
```diff
@@ -0,0 +1,155 @@
...
+	metadataClient metadata.Interface,
+	eventRecorder events.Recorder,
+	gcResourceList []string,
+) factory.Controller {
+	clusterPatcher := patcher.NewPatcher[
+		*clusterv1.ManagedCluster, clusterv1.ManagedClusterSpec, clusterv1.ManagedClusterStatus](
+		clusterClient.ClusterV1().ManagedClusters())
+
+	gcResources := []schema.GroupVersionResource{}
+	if len(gcResourceList) == 0 {
```
skeeey, line 76: nit: can we give a default value to `GCResourceList` instead of this?

pkg/registration/hub/gc/gc_resources.go
```diff
@@ -0,0 +1,166 @@
...
+	return len(resourceList.Items), finalizerPendingCnt
+}
+
+// stringToCleanupPriority is to convert a string to a cleanupPriority.
+// the range of cleanupPriority is [0,100].
+// set cleanupPriority to 0 if the string is not an int number or out of the range.
+func stringToCleanupPriority(value string) cleanupPriority {
+	priority, err := strconv.Atoi(value)
+	if err != nil || cleanupPriority(priority) > maxCleanupPriority ||
+		cleanupPriority(priority) < minCleanupPriority {
```
skeeey, line 162: nit: may need a warning message here

skeeey:
LGTM

only have some nit comments, leave other reviewer to take a look

pkg/registration/hub/gc/controller.go
zhiweiyin318: the gcResourceList is from the flag `gc-resource-list` with the default values addon and manifetworks.

pkg/registration/hub/gc/gc_resources.go
zhiweiyin318: added.

PR #231, opened 2023-07-25
:sparkles: Support override addon images by the cluster's annotation

zhujian7:
## Summary

Make the addon manger support override addon agents' images by the cluster's annotation.
If the managedCluster has annotation `open-cluster-management.io/image-registries` in format:
```
{
   "registries":[
      {
         "mirror":"quay.io/ocm/addon-examples",
         "source":"quay.io/open-cluster-management/addon-examples"
      }
   ]
}
```
For example:
```
apiVersion: cluster.open-cluster-management.io/v1
kind: ManagedCluster
metadata:
  annotations:
    open-cluster-management.io/image-registries: '{"registries":[{"mirror":"quay.io/ocm/addon-examples","source":"quay.io/open-cluster-management/addon-examples"}]}'
```
The addon-manager will override agent images for all the template-type addons whose images can match the `source` field.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/addon/templateagent/template_agent.go
+++ b/pkg/addon/templateagent/template_agent.go
@@ -116,6 +116,7 @@ func (a *CRDTemplateAgentAddon) GetAgentAddonOptions() agent.AgentAddonOptions {
 			CSRApproveCheck:   a.TemplateCSRApproveCheckFunc(),
 			CSRSign:           a.TemplateCSRSignFunc(),
 		},
+		AgentDeployTriggerClusterFilter: utils.ClusterImageRegistriesAnnotationChanged,
 	}
 }
 
--- a/pkg/addon/templateagent/values.go
+++ b/pkg/addon/templateagent/values.go
@@ -158,29 +158,13 @@ func hubKubeconfigPath() string {
 	return "/managed/hub-kubeconfig/kubeconfig"
 }
 
-const (
-	// ClusterImageRegistriesAnnotation represent the annotation key for image registries, the vale of the annotation
-	// should be a json string like this:
-	//
-	// {
-	//   "registries": [
-	//     {
-	//       "source": "quay.io/ocm",
-	//       "mirrors": "quay.io/open-cluster-management"
-	//     }
-	//   ]
-	// }
-	// TODO: move this to the api repo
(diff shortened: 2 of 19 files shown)


pkg/addon/templateagent/values.go
```diff
@@ -154,3 +157,49 @@ func (a *CRDTemplateAgentAddon) sortValueKeys(value addonfactory.Values) []strin
...
+	klog.V(4).Infof("Try to get image registries from annotation %v", annotations[ClusterImageRegistriesAnnotation])
+	if len(annotations[ClusterImageRegistriesAnnotation]) == 0 {
+		return values, nil
+	}
+	type ImageRegistries struct {
+		Registries []addonapiv1alpha1.ImageMirror `json:"registries"`
+	}
+
+	imageRegistries := ImageRegistries{}
+	err := json.Unmarshal([]byte(annotations[ClusterImageRegistriesAnnotation]), &imageRegistries)
```
zhiweiyin318, line 191: can it be unmarshalled correctly if there is pullSecret filed in the annotation?

zhujian7: yes

zhujian7:
/unhold 
cc @qiujian16

zhujian7:
/unhold
@qiujian16 ptal

pkg/addon/templateagent/template_agent.go
```diff
@@ -116,6 +117,9 @@ func (a *CRDTemplateAgentAddon) GetAgentAddonOptions() agent.AgentAddonOptions {
 			CSRApproveCheck:   a.TemplateCSRApproveCheckFunc(),
 			CSRSign:           a.TemplateCSRSignFunc(),
 		},
+		AgentDeployTriggerClusterFilter: func(old, new *clusterv1.ManagedCluster) bool {
+			return !equality.Semantic.DeepEqual(old.Annotations, new.Annotations)
```
qiujian16, line 121: do you think we should only check the change of certain key? to reduce the reconcile call.

zhujian7: fixed, PTAL

PR #234, opened 2023-07-25
:sparkles: Add ClusterAnnotations support.

xuezhaojun:
## Summary
The feature supports us to assign the annotations**(must have "agent.open-cluster-management.io" as prefix)** when creating a managed cluster from the agent side.

## Related issue(s)

https://github.com/open-cluster-management-io/api/pull/252


qiujian16:
/approve

/assign @zhiweiyin318 

@zhujian7 is there any case you want registry to be also set as this?

zhujian7:
> is there any case you want registry to be also set as this?

emmm, not yet. So far we should only set up the registry from the hub

Issue #233, opened 2023-07-25
Flaky unit test for the placement scheduling


zhujian7:
```
--- FAIL: TestBind (0.02s)
    --- FAIL: TestBind/create_placementdecision_with_only_cluster_per_decision_group (0.00s)
        scheduling_controller_test.go:1429: expected clusters selected: cluster1, but got [{cluster3 }]
        scheduling_controller_test.go:1429: expected clusters selected: cluster2, but got [{cluster1 }]
        scheduling_controller_test.go:1429: expected clusters selected: cluster3, but got [{cluster2 }]
FAIL
        open-cluster-management.io/ocm/pkg/placement/controllers/scheduling     coverage: 75.5% of statements
FAIL    open-cluster-management.io/ocm/pkg/placement/controllers/scheduling     0.298s
```

see https://github.com/open-cluster-management-io/ocm/actions/runs/5654038071/job/15316306014?pr=231

cc @haoqing0110

qiujian16:
I see this error sometimes also

/kind bug
/assign @haoqing0110

haoqing0110:
Will fix in https://github.com/open-cluster-management-io/ocm/pull/235

PR #235, opened 2023-07-26
:bug: fix the placement decision clusters have random order issue

haoqing0110:
## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/233

### DIFF
--- a/pkg/placement/controllers/scheduling/scheduling_controller.go
+++ b/pkg/placement/controllers/scheduling/scheduling_controller.go
@@ -465,12 +465,6 @@ func (c *schedulingController) generatePlacementDecisionsAndStatus(
 
 	// generate placement decision for each decision group
 	for decisionGroupIndex, decisionGroup := range decisionGroups {
-		// sort clusterdecisions by cluster name
-		clusterDecisions := decisionGroup.clusterDecisions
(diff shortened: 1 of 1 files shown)


pkg/placement/controllers/scheduling/scheduling_controller.go
```diff
@@ -517,7 +517,9 @@ func (c *schedulingController) generateDecisionGroups(
 
 	// The rest of the clusters will also be put into decision groups.
 	var matched []clusterapiv1beta1.ClusterDecision
-	for _, cluster := range clusterNames.UnsortedList() {
+	clusterNameList := clusterNameSet.UnsortedList()
+	sort.Strings(clusterNameList)
+	for _, cluster := range clusterNameList {
 		matched = append(matched, clusterapiv1beta1.ClusterDecision{
```
zhujian7, line 523: Not sure if sorting the `matched` is better, because if the `matched` get some results from `matched, status := filterClustersBySelector(d.ClusterSelector, clusters, clusterNameSet)`, are they in order？

haoqing0110: Good suggestion. Need to sort all the matched before dividing into groups.

PR #245, opened 2023-08-04
🐛 set tls min version to 1.2

elgnay:
## Summary
Some cluster with old Kubernetes version may not support TLS 1.3 yet.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/work/webhook/start.go
+++ b/pkg/work/webhook/start.go
@@ -36,7 +36,7 @@ func (c *Options) RunWebhookServer() error {
 		WebhookServer: webhook.NewServer(webhook.Options{
 			TLSOpts: []func(config *tls.Config){
 				func(config *tls.Config) {
-					config.MinVersion = tls.VersionTLS13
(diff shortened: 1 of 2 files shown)


pkg/work/webhook/start.go
```diff
@@ -34,13 +32,9 @@ func (c *Options) RunWebhookServer() error {
...
 		HealthProbeBindAddress: ":8000",
 		WebhookServer: webhook.NewServer(webhook.Options{
-			TLSOpts: []func(config *tls.Config){
-				func(config *tls.Config) {
-					config.MinVersion = tls.VersionTLS13
-				},
-			},
-			Port:    c.Port,
-			CertDir: c.CertDir,
+			TLSMinVersion: "1.2",
```
qiujian16, line 35: I think TLSMinVersion is deprecated, can we use TLSOpts instead?

elgnay: Updated the code to use TLSOpts.

PR #252, opened 2023-08-21
:bug:  Revert "Fix PodSecurity Warnings"

zhiweiyin318:
This reverts commit ecc541d186dbf80a35c9cee68762aa7f17791e19.

## Related issue(s)

Fixes #


xuezhaojun:
@zhiweiyin318  @qiujian16  Do we consider do a "fix" instead of a revert? That means we only "remove the seccompProfile" but leave the rest part still.

qiujian16:
let's do revert at first to avoid breaking the e2e. We need an approach to solve ssc anyway later
/approve
/lgtm

PR #253, opened 2023-08-22
:bug: revert placement decision start index from 1 to keep backward compatibility

haoqing0110:
## Related issue(s)

Fixes #


pkg/placement/controllers/scheduling/scheduling_controller.go
```diff
@@ -465,7 +465,8 @@ func (c *schedulingController) generatePlacementDecisionsAndStatus(
 
 	// generate placement decision for each decision group
 	for decisionGroupIndex, decisionGroup := range decisionGroups {
-		// generate placement decisions and status, group and placement name index starts from 0.
+		// generate placement decisions and status, group name index starts from 0, placement name index
+		// starts from 1 to keep backward compatibility.
```
qiujian16, line 469: add a todo here, we should be consistent with index or using a random generate name when version bumps

haoqing0110: added

Issue #191, opened 2023-06-20
use contextual logging in ocm


qiujian16:
**Describe the enhancement**
with contextual logging, it is easier to read and integrate with third party logging tools
https://www.kubernetes.dev/blog/2022/05/25/contextual-logging/

PR #254, opened 2023-08-22
:seedling: adding contextual logging in placement component

nitishchauhan0022:
Summary
Adding contextual logging

Related issue(s)
Fixes https://github.com/open-cluster-management-io/ocm/issues/191

For More Info:
https://www.kubernetes.dev/blog/2022/05/25/contextual-logging/
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md

### DIFF
--- a/pkg/placement/controllers/scheduling/schedule.go
+++ b/pkg/placement/controllers/scheduling/schedule.go
@@ -171,6 +171,7 @@ func (s *pluginScheduler) Schedule(
 	placement *clusterapiv1beta1.Placement,
 	clusters []*clusterapiv1.ManagedCluster,
 ) (ScheduleResult, *framework.Status) {
+	logger := klog.FromContext(ctx)
 	filtered := clusters
 	finalStatus := framework.NewStatus("", framework.Success, "")
 
@@ -190,7 +191,7 @@ func (s *pluginScheduler) Schedule(
 		case status.IsError():
 			return results, status
 		case status.Code() == framework.Warning:
-			klog.Warningf("%v", status.Message())
+			logger.Info("Warning status message", "message", status.Message())
 			finalStatus = status
 		}
 
@@ -207,7 +208,7 @@ func (s *pluginScheduler) Schedule(
 	case status.IsError():
 		return results, status
 	case status.Code() == framework.Warning:
-		klog.Warningf("%v", status.Message())
+		logger.Info("Warning status message", "message", status.Message())
 		finalStatus = status
 	}
 
@@ -217,7 +218,7 @@ func (s *pluginScheduler) Schedule(
 	case status.IsError():
 		return results, status
 	case status.Code() == framework.Warning:
-		klog.Warningf("%v", status.Message())
+		logger.Info("Warning status message", "message", status.Message())
 		finalStatus = status
 	}
 
@@ -235,7 +236,7 @@ func (s *pluginScheduler) Schedule(
 		case status.IsError():
 			return results, status
 		case status.Code() == framework.Warning:
-			klog.Warningf("%v", status.Message())
+			logger.Info("Warning status message", "message", status.Message())
 			finalStatus = status
 		}
 
--- a/pkg/placement/plugins/addon/addon.go
+++ b/pkg/placement/plugins/addon/addon.go
@@ -82,7 +82,7 @@ func (c *AddOn) Score(ctx context.Context, placement *clusterapiv1beta1.Placemen
 		// get AddOnPlacementScores CR with resourceName
 		addOnScores, err := c.handle.ScoreLister().AddOnPlacementScores(namespace).Get(c.resourceName)
 		if err != nil {
-			klog.Warningf("Getting AddOnPlacementScores failed: %s", err)
+			klog.FromContext(ctx).Info("Failed to get AddOnPlacementScores", "error", err)
 			continue
 		}
 
--- a/pkg/placement/plugins/tainttoleration/taint_toleration.go
+++ b/pkg/placement/plugins/tainttoleration/taint_toleration.go
@@ -91,7 +91,7 @@ func (pl *TaintToleration) Filter(ctx context.Context, placement *clusterapiv1be
 func (pl *TaintToleration) RequeueAfter(ctx context.Context, placement *clusterapiv1beta1.Placement) (plugins.PluginRequeueResult, *framework.Status) {
 	status := framework.NewStatus(pl.Name(), framework.Success, "")
 	// get exist decisions clusters
-	decisionClusterNames, decisionClusters := getDecisionClusters(pl.handle, placement)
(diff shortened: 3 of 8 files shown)


pkg/placement/plugins/addon/addon.go
```diff
@@ -82,7 +82,7 @@ func (c *AddOn) Score(ctx context.Context, placement *clusterapiv1beta1.Placemen
 		// get AddOnPlacementScores CR with resourceName
 		addOnScores, err := c.handle.ScoreLister().AddOnPlacementScores(namespace).Get(c.resourceName)
 		if err != nil {
-			klog.Warningf("Getting AddOnPlacementScores failed: %s", err)
+			klog.FromContext(ctx).Info("Failed to get AddOnPlacementScores", "error", err)
```
haoqing0110, line 85: How about add the logger to interface [Score](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/placement/plugins/interface.go#L55
)
```
type Prioritizer interface {
	Plugin

	// Score gives the score to a list of the clusters, it returns a map with the key as
	// the cluster name.
	Score(ctx context.Context, logger Logger, placement *clusterapiv1beta1.Placement, clusters []*clusterapiv1.ManagedCluster) (PluginScoreResult, *framework.Status)
}
```
Then only need to pass the logger to Score in https://github.com/open-cluster-management-io/ocm/blob/main/pkg/placement/controllers/scheduling/schedule.go#L231
```
		scoreResult, status := p.Score(ctx, logger, placement, filtered)
```
So that all the placement plugins can use the logger, and don't need to init it by plugin itself.

xuezhaojun, line 85: Feels like we should use `klog.FromContext(ctx).Warn` instead of `Info` here to keep it the same as before?

pkg/placement/controllers/scheduling/schedule.go
```diff
@@ -190,7 +191,7 @@ func (s *pluginScheduler) Schedule(
 		case status.IsError():
 			return results, status
 		case status.Code() == framework.Warning:
-			klog.Warningf("%v", status.Message())
```
xuezhaojun, line 193: Can we keep the level still be `Warn`

pkg/placement/plugins/tainttoleration/taint_toleration.go
```diff
@@ -226,15 +226,15 @@ func getDecisionClusterNames(handle plugins.Handle, placement *clusterapiv1beta1
...
+func getDecisionClusters(logger klog.Logger, handle plugins.Handle, placement *clusterapiv1beta1.Placement) (sets.String, []*clusterapiv1.ManagedCluster) {
 	// get existing decision cluster name
 	decisionClusterNames := getDecisionClusterNames(handle, placement)
 
 	// get existing decision clusters
 	decisionClusters := []*clusterapiv1.ManagedCluster{}
 	for c := range decisionClusterNames {
 		if managedCluser, err := handle.ClusterLister().Get(c); err != nil {
-			klog.Warningf("Failed to get ManagedCluster: %s", err)
+			logger.Info("Failed to get ManagedCluster", "error", err)
```
xuezhaojun, line 237: The same concern as the above.

pkg/placement/plugins/addon/addon.go
haoqing0110, line 85: How about adding the logger to schedulerHandler, so that each plugin can use the logger directly, don't need to init it each time by plugin itself. 
https://github.com/haoqing0110/OCM/commit/deffa73b2d94fe61cab62173df7016e1e6a0e54f express my idea.

pkg/placement/plugins/tainttoleration/taint_toleration.go
```diff
@@ -91,7 +91,7 @@ func (pl *TaintToleration) Filter(ctx context.Context, placement *clusterapiv1be
 func (pl *TaintToleration) RequeueAfter(ctx context.Context, placement *clusterapiv1beta1.Placement) (plugins.PluginRequeueResult, *framework.Status) {
 	status := framework.NewStatus(pl.Name(), framework.Success, "")
 	// get exist decisions clusters
-	decisionClusterNames, decisionClusters := getDecisionClusters(pl.handle, placement)
+	decisionClusterNames, decisionClusters := getDecisionClusters(klog.FromContext(ctx), pl.handle, placement)
```
haoqing0110, line 94: Same as above, this can bring benefits to all the plugins.

pkg/placement/plugins/addon/addon.go
xuezhaojun: A different opinion, to me add the logger to schedulerHandler feels like we still hold a global logger but in another format.

Recap the idea of contextual log https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/3077-contextual-logging/README.md , we want:

```
Log calls that were already converted to structured logging need to be updated to use a logger that gets passed into the function in one of two ways:

as explicit parameter
attached to a context.Context
```

nitishchauhan0022: Hey @haoqing0110 thanks, it is a great idea. I have updated it as per this. Please take a look

pkg/placement/plugins/tainttoleration/taint_toleration.go
nitishchauhan0022: updated it

pkg/placement/plugins/addon/addon.go
nitishchauhan0022: hey @xuezhaojun, warning function has been removed. Please take a look. [MigratingGuideline](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md)

image

xuezhaojun: Got it! Thanks for explaining.

pkg/placement/controllers/scheduling/schedule.go
xuezhaojun: /Close

pkg/placement/plugins/tainttoleration/taint_toleration.go
xuezhaojun: /Close

pkg/placement/plugins/addon/addon.go
nitishchauhan0022: @xuezhaojun It is a valid point, if we want to configure at a function level.

haoqing0110: @nitishchauhan0022 @xuezhaojun 's suggestion makes sense I think, checking the Kubernetes scheduler, they init the logger inside plugin https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/noderesources/resource_allocation.go#L53

nitishchauhan0022: Ok, i will revert back to previous changes

PR #258, opened 2023-08-28
✨ support singleton in hosted mode

qiujian16:
## Related issue(s)

Fixes #

### DIFF
--- a/manifests/klusterlet/management/klusterlet-agent-deployment.yaml
+++ b/manifests/klusterlet/management/klusterlet-agent-deployment.yaml
@@ -127,5 +127,5 @@ spec:
       {{if eq .InstallMode "SingletonHosted"}}
       - name: spoke-kubeconfig-secret
         secret:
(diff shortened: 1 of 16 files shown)


manifests/klusterlet/management/klusterlet-agent-deployment.yaml
```diff
@@ -115,3 +124,8 @@ spec:
       - name: hub-kubeconfig
         emptyDir:
           medium: Memory
+      {{if eq .InstallMode "SingletonHosted"}}
+      - name: spoke-kubeconfig-secret
+        secret:
+          secretName: {{ .ExternalManagedKubeConfigWorkSecret }}
```
zhujian7, line 130: for singletonHosted mode, will it still generate two(external registration/work) secrets? why use the work secret here?

qiujian16: yeah, it is a bit confusing. Updated and add some comment.

PR #259, opened 2023-08-29
✨ Implement ManifestWorkReplicaSet RollOut strategy 

serngawy:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
+++ b/pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
@@ -12,12 +12,12 @@ import (
 
 	clusterlister "open-cluster-management.io/api/client/cluster/listers/cluster/v1beta1"
 	worklisterv1 "open-cluster-management.io/api/client/work/listers/work/v1"
-	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
+	clusterv1alpha1 "open-cluster-management.io/api/cluster/v1alpha1"
 	"open-cluster-management.io/api/utils/work/v1/workapplier"
 	workv1 "open-cluster-management.io/api/work/v1"
 	workapiv1alpha1 "open-cluster-management.io/api/work/v1alpha1"
 
-	"open-cluster-management.io/ocm/pkg/common/helpers"
+	"open-cluster-management.io/ocm/pkg/work/helper"
 )
 
 // deployReconciler is to manage ManifestWork based on the placement.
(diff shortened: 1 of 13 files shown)


pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_test.go
```diff
@@ -200,3 +202,255 @@ func TestDeployReconcileAsPlacementNotExist(t *testing.T) {
...
+		} else if plcSummary.Name == placement2.Name {
+			assert.Equal(t, plcSummary, expectPlcSummary2)
+		} else {
+			t.Fatal("PlacementSummary Name not Exist")
+		}
+	}
+
+	// Check for the expected manifestWorkReplicaSetSummary
+	mwrSetSummary := workapiv1alpha1.ManifestWorkReplicaSetSummary{
+		Total:       clsPerGroup + int(placement1.Status.NumberOfSelectedClusters),
```
haoqing0110, line 433: What if there's clusters overlap in placement1 and placement2, will the cluster number be counted twice in the summary?

haoqing0110: My thought is a cluster counted twice makes the meaning of [Total](https://github.com/open-cluster-management-io/api/blob/main/work/v1alpha1/types_manifestworkreplicaset.go#L116) confusing to the user, since the actual number of ManifestWorks is less than it. 

For addon, one addon only belongs to one placement (the latter one), and use that rollout strategy do rollout. Maybe it's not the same as the mwrSet here, just for reference

qiujian16:
sorry I am stuck by other stuff this week, will go through this PR early next week.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
```diff
@@ -31,90 +32,121 @@ type deployReconciler struct {
...
+		}
 
-	var errs []error
-	addedClusters, deletedClusters, existingClusters := sets.New[string](), sets.New[string](), sets.New[string]()
-	for _, mw := range manifestWorks {
-		existingClusters.Insert(mw.Namespace)
-	}
+		for _, mw := range manifestWorks {
+			// Check if ManifestWorkTemplate changes, ManifestWork will need to be updated.
+			if !equality.Semantic.DeepEqual(mwrSet.Spec.ManifestWorkTemplate, mw.Spec) {
```
qiujian16, line 60: this check is not quite reliable, since there could diff on somefield with empty or nil values. It is better to user the work applier helper.

I think we should always apply, and check the returned updated value to know if if it is updated.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
```diff
@@ -124,9 +156,39 @@ func (d *deployReconciler) reconcile(ctx context.Context, mwrSet *workapiv1alpha
...
+func (d *deployReconciler) clusterRolloutStatusFunc(clusterName string, manifestWork workv1.ManifestWork) (clusterv1alpha1.ClusterRolloutStatus, error) {
+	clsRolloutStatus := clusterv1alpha1.ClusterRolloutStatus{
+		ClusterName:        clusterName,
+		LastTransitionTime: &manifestWork.CreationTimestamp,
+		// Default status as ToApply
+		Status: clusterv1alpha1.ToApply,
+	}
+	// applied and Progressing conditions return status as Progressing
+	if apimeta.IsStatusConditionTrue(manifestWork.Status.Conditions, workv1.WorkApplied) ||
+		apimeta.IsStatusConditionTrue(manifestWork.Status.Conditions, workv1.WorkProgressing) {
```
qiujian16, line 177: do we have a WorkProgressing status right now?

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
```diff
@@ -124,9 +156,39 @@ func (d *deployReconciler) reconcile(ctx context.Context, mwrSet *workapiv1alpha
...
+		clsRolloutStatus.Status = clusterv1alpha1.Progressing
+	}
+	// Available condition return status as Succeeded
+	if apimeta.IsStatusConditionTrue(manifestWork.Status.Conditions, workv1.WorkAvailable) {
+		clsRolloutStatus.Status = clusterv1alpha1.Succeeded
+	}
+	// Degraded condition return status as Failed
+	if apimeta.IsStatusConditionTrue(manifestWork.Status.Conditions, workv1.WorkDegraded) {
```
qiujian16, line 185: There is not degraded status defined yet. We should add some note here.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
qiujian16: Add a TODO here: we should have an interface in applier to check whether a mw should be applied.

qiujian16 (commented):
/approve

@haoqing0110 would you take a final look on this?

haoqing0110:
LGTM
Only one concern for the overlapped clusters being counted twice in the summary message. This could be a follow-up fix when receiving more feedback.

serngawy:
> LGTM Only one concern for the overlapped clusters being counted twice in the summary message. This could be a follow-up fix when receiving more feedback.

Yes, we decided to keep it as it is for now (manifestWork will be associated with the latest placement) cause it can be valid for both ways.  Will wait for feedback if any changes is necessary.

PR #260, opened 2023-08-31
✨ support proxy between hub cluster and managed cluster

elgnay:
## Summary
Allow user to configure proxy settings in bootstrap hub kubeconfig.
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/clientcert/certificate.go
+++ b/pkg/registration/clientcert/certificate.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	csrclient "k8s.io/client-go/kubernetes/typed/certificates/v1"
 	certificatesv1listers "k8s.io/client-go/listers/certificates/v1"
-	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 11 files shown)


pkg/registration/clientcert/certificate.go
```diff
@@ -147,13 +147,19 @@ func getCertValidityPeriod(secret *corev1.Secret) (*time.Time, *time.Time, error
 
 // BuildKubeconfig builds a kubeconfig based on a rest config template with a cert/key pair
 func BuildKubeconfig(clientConfig *restclient.Config, certPath, keyPath string) clientcmdapi.Config {
```
qiujian16, line 149: why we still need this func? or I think we can just have BuildKubeconfig to also use proxyURL as an input, or the we should use cliemtcmdapi.config instead of rest.Config as the input.

elgnay: The signature of BuildKubeconfig has been updated as suggested.

qiujian16:
/approve

would you also check the codecov? it seems it reduces a bit

xuezhaojun (commented):
/lgtm

xuezhaojun:
/hold

Looks clear and good to me! feel free to unhold if no updates.

PR #261, opened 2023-09-04
🐛 move patch routine to copy crd

qiujian16:
This can avoid redudant file generated in _output

## Related issue(s)

Fixes #


haoqing0110:
LGTM
Just one nit: https://github.com/open-cluster-management-io/ocm/blob/main/hack/init.sh#L18 PATCHED_DIR seems no longer be used, should be deleted.

Makefile
```diff
@@ -64,7 +61,7 @@ update-csv: ensure-operator-sdk
 	rm ./deploy/cluster-manager/olm-catalog/cluster-manager/manifests/cluster-manager_v1_serviceaccount.yaml
 	rm ./deploy/klusterlet/olm-catalog/klusterlet/manifests/klusterlet_v1_serviceaccount.yaml
 
-verify-crds: patch-crd
+verify-crds:
```
haoqing0110, line 64: seems need to pass parameter $(YAML_PATCH) to hack/verify-crds.sh

PR #263, opened 2023-09-04
🐛 Remove old agent when upgrade to singleton

qiujian16:
## Related issue(s)

Fixes #


pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_runtime_reconcile.go
```diff
@@ -165,6 +174,18 @@ func (r *runtimeReconcile) installSingletonAgent(ctx context.Context, klusterlet
 		return klusterlet, reconcileStop, err
 	}
 
+	// clean registration/work if there is any
```
zhujian7, line 177: do we need to clean the other resources like role/rolebindings?

qiujian16: they are shared and will be updated accordingly when mode is changed, so not need to remove.

PR #265, opened 2023-09-04
🐛 correct the agent deployment name in different install mode

elgnay:
## Related issue(s)

Fixes #


pkg/operator/operators/klusterlet/controllers/bootstrapcontroller/bootstrapcontroller.go
```diff
@@ -181,7 +181,11 @@ func (k *bootstrapController) sync(ctx context.Context, controllerContext factor
 
 func (k *bootstrapController) processRebootstrap(ctx context.Context, agentNamespace string, klusterlet *operatorapiv1.Klusterlet,
 	recorder events.Recorder, requeueFunc func(time.Duration)) error {
-	deploymentName := fmt.Sprintf("%s-registration-agent", klusterlet.Name)
+	deploymentName := fmt.Sprintf("%s-agent", klusterlet.Name)
+	if len(klusterlet.Spec.DeployOption.Mode) == 0 || klusterlet.Spec.DeployOption.Mode == operatorapiv1.InstallModeDefault ||
```
qiujian16, line 185: I think it is not Singleton. We have a helper func to check it

elgnay: Updated as suggested.

qiujian16 (commented):
/approve
/lgtm

PR #266, opened 2023-09-05
:sparkles: set agent install namespace from addon deployment config for template type addons

zhujian7:
## Summary

set agent install namespace from addon deployment config for template type addons, so for template type addons, users can use the addonDeploymentConfig API to configure the agent install namespace.

## Related issue(s)

Fixes #


qiujian16 (commented):
/approve
/lgtm

PR #275, opened 2023-09-11
🐛 fix the AlreadyExists logic if the manifest is not cached by G…

elgnay:
…etter when applying a manifest

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/common/apply/interface.go
+++ b/pkg/common/apply/interface.go
@@ -19,6 +19,8 @@ type Getter[T runtime.Object] interface {
 
 // Client is a wrapper interface of client
 type Client[T runtime.Object] interface {
+	Get(ctx context.Context, name string, opts metav1.GetOptions) (T, error)
+
 	Create(ctx context.Context, obj T, opts metav1.CreateOptions) (T, error)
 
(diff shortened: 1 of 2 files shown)


pkg/common/apply/interface.go
```diff
@@ -43,20 +45,24 @@ func Apply[T runtime.Object](
 	existing, err := getter.Get(requiredAccessor.GetName())
 	if errors.IsNotFound(err) {
 		actual, createErr := client.Create(ctx, required, metav1.CreateOptions{})
-		if errors.IsAlreadyExists(createErr) {
-			return required, false, nil
-		}
-		if createErr == nil {
+		switch {
+		case errors.IsAlreadyExists(createErr):
+			actual, getErr := client.Get(ctx, requiredAccessor.GetName(), metav1.GetOptions{})
```
qiujian16, line 50: should we add an comment on why we need to get again?

qiujian16:
/approve

only thing is I think we need to add a comment

pkg/common/apply/interface.go
elgnay: Comment added

PR #276, opened 2023-09-12
:bug: add POD_NAME for event refobject

zhiweiyin318:
## Summary
the event reported by each deployment have the incorrect object because cannot find the ref object of each component.
add POD_NAME env for event to get correct ref object.
## Related issue(s)

Fixes #


qiujian16 (commented):
/approve
/lgtm

manifests/cluster-manager/hub/cluster-manager-registration-webhook-clusterrole.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 rules:
 # Allow managedcluster admission to get/list/watch configmaps
 - apiGroups: [""]
-  resources: ["configmaps"]
+  resources: ["configmaps", "pods"]
```
qiujian16, line 8: why we need this?

zhiweiyin318: removed. webhook does not use library-go to create event, so does not need the GET pod rbac.

zhiweiyin318:
@qiujian16  we cannot set ComponentOwnerReference because we cannot get the deployment names at the beginning, the deployment names have a prefix `klusterterName- `or `clustermanagerName-`.
seems like we have to use env POD_NAME for event, and the deployment requires the GET pods and replicsets rbac to get OwnerReference of pods. 
another option is to set a fixed name OwnerReference, it can work because library-go will not  get the ComponentOwnerReference, only fill it to the event spec and create the event in the same ns.

manifests/klusterlet/managed/klusterlet-registration-clusterrole.yaml
```diff
@@ -10,6 +10,14 @@ rules:
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["get", "list", "watch"]
+# get pods and replicasets is for event creation
```
qiujian16, line 13: I do not think we need permission on managed cluster, we only need it in management part, right?

zhiweiyin318: the klusterlet registration/work agents also create event on managed cluster, so need the permission too

zhiweiyin318: removed since added on the wrong place.

qiujian16 (commented):
/approve
/lgtm

PR #284, opened 2023-09-20
:bug:  fix failed update roleRef in clusterrolebinding in upgrade case

zhiweiyin318:
## Summary
the roleRef is changed in work-execution clusterrolebinding, so failed to update the clusterrolebinding during upgrade scenario. 
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_managed_reconcile.go
+++ b/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_managed_reconcile.go
@@ -38,9 +38,12 @@ var (
 		"klusterlet/managed/klusterlet-work-clusterrole.yaml",
 		"klusterlet/managed/klusterlet-work-clusterrole-execution.yaml",
 		"klusterlet/managed/klusterlet-work-clusterrolebinding.yaml",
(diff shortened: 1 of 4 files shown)


pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_managed_reconcile.go
```diff
@@ -38,7 +38,7 @@ var (
 		"klusterlet/managed/klusterlet-work-clusterrole.yaml",
 		"klusterlet/managed/klusterlet-work-clusterrole-execution.yaml",
 		"klusterlet/managed/klusterlet-work-clusterrolebinding.yaml",
-		"klusterlet/managed/klusterlet-work-clusterrolebinding-execution.yaml",
+		"klusterlet/managed/klusterlet-work-clusterrolebinding-aggregate.yaml",
```
zhujian7, line 41: should we also add the old `open-cluster-management:{{ .KlusterletName }}-work:execution` in the list?

zhiweiyin318: defined another cleanedManagedStaticResourceFiles to delete all manifests.

PR #289, opened 2023-10-08
🐛 Fix release image manifest error

qiujian16:
## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/releaseimage.yml
+++ b/.github/workflows/releaseimage.yml
@@ -74,6 +74,11 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ env, images ]
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/releaseimage.yml
```diff
@@ -74,6 +74,8 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ env, images ]
     steps:
+      - name: checkout code
+        uses: actions/checkout@v3
```
zhujian7, line 78:
```suggestion
        uses: actions/checkout@v3
        with:
          fetch-depth: 1
          path: go/src/open-cluster-management.io/ocm
```

PR #301, opened 2023-10-28
:bug: should use hubClient instead of bootstrapClient for a  creatingClusterController.

xuezhaojun:
## Summary

The controller resync every 1 min through the life-cycle of the registration-agent, but the bootstrapkubeconfig is expected to be a short-term thing.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/spoke/spokeagent.go
+++ b/pkg/registration/spoke/spokeagent.go
@@ -177,19 +177,21 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 	if err != nil {
 		return err
 	}
+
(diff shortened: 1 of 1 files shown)


qiujian16:
assign @skeeey

xuezhaojun:
> /assign @skeeey
> 
> I do not remember whether we want creating cluster using hub kubeconfig (I thought it was desired) and if it has the permission to do that.

Double check the code:
https://github.com/open-cluster-management-io/ocm/blob/c4bfa0814ab3b3898f1ca0aaf7ea9a6969d02988/pkg/registration/spoke/registration/creating_controller.go#L58-L62

It seems although the controller will live, it won't function after the bootstrap stage. 

Then no need to using hubClient, PR close.

PR #310, opened 2023-11-09
:warning: Update OCM APIs and apply Rollout Strategy API changes

serngawy:
## Summary

Update OCM APIs and apply Rollout Strategy API changes

## Related issue(s)

Fixes #


pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler_test.go
```diff
@@ -799,6 +856,9 @@ func TestAddonConfigReconcile(t *testing.T) {
 									{GroupName: "group1"},
 								},
 							},
+							RolloutConfig: clusterv1alpha1.RolloutConfig{
```
haoqing0110, line 859: Will the test case report error if not adding `None` explicitly? since I though empty string will works as `None` refer to https://github.com/open-cluster-management-io/api/blob/main/cluster/v1alpha1/helpers.go#L554.

haoqing0110: Create a fix PR https://github.com/open-cluster-management-io/api/pull/295, I think we should allow the user not define ProgressivePerGroup filed and just use the default value.

serngawy: okay, I agree. will update the API commit version based on the above PR fix.

qiujian16:
/approve

Leave to @haoqing0110 for final review.

PR #312, opened 2023-11-14
:sparkles:  Add a condition to report when hub and agent clock out of sync.

xuezhaojun:
## Summary

When hub and agent clock out of sync, the managed cluster status will show as 'Unknown' even the agent functions well.

Here we add a new condition to report clock out-of-sync, when we detect the clock of hub and a agent is not synced.

**The strategy:** 
Every time hub find the `renewTime` updated, we compare the `now` with `renewTime`. The 2 values should close to each other, if the `now` is faster than `renewTime` too much (over a lease duration), we will report clock out-of-sync.

depends on: https://github.com/open-cluster-management-io/api/pull/296

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/hub/lease/clocksynccontroller.go
+++ b/pkg/registration/hub/lease/clocksynccontroller.go
@@ -1,0 +1,150 @@
+package lease
+
+import (
+	"context"
+	"time"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	coordv1 "k8s.io/api/coordination/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	coordinformers "k8s.io/client-go/informers/coordination/v1"
+	coordlisters "k8s.io/client-go/listers/coordination/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	clusterv1informer "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterv1listers "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+
+	"open-cluster-management.io/ocm/pkg/common/patcher"
+	"open-cluster-management.io/ocm/pkg/common/queue"
+)
+
+type clockSyncController struct {
+	patcher       patcher.Patcher[*clusterv1.ManagedCluster, clusterv1.ManagedClusterSpec, clusterv1.ManagedClusterStatus]
+	clusterLister clusterv1listers.ManagedClusterLister
+	leaseLister   coordlisters.LeaseLister
+	eventRecorder events.Recorder
+}
+
+const (
+	clockSyncControllerName = "ClockSyncController"
+)
+
+func NewClockSyncController(
+	clusterClient clientset.Interface,
+	clusterInformer clusterv1informer.ManagedClusterInformer,
+	leaseInformer coordinformers.LeaseInformer,
+	recorder events.Recorder,
+) factory.Controller {
+	c := &clockSyncController{
+		patcher: patcher.NewPatcher[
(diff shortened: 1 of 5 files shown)


skeeey:
I remembered we intentionally didn't use local time here, because we want to keep the time sync between the managed cluster and hub cluster

/cc @qiujian16

xuezhaojun:
> I remembered we intentionally didn't use local time here, because we want to keep the time sync between the managed cluster and hub cluster
> 
> /cc @qiujian16

That's new to me, do we know the motivation of keeping clocks sync between hub and agents?
And if we intentionally avoid using local time, we should mention this as a prerequisite of using ocm.

xuezhaojun:
Find the comment: https://github.com/open-cluster-management-io/registration/pull/40#discussion_r438329371

It seems the initial implement is based on probeTimestamp but we change to using observedLease directly in this PR: https://github.com/open-cluster-management-io/registration/pull/43

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_test.go
```diff
@@ -222,7 +222,9 @@ func TestDeployWithRolloutStrategyReconcileAsExpected(t *testing.T) {
 	perGoupeRollOut := clusterv1alpha1.RolloutStrategy{
 		Type: clusterv1alpha1.ProgressivePerGroup,
 		ProgressivePerGroup: &clusterv1alpha1.RolloutProgressivePerGroup{
-			Timeout: clusterv1alpha1.Timeout{Timeout: "None"},
+			RolloutConfig: clusterv1alpha1.RolloutConfig{
+				Timeout: "None",
```
xuezhaojun, line 226: This is a catch up of: https://github.com/open-cluster-management-io/api/pull/281

xuezhaojun:
Hi, @qiujian16 , I tried the bareInformer approach as well, does that means:

1. we need a independent controller watch leaseinformer using the following bare setting:
```go
 informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
        UpdateFunc: func(oldObj, newObj interface{}) {
            oldLease := oldObj.(*corev1.Lease)
            newLease := newObj.(*corev1.Lease)
            if oldLease.Spec.RenewTime.Before(newLease.Spec.RenewTime) {
              c.equeue(newLease)
            }
        },
    })
```

2. In the new controller,  every reconcile is triggered by a `renewTime` update event, and when it occurs, we assume the `now` on the agent-side which is lease's renewTime, should be close to(in 1 leaseDuration)  the `now` on the hub-side which we can get by `time.Now()`


The concern of this approach is, [the controller has a default 10 hours resync setting](https://github.com/kubernetes-sigs/controller-runtime/blob/b81fc66fded5b2e773c91394f9ff3060fcfb9539/pkg/cache/cache.go#L149), this resync is not triggered by the renew update event. 

This may cause misreport of out-of-sync condition.

qiujian16:
I think you should have a separated controller to update this condition. The reconcile logic is not quite the same as what lease did before.

xuezhaojun:
> I think you should have a separated controller to update this condition. The reconcile logic is not quite the same as what lease did before.

Got it, do you think we can seprete a controller but still use `savedLease` approach to avoid misreport by 10 hours resync of controller-runtime?

qiujian16:
My understanding is if you make it in the different controller, you would not need a local cache. isn't it?

pkg/registration/hub/lease/clocksynccontroller.go
```diff
@@ -0,0 +1,200 @@
...
+			leaseRenewTimeUpdateInformer,
+		).
+		WithSync(c.sync).
+		ToController(clockSyncControllerName, recorder)
+}
+
+func renewUpdateInfomer(queue workqueue.RateLimitingInterface, leaseInformer coordinformers.LeaseInformer) cache.SharedIndexInformer {
+	leaseRenewTimeUpdateInformer := leaseInformer.Informer()
+	_, err := leaseRenewTimeUpdateInformer.AddEventHandler(&cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(oldObj, newObj interface{}) {
```
xuezhaojun, line 106: Only the renew update event will trigger the reconcile, but I'm afriad of the default 10 hours reconcile of controller-runtime will cause a incorrect reconcile. That's why we still need the local cache. 
@qiujian16

qiujian16: I am not sure I follow. I do not think it needs to care the resync, the reconile should be only triggered upon update event.

xuezhaojun: Got it, the controller updated.

qiujian16 (commented):
/approve

/assign @elgnay

PR #316, opened 2023-11-22
:bug:: remove managedclustersetbinding v1beta1 in webhook.

xuezhaojun:
## Summary

We have split the webhook configuration of managedclustersetbinding to 2 configs in this PR: https://github.com/open-cluster-management-io/registration-operator/pull/293

The previous release are not affected and all functionalities are fine, just remove v1beta1 content to make versions more clear.

## Related issue(s)

Fixes #


manifests/cluster-manager/hub/cluster-manager-registration-webhook-clustersetbinding-validatingconfiguration.yaml
```diff
@@ -22,6 +22,6 @@ webhooks:
     - v1beta2
     resources:
     - managedclustersetbindings
-  admissionReviewVersions: ["v1beta1","v1"]
+  admissionReviewVersions: ["v1beta2","v1"]
```
zhujian7, line 25: seems the [admissionReviewVersions](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#webhook-request-and-response) is not related to the to-be-validated resources version? Not sure if we should change this.

xuezhaojun: In fact, this list is not used in our customized webhook logic,  the only field we leverage is UserInfo: https://github.com/open-cluster-management-io/ocm/blob/6571a8a4d9aadae5014c9243a5acdb3667e4758f/pkg/registration/webhook/v1beta2/managedclustersetbinding_validating.go#L39

So in this configuration, we can put any version for example "v2".  We change it to v1beta2 just to aglin with our version migration process.

xuezhaojun: Got it, changed back.

qiujian16 (commented):
/approve
/lgtm

PR #320, opened 2023-11-28
📖 Fix typos in ClusterServiceVersion description

grdryn:
-------

## Summary

Some minor typo fixes in the description that gets displayed in places
like the operator hub.

## Related issue(s)

Fixes #

### DIFF
--- a/deploy/klusterlet/config/manifests/bases/klusterlet.clusterserviceversion.yaml
+++ b/deploy/klusterlet/config/manifests/bases/klusterlet.clusterserviceversion.yaml
@@ -23,10 +23,10 @@ spec:
       name: klusterlets.operator.open-cluster-management.io
       version: v1
   description: |
-    The Klusterlet provides the registration to the Hub clusters as a managed cluster. This operator supports the installation and upgrade of Klusterlet.
+    The Klusterlet provides the registration to the Hub cluster as a managed cluster. This operator supports the installation and upgrade of Klusterlet.
 
     ## Prerequisites
-    You need a Hub cluster which has installed ClusterManager.
+    You need a Hub cluster with ClusterManager installed.
 
     ## How to Install
     * Install Klusterlet operator by following instructions in top right button `Install`.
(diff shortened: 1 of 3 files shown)


deploy/klusterlet/config/manifests/bases/klusterlet.clusterserviceversion.yaml
```diff
@@ -23,18 +23,18 @@ spec:
...
   description: |
-    The Klusterlet proivdes the registation to the Hub clusters as a managed cluster. This operator supports the installation and upgrade of Klusterlet.
+    The Klusterlet provides the registration to the Hub cluster as a managed cluster. This operator supports the installation and upgrade of Klusterlet.
 
     ## Prerequisites
-    You need a Hub cluster which has installed ClusterManager.
+    You need a Hub cluster with ClusterManager installed.
 
     ## How to Install
     * Install Klusterlet operator by following instructions in top right button `Install`.
```
grdryn, line 32: At least in the OperatorHub in OpenShift 4.14, the "Install" button doesn't appear to be in the top _right_. I'm not sure if it is in other environments. If so, maybe this could/should be a little more accommodating of potential UI changes?

zhiweiyin318:
Thanks for the contribution.  
we need add `-s` when `git commit`  to pass the DCO check.
could you run `make update-csv ` locally which is to update the[ klusterlet csv file](https://github.com/open-cluster-management-io/ocm/blob/main/deploy/klusterlet/olm-catalog/klusterlet/manifests/klusterlet.clusterserviceversion.yaml) ?

deploy/klusterlet/config/manifests/bases/klusterlet.clusterserviceversion.yaml
zhiweiyin318: I think the install button is displayed on the https://operatorhub.io/operator/klusterlet page.

grdryn: Ah right, I didn't think of that page.

This text also appears in the in-cluster OperatorHub in OpenShift, where the "Install" button tends to be further left than the text, so I think my point about not being too specific to one particular UI may still be relevant? I don't intend to make any changes around that in this PR, just highlighting it in case it could be improved.

image

grdryn:
> we need add `-s` when `git commit` to pass the DCO check.
> could you run `make update-csv ` locally which is to update the[ klusterlet csv file](https://github.com/open-cluster-management-io/ocm/blob/main/deploy/klusterlet/olm-catalog/klusterlet/manifests/klusterlet.clusterserviceversion.yaml) ?

@zhiweiyin318 sorry, I completely missed this at the time. I just noticed it now when I got a notification that this needed rebasing (you addressed some of these typos in #548, so this commit conflicted).

I've rebased and signed-off the commit and run `make update-csv` as requested. Let me know if there's anything else I need to do here. Thanks!

PR #321, opened 2023-11-28
:sparkles: supporting cloudevents for work agent

skeeey:
## Related issue(s)

Fixes #


pkg/work/spoke/spokeagent.go
```diff
@@ -145,49 +139,101 @@ func (o *WorkAgentConfig) RunWorkloadAgent(ctx context.Context, controllerContex
...
 	go appliedManifestWorkController.Run(ctx, 1)
 	go manifestWorkController.Run(ctx, 1)
 	go manifestWorkFinalizeController.Run(ctx, manifestWorkFinalizeControllerWorkers)
 	go availableStatusController.Run(ctx, availableStatusControllerWorkers)
+
 	<-ctx.Done()
+
 	return nil
 }
+
```
qiujian16, line 191: it needs comment here. we should explain why we build like this.

skeeey: added

pkg/work/spoke/options.go
```diff
@@ -4,19 +4,22 @@ import (
...
 
 	"github.com/spf13/pflag"
+	"open-cluster-management.io/api/cloudevents/generic/options/mqtt"
 )
 
 // WorkloadAgentOptions defines the flags for workload agent
 type WorkloadAgentOptions struct {
 	StatusSyncInterval                     time.Duration
 	AppliedManifestWorkEvictionGracePeriod time.Duration
+	MQTTOptions                            *mqtt.MQTTOptions
```
clyang82, line 14: Do we need to start webhook server if it is configured to run with mqtt option? I think we may not. https://github.com/open-cluster-management-io/ocm/blob/8a4c834ebf0f3c71ab9209572fb0b11cc942f5d5/cmd/work/main.go#L62

clyang82:
How to test the MQTT protocol here?

pkg/work/spoke/options.go
skeeey: this is options only for agent，the webhook-server is a sub command for work agent, we can control whether to start it outside

```
./work --help
Spoke Cluster Work

Available Commands:
  agent          Start the Work Agent
  completion     Generate the autocompletion script for the specified shell
  help           Help about any command
  manager        Start the Work Hub Manager
  webhook-server Start the work webhook server
```

to use this `work agent --mqtt-broker=...`

skeeey:
> How to test the MQTT protocol here?

I'm not sure, the work client with MQTT is already tested in the api repo, but for work agent, now we don't have a server supporting this, currently, we may not be able to start the e2e test in the ocm, after the maestro server is ready, we may use it to do the e2e test in the ocm?

morvencao (commented):
Do we also need to add cloudevents support to the singleton agent ? https://github.com/morvencao/OCM/tree/main/pkg/singleton/spoke

skeeey:
> Do we also need to add cloudevents support to the singleton agent ? https://github.com/morvencao/OCM/tree/main/pkg/singleton/spoke

the singleton agent calls the work agent code, so this support will be also in the singleton agent

qiujian16:
TODO, instead of using flags. It will be better to have a config file for mqtt setup. We should only have one flag to specify driver for the work agent

skeeey:
> TODO, instead of using flags. It will be better to have a config file for mqtt setup. We should only have one flag to specify driver for the work agent

good suggestion, add the TODO to the code firstly :)

PR #328, opened 2023-12-06
:bug: migration-controller depending on cluster-manager condition.

xuezhaojun:
## Changes

1. Extract `parseStorageVersionMigrationFiles` and put it at the beginning of the reconcile.
    1. motivation: the `removeStorageVersionMigrations` , `applyStorageVersionMigrations`, `syncStorageVersionMigrationsCondition` all have the parse files process, we should extract them into one place.
2. Move `supportStorageVersionMigration` before `removeStorageVersionMigrations` 
    1. motivation: the `supported` is the base of every StorageVersionMigration related process, it should be done at the beginning. 
3. Change `applyStorageVersionMigrations` to `createStorageVersionMigrations` , add CRD check and remove the `update` part
    1. motivation: If the current CRD storage version equals to the version base set in the storageversionmigration, then no need to create the svm instance, because if will still be restored using the same version.
    2. motivation: the `storageversionmigration` is a create-only job-style object, the version field update won’t trigger another round of migration.
4. Remove `testMigrationRequestFiles` and any code locked with specific CRD types. Using `foo` and `bar` instead.
    1. motivation: We want to create test cases that won’t be affected by actual different migrationfiles cross releases.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
+++ b/pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
@@ -3,6 +3,7 @@ package migrationcontroller
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/openshift/library-go/pkg/assets"
 	"github.com/openshift/library-go/pkg/controller/factory"
@@ -56,6 +57,7 @@ const (
 	MigrationSucceeded    = "MigrationSucceeded"
 
 	migrationRequestCRDName = "storageversionmigrations.migration.k8s.io"
+	reSyncTime              = time.Second * 5
 )
 
 type crdMigrationController struct {
@@ -65,6 +67,7 @@ type crdMigrationController struct {
 	clusterManagerLister      operatorlister.ClusterManagerLister
 	recorder                  events.Recorder
 	generateHubClusterClients func(hubConfig *rest.Config) (apiextensionsclient.Interface, migrationv1alpha1client.StorageVersionMigrationsGetter, error)
+	parseMigrations           func() ([]*migrationv1alpha1.StorageVersionMigration, error)
 }
 
 // NewCRDMigrationController construct crd migration controller
@@ -81,6 +84,7 @@ func NewCRDMigrationController(
 			*operatorapiv1.ClusterManager, operatorapiv1.ClusterManagerSpec, operatorapiv1.ClusterManagerStatus](
 			clusterManagerClient),
 		clusterManagerLister:      clusterManagerInformer.Lister(),
+		parseMigrations:           parseStorageVersionMigrationFiles,
 		recorder:                  recorder,
 		generateHubClusterClients: generateHubClients,
 	}
@@ -94,7 +98,12 @@ func (c *crdMigrationController) sync(ctx context.Context, controllerContext fac
 	clusterManagerName := controllerContext.QueueKey()
 	klog.V(4).Infof("Reconciling ClusterManager %q", clusterManagerName)
 
-	if len(migrationRequestFiles) == 0 {
+	// if no migration files exist, do nothing and exit the reconcile
+	migrations, err := c.parseMigrations()
+	if err != nil {
+		return err
+	}
+	if len(migrations) == 0 {
 		return nil
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
```diff
@@ -188,47 +195,52 @@ func supportStorageVersionMigration(ctx context.Context, apiExtensionClient apie
...
+		for _, version := range crd.Spec.Versions {
+			if version.Name == migration.Spec.Resource.Version && version.Storage {
+				storageVersion = version.Name // find the current storage version of the CRD
+				break
+			}
+		}
+		if storageVersion == migration.Spec.Resource.Version {
+			errs = append(errs, fmt.Errorf("the current storage version of %v is %v, which is the same as the version in the migration CR %v",
+				resourceToCRDName(migration.Spec.Resource.Resource, migration.Spec.Resource.Group),
+				storageVersion, migration.Name))
```
qiujian16, line 239: if we continue here, how does the controller get triggered after crd is updated later?

qiujian16: I am thinking you would want to requeue instead of error fallback.

qiujian16: right, I think we probably should requeue under these conditions, or we need to watch crds.

pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
```diff
@@ -140,24 +141,38 @@ func (c *crdMigrationController) sync(ctx context.Context, controllerContext fac
...
 	if applied := meta.IsStatusConditionTrue(clusterManager.Status.Conditions, clusterManagerApplied); !applied {
 		controllerContext.Queue().AddRateLimited(clusterManagerName)
 		return nil
 	}
 
-	err = applyStorageVersionMigrations(ctx, migrationClient, c.recorder)
+	err = checkCRDStorageVersion(ctx, migrations, apiExtensionClient)
+	if err != nil {
+		klog.Errorf("Failed to check CRD current storage version. %v", err)
+		controllerContext.Queue().Add(clusterManagerName)
```
qiujian16, line 158: AddAfter for a certain period. You would not want to requeue immediately.

qiujian16:
/approve

@elgnay would you take a look pls?

pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
```diff
@@ -188,75 +225,103 @@ func supportStorageVersionMigration(ctx context.Context, apiExtensionClient apie
...
 	}
 	return nil
 }
 
-func applyStorageVersionMigrations(ctx context.Context,
-	migrationClient migrationv1alpha1client.StorageVersionMigrationsGetter, recorder events.Recorder) error {
+// 1.The CRD must exists before the migration CR is created.
+// 2.The CRD must have at least 2 version.
+// 3.The version set in the migration CR must exist in the CRD.
+// 4.The currrent storage vesion in CRD should not be the version in the migration CR, otherwise during the migration, the
```
xuezhaojun, line 245: This requires we following this practice in the following upgrade flow: when create a StorageVersionMigration, the version must be the "migration from" version, not the "migration to" version. @qiujian16 

cc @elgnay

xuezhaojun:
@elgnay  PR update:
1. Add more check step for CRD: a. CRD must have at lease 2 version; b. the version in migration CR must be included in the CRD
2. Update the condition when checkCRD or createMigration fails.
3. Change to AddRateLimit.

@qiujian16 In the last offline discussion with YangLe, he mentioned the point that StorageVersionMigration doesn't limit the `version` in its spec must be the "migration from" version, the `version` in its spec could be any versions in the CRD.

That means we have to make it a convention of the upgrade flow, more details see: https://github.com/open-cluster-management-io/ocm/pull/328/files#r1424801861

xuezhaojun:
/assign @elgnay 
Please take another review when you have time, thanks!

PR #330, opened 2023-12-07
:seedling: Remove unnecessary code for removing managed cluster resources

xuezhaojun:
## Summary

When hubAcceptsClient toggle between `true` and `false`, we want to keep the resources.

The resources we want to keep are:

```go
        // Hub cluster-admin accepts the spoke cluster, we apply
	// 1. clusterrole and clusterrolebinding for this spoke cluster.
	// 2. namespace for this spoke cluster.
	// 3. role and rolebinding for this spoke cluster on its namespace.
	resourceResults := c.applier.Apply(
		ctx,
		syncCtx.Recorder(),
		helpers.ManagedClusterAssetFn(manifests.RBACManifests, managedClusterName),
		staticFiles...,
	)
```

If the namespace of the spoke cluster removed, then all manifestwork applied in that namespace will also gone. We want to avoid that.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/hub/managedcluster/controller.go
+++ b/pkg/registration/hub/managedcluster/controller.go
@@ -108,12 +108,9 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 			return nil
 		}
 
-		// Hub cluster-admin denies the current spoke cluster, we remove its related resources and update its condition.
+		// Hub cluster-admin denies the current spoke cluster, update its condition.
+		// The resources applied on this spoke will keep there.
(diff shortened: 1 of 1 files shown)


qiujian16:
hrm I do not thin we remove ns right now... @skeeey ?

skeeey:
we will keep the ns when a managed cluster is denied, https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/managedcluster/controller.go#L31

skeeey:
we may, but it's difficult to handle the namespace, because there are too many cases we need consider (like the acm import controller does)

PR #342, opened 2024-01-09
:bug: Fix: using ownerRef to control the removal of storageversionmigrations.

xuezhaojun:
## Summary

When create a StorageVersionMigration, using the cluster-manager as the ownerReference. When cluster-manager is removed, all Migrations will also be removed.

## Related issue(s)

Previously, it uses a remove function that depending on:

https://github.com/open-cluster-management-io/ocm/blob/40135fd4baf762630ffc161b6341837081dc05c0/pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go#L52

From release to release the files list are changed, but during the upgrade the cluster-manager keep exisitng, so that old storageversionmigration CRs won't get removed ever.

Fixes #

### DIFF
--- a/pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
+++ b/pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go
@@ -143,11 +143,6 @@ func (c *crdMigrationController) sync(ctx context.Context, controllerContext fac
 		return err
 	}
 
-	// if the ClusterManager is deleting, we remove its related resources on hub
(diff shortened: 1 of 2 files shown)


qiujian16:
but the clustermanager will not be removed during upgrade, only uninstall will remove cluster manager. Is it what you are expecting?

xuezhaojun:
> but the clustermanager will not be removed during upgrade, only uninstall will remove cluster manager. Is it what you are expecting?

Yes, keep them won't harm anything.  This PR just make sure when we uninstall clustermanager, everything we installed will be cleaned up.

PR #343, opened 2024-01-11
:bug: Fix: move Enable ReplicaSet at the beginning and add more logs.

xuezhaojun:
## Summary

Extend the timeout of manifestworkreplicaset e2e cases to aviod flake error, also add more log for debugging. 

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/manifestworkreplicaset_test.go
+++ b/test/e2e/manifestworkreplicaset_test.go
@@ -31,19 +31,6 @@ var _ = ginkgo.Describe("Test ManifestWorkReplicaSet", func() {
 
 	ginkgo.BeforeEach(func() {
 		nameSuffix = rand.String(6)
-		// Enable manifestWorkReplicaSet feature if not enabled
-		gomega.Eventually(func() error {
-			return t.EnableWorkFeature("ManifestWorkReplicaSet")
-		}, t.EventuallyTimeout*5, t.EventuallyInterval*5).Should(gomega.Succeed())
-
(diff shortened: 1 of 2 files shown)


test/e2e/manifestworkreplicaset_test.go
```diff
@@ -303,46 +303,55 @@ var _ = ginkgo.Describe("Test ManifestWorkReplicaSet", func() {
...
+					return err
 				}
 
-				return mwrSet.Status.Summary.Total == numOfClusters
-			}, t.EventuallyTimeout*5, t.EventuallyInterval*5).Should(gomega.BeTrue())
+				if mwReplicaSet.Status.Summary.Total != numOfClusters {
+					return fmt.Errorf("total number of clusters is not correct, expect %d, got %d", numOfClusters, mwReplicaSet.Status.Summary.Total)
+				}
+				return nil
+			}, t.EventuallyTimeout*10, t.EventuallyInterval*10).Should(gomega.Succeed())
```
qiujian16, line 320: this seems really long...why do we need 10 times?

qiujian16 (commented):
/approve
/lgtm

test/e2e/manifestworkreplicaset_test.go
xuezhaojun: Yes, but think in another way, we can move the **Enable** and **Disable** the **ReplicaSet** at the beginning to save some time.

PR #345, opened 2024-01-17
:bug: Remove codecov token.

xuezhaojun:
## Summary

Recently, the codecov push keeps fail, through test we know the token is not required for public repos, so we can remove the token to fix that failure, the approach is verified in cluster-proxy-addon repo: https://github.com/open-cluster-management-io/cluster-proxy/pull/174

Also the OCM_BOT_TOKEN also shows "Bad credentials" when we try to trigger clusteradmin e2e test after PR merged, can we have a check on that token in case it's out-of-date? @qiujian16 

## Related issue(s)

Fixes #

https://github.com/open-cluster-management-io/ocm/actions/runs/7497318242/job/20410721816

### DIFF
--- a/.github/workflows/post.yml
+++ b/.github/workflows/post.yml
@@ -35,7 +35,6 @@ jobs:
       - name: report coverage
         uses: codecov/codecov-action@v3
         with:
-          token: ${{ secrets.CODECOV_UPLOAD_TOKEN }}
           files: ./coverage.out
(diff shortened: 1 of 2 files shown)


xuezhaojun:
/assign @qiujian16 

No code change, just remove the codecov token to avoid bad cridetials issue now and in the future.

PR #351, opened 2024-01-19
✨ Configurable qos resources for containers managed by cluster-manager and klusterlet

promid:
## Summary
Configurable qos resources for containers managed by cluster-manager and klusterlet
## Related issue(s)

Fixes #

## Related PR
https://github.com/open-cluster-management-io/api/pull/316

### DIFF
--- a/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller_test.go
+++ b/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller_test.go
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ghodss/yaml"
+	"github.com/openshift/library-go/pkg/assets"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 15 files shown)


promid:
/hold
Pending https://github.com/open-cluster-management-io/api/pull/316 this PR

pkg/common/helpers/clusters_test.go
```diff
@@ -1,114 +0,0 @@
-package helpers
```
haoqing0110, line 1: The `PlacementDecisionGetter` is being used by other controllers, `GetClusterChanges` and the clusters_test.go is used to test the `PlacementDecisionGetter` I think.

promid: I removed this because functions used in pkg/common/helpers/clusters.go are not found after I updated open-cluster-management.io/api to the latest commit in go.mod
Confirmed with @qiujian16, the `not found` issue will be fixed by https://github.com/open-cluster-management-io/addon-framework/pull/232.
I will update this PR then after the above PR is merged.

haoqing0110: @promid get it!

pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller_test.go
```diff
@@ -1051,6 +1055,90 @@ func TestDeployOnKube111(t *testing.T) {
...
+		req := corev1.ResourceRequirements{}
+		actualRequests := req.String()
+		if actualResources != actualRequests {
+			t.Errorf("expect:\n%s\nbut got:\n%s", actualRequests, actualResources)
+		}
+	}
+}
+
+func TestRenderingResourceRequirements(t *testing.T) {
+	r := &operatorapiv1.ResourceRequirement{
```
qiujian16, line 1110: seems to me the three tests can be put in one test func with different cases. It also worth a test case with type is ResourceRequirement and ResourceRequirements is nil or empty

promid: merged into one unit test with 3 cases

haoqing0110:
LGTM

promid:
executed `make fmt-imports`

qiujian16:
/approve
/lgtm

I am fine to merge this, I think we also need to add an integration test case as a followup

PR #352, opened 2024-01-19
:sparkles: support cloudevents for manifestworkreplicaset

skeeey:
## Related issue(s)

Fixes #


pkg/work/hub/manager.go
```diff
@@ -2,73 +2,166 @@ package hub
...
 )
 
-// RunWorkHubManager starts the controllers on hub.
-func RunWorkHubManager(ctx context.Context, controllerContext *controllercmd.ControllerContext) error {
-	hubWorkClient, err := workclientset.NewForConfig(controllerContext.KubeConfig)
-	if err != nil {
-		return err
+const (
+	sourceID       = "mwrsctrl"
+	sourceClientID = "mwrsctrl-client"
```
morvencao, line 29: nit: Is this safe to have the same hard-code clientID for mwrs controller?
does this mean anytime, only one mwrs controller instance is working?

skeeey: maybe need to expose these with flags

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
```diff
@@ -46,6 +46,7 @@ spec:
         args:
           - "/work"
           - "manager"
+          - "--work-driver=kube"
```
morvencao, line 49: add `--cloudevents-client-id=$(POD_NAME)`?

skeeey: currently, we make the `cloudevents-client-id` optional, it only need when we use cloudevents, current case is kube, we can omit it

skeeey:
/hold

the sdk-go has a bug, I will fix it, then update the sdk-go again

PR #372, opened 2024-03-11
:bug: Fix mode replacement for hosted mode

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-10392

## Summary

The default mode was set to `Singleton`, so the substitution command needs to also be updated to reflect the new default.

## Related issue(s)

- #365


qiujian16 (commented):
/approve
/lgtm

PR #375, opened 2024-03-12
WIP :bug: Fix: make kube a default value of workloadsourcedriver.

xuezhaojun:
## Summary

We want use kube as the default driver even no driver set.

### DIFF
--- a/pkg/work/spoke/options.go
+++ b/pkg/work/spoke/options.go
@@ -4,6 +4,8 @@ import (
 	"time"
 
 	"github.com/spf13/pflag"
+
+	"open-cluster-management.io/sdk-go/pkg/cloudevents/work"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/work/spoke/options.go
```diff
@@ -28,6 +30,7 @@ func NewWorkloadAgentOptions() *WorkloadAgentOptions {
 		MaxJSONRawLength:                       1024,
 		StatusSyncInterval:                     10 * time.Second,
 		AppliedManifestWorkEvictionGracePeriod: 60 * time.Minute,
+		WorkloadSourceDriver:                   work.ConfigTypeKube,
```
qiujian16, line 33: also need to default the kubcoonfig file path

xuezhaojun: You're right, added.

xuezhaojun:
Fixed by another PR, close
/close

PR #377, opened 2024-03-13
:sparkles: Retrigger CSR when certs doesn't has desired org and ou of subject(registrationConfig).

xuezhaojun:
## Summary

Before: the CSR only be retriggerred when subject.CommonName doesn't match.

Now: if there is no certs match subject (commonnam + org + ou) , CSR get retriggered.

### DIFF
--- a/pkg/registration/clientcert/certificate.go
+++ b/pkg/registration/clientcert/certificate.go
@@ -2,9 +2,11 @@ package clientcert
 
 import (
 	"context"
+	"crypto/x509"
 	"crypto/x509/pkix"
 	"errors"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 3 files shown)


pkg/registration/clientcert/certificate.go
```diff
@@ -92,17 +93,69 @@ func IsCertificateValid(logger klog.Logger, certData []byte, subject *pkix.Name)
...
 }
 
+func certMatchSubject(logger klog.Logger, cert *x509.Certificate, subject *pkix.Name) bool {
+	// check commonName
+	if cert.Subject.CommonName != subject.CommonName {
+		return false
+	}
+
+	// check groups(origanization)
+	if len(subject.Organization) > 0 {
```
qiujian16, line 115: why not just deepEqual?

xuezhaojun: Agree, updated.

qiujian16 (commented):
/approve
/lgtm

PR #378, opened 2024-03-14
🐛 Fix the valid client certificate check

mprahl:
## Summary

The commit edef33de92cd9af45709f24f8b35d4e4e96af822 introduced additional checks on the certificate subject. The issue is that the order of the groups and organizational units are not consistent, so it was causing the addon-framework to generate several CSRs with many coming back as invalid.

This takes the approach of sorting the strings before doing the comparison.

### DIFF
--- a/pkg/registration/clientcert/certificate.go
+++ b/pkg/registration/clientcert/certificate.go
@@ -6,7 +6,6 @@ import (
 	"crypto/x509/pkix"
 	"errors"
 	"fmt"
-	"reflect"
 	"time"
 
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -14,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	certificatesinformers "k8s.io/client-go/informers/certificates"
 	certificatesv1informers "k8s.io/client-go/informers/certificates/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


qiujian16:
Do you have an idea why the order is not consistent given it is a slice? Alternatively, I would suggest to use Set[string].

mprahl:
> Do you have an idea why the order is not consistent given it is a slice? Alternatively, I would suggest to use Set[string].

I don't know why it's not ordered but I can try and figure that out.

Thanks for the suggestion! I didn't know apimachinery included a set library.

mprahl:
> Do you have an idea why the order is not consistent given it is a slice? Alternatively, I would suggest to use Set[string].

It seems that `x509.CreateCertificateRequest` uses a set under the hood. You can see this simple example of how the parsed CSR returns the organizations as sorted:

```go
package main

import (
	"crypto/ecdsa"
	"crypto/elliptic"
	"crypto/x509"
	"crypto/x509/pkix"
	"fmt"
	"math/rand"
	"net"
)

func main() {
	insecureRand := rand.New(rand.NewSource(0)) //nolint:gosec
	pk, err := ecdsa.GenerateKey(elliptic.P256(), insecureRand)
	if err != nil {
		panic(err)
	}

	csrb, err := x509.CreateCertificateRequest(insecureRand, &x509.CertificateRequest{
		Subject: pkix.Name{
			CommonName: "system:open-cluster-management:cluster:cluster1:addon:config-policy-controller:agent:config-policy-controller",
			Organization: []string{
				"system:open-cluster-management:cluster:cluster1:addon:config-policy-controller",
				"system:open-cluster-management:addon:config-policy-controller",
				"system:authenticated",
			},
		},
		DNSNames:       []string{},
		EmailAddresses: []string{},
		IPAddresses:    []net.IP{},
	}, pk)
	if err != nil {
		panic(err)
	}

	csrParsed, err := x509.ParseCertificateRequest(csrb)
	if err != nil {
		panic(err)
	}

	fmt.Println(csrParsed.Subject.Organization)
}
```


The output is:

```bash
/tmp/temp-csr via 🐹 v1.22.0 ❯ go run main.go
[system:authenticated system:open-cluster-management:addon:config-policy-controller system:open-cluster-management:cluster:cluster1:addon:config-policy-controller]
```

qiujian16:
ah thanks for the info.

/approve
/lgtm

PR #381, opened 2024-03-19
✨ support work driver config for cluster manager.

morvencao:
## Summary

This PR adds the capability to configure the work driver for the work controller via the cluster manager.

## Related issue(s)

https://issues.redhat.com/browse/ACM-10422

### DIFF
--- a/manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
+++ b/manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
@@ -46,7 +46,13 @@ spec:
         args:
           - "/work"
           - "manager"
-          - "--work-driver=kube"
+          {{ if .CloudEventsDriverEnabled }}
+          - "--work-driver={{ .WorkDriver }}"
+          {{ if ne .WorkDriver "kube" }}
+          - "--cloudevents-client-id=work-controller-$(POD_NAME)"
+          - "--work-driver-config=/var/run/secrets/work/config.yaml"
+          {{ end }}
+          {{ end }}
           {{ if .HostedMode }}
           - "--kubeconfig=/var/run/secrets/hub/kubeconfig"
           {{ end }}
@@ -97,6 +103,11 @@ spec:
           name: kubeconfig
           readOnly: true
         {{ end }}
+        {{ if and .CloudEventsDriverEnabled (ne .WorkDriver "kube") }}
+        - mountPath: /var/run/secrets/work
+          name: workdriverconfig
+          readOnly: true
+        {{ end }}
       volumes:
       - name: tmpdir
         emptyDir: { }
@@ -104,4 +115,9 @@ spec:
       - name: kubeconfig
         secret:
           secretName: work-controller-sa-kubeconfig
+      {{ end }}
+      {{ if and .CloudEventsDriverEnabled (ne .WorkDriver "kube") }}
+      - name: workdriverconfig
+        secret:
+          secretName: work-driver-config
       {{ end }}
--- a/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
+++ b/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
@@ -58,7 +58,8 @@ type clusterManagerController struct {
 		mwctrEnabled, addonManagerEnabled bool) error
 	generateHubClusterClients func(hubConfig *rest.Config) (kubernetes.Interface, apiextensionsclient.Interface,
 		migrationclient.StorageVersionMigrationsGetter, error)
-	skipRemoveCRDs bool
+	skipRemoveCRDs    bool
+	operatorNamespace string
 }
 
 type clusterManagerReconcile interface {
@@ -83,6 +84,7 @@ func NewClusterManagerController(
 	configMapInformer corev1informers.ConfigMapInformer,
 	recorder events.Recorder,
 	skipRemoveCRDs bool,
+	operatorNamespace string,
 ) factory.Controller {
 	controller := &clusterManagerController{
 		operatorKubeClient: operatorKubeClient,
@@ -97,6 +99,7 @@ func NewClusterManagerController(
 		ensureSAKubeconfigs:       ensureSAKubeconfigs,
 		cache:                     resourceapply.NewResourceCache(),
 		skipRemoveCRDs:            skipRemoveCRDs,
+		operatorNamespace:         operatorNamespace,
 	}
 
 	return factory.New().WithSync(controller.sync).
@@ -132,6 +135,12 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 		return err
 	}
 
+	// default driver is kube
+	workDriver := operatorapiv1.WorkDriverTypeKube
+	if clusterManager.Spec.WorkConfiguration != nil && clusterManager.Spec.WorkConfiguration.WorkDriver != "" {
+		workDriver = clusterManager.Spec.WorkConfiguration.WorkDriver
+	}
+
 	// This config is used to render template of manifests.
 	config := manifests.HubConfig{
 		ClusterManagerName:      clusterManager.Name,
@@ -150,6 +159,7 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 		},
 		ResourceRequirementResourceType: helpers.ResourceType(clusterManager),
 		ResourceRequirements:            resourceRequirements,
+		WorkDriver:                      string(workDriver),
 	}
 
 	var registrationFeatureMsgs, workFeatureMsgs, addonFeatureMsgs string
@@ -169,6 +179,7 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	}
 	config.WorkFeatureGates, workFeatureMsgs = helpers.ConvertToFeatureGateFlags("Work", workFeatureGates, ocmfeature.DefaultHubWorkFeatureGates)
 	config.MWReplicaSetEnabled = helpers.FeatureGateEnabled(workFeatureGates, ocmfeature.DefaultHubWorkFeatureGates, ocmfeature.ManifestWorkReplicaSet)
+	config.CloudEventsDriverEnabled = helpers.FeatureGateEnabled(workFeatureGates, ocmfeature.DefaultHubWorkFeatureGates, ocmfeature.CloudEventsDrivers)
 
 	var addonFeatureGates []operatorapiv1.FeatureGate
 	if clusterManager.Spec.AddOnManagerConfiguration != nil {
@@ -211,6 +222,8 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	reconcilers := []clusterManagerReconcile{
 		&crdReconcile{cache: n.cache, recorder: n.recorder, hubAPIExtensionClient: hubApiExtensionClient,
 			hubMigrationClient: hubMigrationClient, skipRemoveCRDs: n.skipRemoveCRDs},
+		&secretReconcile{cache: n.cache, recorder: n.recorder, operatorKubeClient: n.operatorKubeClient,
+			hubKubeClient: hubClient, operatorNamespace: n.operatorNamespace},
 		&hubReoncile{cache: n.cache, recorder: n.recorder, hubKubeClient: hubClient},
 		&runtimeReconcile{cache: n.cache, recorder: n.recorder, hubKubeConfig: hubKubeConfig, hubKubeClient: hubClient,
 			kubeClient: managementClient, ensureSAKubeconfigs: n.ensureSAKubeconfigs},
--- a/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller_test.go
+++ b/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller_test.go
@@ -31,6 +31,7 @@ import (
 
 	fakeoperatorlient "open-cluster-management.io/api/client/operator/clientset/versioned/fake"
 	operatorinformers "open-cluster-management.io/api/client/operator/informers/externalversions"
+	ocmfeature "open-cluster-management.io/api/feature"
 	operatorapiv1 "open-cluster-management.io/api/operator/v1"
 	"open-cluster-management.io/sdk-go/pkg/patcher"
 
@@ -75,6 +76,7 @@ func newClusterManager(name string) *operatorapiv1.ClusterManager {
 			},
 			WorkConfiguration: &operatorapiv1.WorkConfiguration{
 				FeatureGates: []operatorapiv1.FeatureGate{featureGate},
+				WorkDriver:   operatorapiv1.WorkDriverTypeKube,
 			},
 		},
 	}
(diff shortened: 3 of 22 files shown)


pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -209,6 +216,17 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	}
 	managementClient := n.operatorKubeClient // We assume that operator is always running on the management cluster.
 
+	// If the work driver is not kube, we need to sync the work driver config secret
+	if workDriver != operatorapiv1.WorkDriverTypeKube {
+		if err := helpers.SyncWorkConfigSecret(ctx, n.operatorKubeClient, clusterManagerNamespace); err != nil {
```
qiujian16, line 221: I think there is already a sync secret func in library-go

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -209,6 +216,17 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	}
 	managementClient := n.operatorKubeClient // We assume that operator is always running on the management cluster.
 
+	// If the work driver is not kube, we need to sync the work driver config secret
+	if workDriver != operatorapiv1.WorkDriverTypeKube {
```
qiujian16, line 220: what happens if clustermanager is create at first and then secret is created?

morvencao: The operator will initially error, but subsequent reconcile will succeed after creating the secret.
Currently we don't watch the secret, do we need that?

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
morvencao: replaced this with SyncSecret from library-go.

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
```diff
@@ -46,10 +46,15 @@ spec:
         args:
           - "/work"
           - "manager"
-          - "--work-driver=kube"
+          - "--work-driver={{ .WorkDriver }}"
+          {{ if eq .WorkDriver "kube" }}
```
zhiweiyin318, line 50: should be like this ?
```
{{ if .HostedMode }}
{{ if eq .WorkDriver "kube" }}
```
hostedMode only in the case` workDirver == kube`?

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
```diff
@@ -92,16 +97,28 @@ spec:
         volumeMounts:
         - name: tmpdir
           mountPath: /tmp
+        {{ if eq .WorkDriver "kube" }}
```
zhiweiyin318, line 100: same question: 
hostedMode only in the case `workDirver == kube`?

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
morvencao: Currently hosted mode only apply in kube work driver.

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
morvencao: same as above.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
qiujian16: I think we need to reflect that the secret is missing.

morvencao: added condition to clustermanager CR's status to reflect that.

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
```diff
@@ -46,10 +46,15 @@ spec:
...
           - "/work"
           - "manager"
-          - "--work-driver=kube"
+          - "--work-driver={{ .WorkDriver }}"
+          {{ if eq .WorkDriver "kube" }}
           {{ if .HostedMode }}
           - "--kubeconfig=/var/run/secrets/hub/kubeconfig"
           {{ end }}
+          {{ else }}
+          - "--cloudevents-client-id=$(POD_NAME)"
```
qiujian16, line 55: not sure if using podname is good choice. The pod name is changed after upgrade. Should the client id be already in the config file?

pkg/operator/helpers/helpers.go
```diff
@@ -724,6 +725,39 @@ func GetHubKubeconfig(ctx context.Context,
...
+	err := operatorClient.CoreV1().Secrets(clusterManagerNamespace).Delete(ctx, WorkDriverConfig, metav1.DeleteOptions{})
+	if err != nil && errors.IsNotFound(err) {
+		return nil
+	}
+
+	return err
+}
+
+// GetOperatorNamespace is used to get the namespace where the operator is running.
+func GetOperatorNamespace() (string, error) {
```
qiujian16, line 748: I thought we already get the namespace from the controllerContext

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -209,6 +217,71 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
...
+				})
+				if conditionChanged {
+					clusterManager.Status.ObservedGeneration = clusterManager.Generation
+					_, err := n.patcher.PatchStatus(ctx, clusterManager, clusterManager.Status, originalClusterManager.Status)
+					if err != nil {
+						return err
+					}
+				}
+			}
+		} else {
```
qiujian16, line 269: why we need to remove it? The else section here seems unnecessary to me. we can just return a condition with true when kube-driver is set.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -209,6 +217,71 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
...
+					if conditionChanged {
+						clusterManager.Status.ObservedGeneration = clusterManager.Generation
+						_, err := n.patcher.PatchStatus(ctx, clusterManager, clusterManager.Status, originalClusterManager.Status)
+						if err != nil {
+							return err
+						}
+					}
+				}
+				return err
+			} else {
```
qiujian16, line 248: to many nested if. should refactor the code here. 
Wouldn't SyncSecret return a notFound error? I think we do not need to get then sync.

directly sync and return error should be ok.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -209,6 +217,71 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	}
 	managementClient := n.operatorKubeClient // We assume that operator is always running on the management cluster.
 
+	if config.MWReplicaSetEnabled {
```
qiujian16, line 220: this if is not necessary, we should alway sync.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -38,9 +38,10 @@ import (
 )
 
 const (
-	clusterManagerFinalizer   = "operator.open-cluster-management.io/cluster-manager-cleanup"
-	clusterManagerApplied     = "Applied"
-	clusterManagerProgressing = "Progressing"
+	clusterManagerFinalizer      = "operator.open-cluster-management.io/cluster-manager-cleanup"
+	clusterManagerApplied        = "Applied"
+	clusterManagerProgressing    = "Progressing"
+	workDriverConfigSecretSynced = "WorkDriverConfigSecretSynced"
```
qiujian16, line 44: maybe just secret synced. We will need to also sync other secret for image pull credential later. These could be in the same condition, different reasons and messages.

FYI @zhiweiyin318

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
morvencao: the client id is not in config file, it's in cloudevents source options, see: https://github.com/open-cluster-management-io/sdk-go/blob/7dbdd1b5c63da7b832705501185138a5e5c530ba/pkg/cloudevents/generic/options/mqtt/sourceoptions.go#L21

per my understanding, it's acceptable if the client is changed, since it's used by MQTT broker to differentiate clients, the key is to make sure each publisher/subscriber has different client id.

pkg/operator/helpers/helpers.go
morvencao: the work-driver-config secret is supposed to be created in the namespace (`open-cluster-management` by default) as the operator, not in the operand namespace (`open-cluster-management-hub` by default), because the operand namespace may not exist before clustermanager is created.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
morvencao: SyncSecret doesn't return an error when the source secret doesn't exist, that's why I added the check before SyncSecret. Also added comments at https://github.com/open-cluster-management-io/ocm/blob/a559f7eeb1ec7bbf4041ceed029cafbddd20d796/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go#L227-L229

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
morvencao: updated this to a general value: "SecretSynced"

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
morvencao: removed this.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller_test.go
```diff
@@ -75,6 +75,7 @@ func newClusterManager(name string) *operatorapiv1.ClusterManager {
 			},
 			WorkConfiguration: &operatorapiv1.WorkConfiguration{
 				FeatureGates: []operatorapiv1.FeatureGate{featureGate},
+				WorkDriver:   operatorapiv1.WorkDriverTypeKube,
```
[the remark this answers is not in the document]
morvencao: added UT for sync work driver config secret and condition check.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
morvencao: This is used for case: non-kube driver is changed to `kube` driver, the condition should be removed, make sense?

qiujian16: I think we can always add the condition.

pkg/operator/helpers/helpers.go
qiujian16: yes, but you can get the namespace from controllerContext which get the namespace the same as what this func does.

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
qiujian16: I am more concerned about if client id is part of authentication/authz, which will be impacted by the change of client id. cc @skeeey

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
morvencao: done, removed this part.

pkg/operator/helpers/helpers.go
morvencao: oh... I see, it's in command ControllerContext: https://github.com/openshift/library-go/blob/0f907b20fed14da2731dbb6a537e97c691579dc7/pkg/controller/controllercmd/builder.go#L41-L60 
updated this.

manifests/cluster-manager/management/cluster-manager-manifestworkreplicaset-deployment.yaml
skeeey: I remembered for some MQTT broker, like the AWS IOT, we need this client ID do authn/authz, but we can use a wildcard in the  authn/authz, I think you may need a fixed format ID here

morvencao: thanks, I updated this to `"--cloudevents-client-id=work-controller-$(POD_NAME)"` with fixed prefix `work-controller-`, so in the future the authn/zuthz can be donw with wildcard `work-controller-*`.

morvencao:
rebased the code.

morvencao:
added feature gate for cloudevents driver so that it doesn't break the current default behaviour.

morvencao:
PATL @qiujian16 @skeeey

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -207,6 +218,26 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
...
+			helpers.WorkDriverConfig,
+			clusterManagerNamespace,
+			helpers.WorkDriverConfig,
+			[]metav1.OwnerReference{},
+		)
+		if err != nil {
+			return err
+		}
+		// TODO: set condition to indicate if the secret is synced successfully
+		// check the source secret exist before syncing and set condition if not exist.
```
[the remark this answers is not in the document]
morvencao: done, added a new secret reconciler.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_secret_reconcile.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (c *secretReconcile) reconcile(ctx context.Context, cm *operatorapiv1.ClusterManager,
+	config manifests.HubConfig) (*operatorapiv1.ClusterManager, reconcileState, error) {
+	if config.CloudEventsDriverEnabled && config.WorkDriver != string(operatorapiv1.WorkDriverTypeKube) {
+		// check the source secret explicitly as the 'helpers.SyncSecret' doesn't return an error
+		// when the source secret is not found.
+		if _, err := c.operatorKubeClient.CoreV1().Secrets(c.operatorNamespace).Get(ctx,
+			workDriverConfig, metav1.GetOptions{}); errors.IsNotFound(err) {
+			// TODO: set condition if the source secret doesn't exist,
+			return cm, reconcileStop,
```
zhiweiyin318, line 40: in helpers.SyncSecret, the synced secret will be deleted if the source secret is not found.
so if return here, the synced workDriverConfig secret will not be deleted if workDriverConfig is deleted.
not sure if this is expected ?

morvencao: the source secret should not be deleted per my understanding.

qiujian16: can we hav a empty secret slice, and append the "work-driver-config" secret if feature gate is enabled?

morvencao: maybe a secret set instead of a secret slice, to avoid duplicated resync.
done.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_secret_reconcile.go
```diff
@@ -0,0 +1,89 @@
...
+	if config.CloudEventsDriverEnabled && config.WorkDriver != string(operatorapiv1.WorkDriverTypeKube) {
+		secretNames = secretNames.Insert(workDriverConfig)
+	} else {
+		secretNames = secretNames.Delete(workDriverConfig)
+	}
+
+	for _, secretName := range secretNames.UnsortedList() {
+		// check the source secret explicitly as the 'helpers.SyncSecret' doesn't return an error
+		// when the source secret is not found.
+		if _, err := c.operatorKubeClient.CoreV1().Secrets(c.operatorNamespace).Get(ctx,
```
qiujian16, line 49: let's remove this additional get for now. It could happen this should return success if the secret does not exist

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_secret_reconcile.go
```diff
@@ -0,0 +1,89 @@
...
+			ctx,
+			c.operatorKubeClient.CoreV1(),
+			c.hubKubeClient.CoreV1(),
+			c.recorder,
+			c.operatorNamespace,
+			secretName,
+			config.ClusterManagerNamespace,
+			secretName,
+			[]metav1.OwnerReference{},
+		); err != nil {
```
qiujian16, line 67: actually, we should append the err to a err slice and return aggregated err finally.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_secret_reconcile.go
morvencao: removed this.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_secret_reconcile.go
morvencao: added aggregated error.

morvencao:
kindly ping @qiujian16

qiujian16 (commented):
/approve

PR #388, opened 2024-03-29
:bug: avoid cluster auto approve failed occasionally

### LINKED ISSUE #386: cluster auto approve could fail occasionally
**Describe the bug**
When enable the AutoApprove feature gate, sometimes the cluster will not be get accepted by the hub due to timing issue.

**To Reproduce**
enable the AutoApprove feature gate with `clusteradmin init` and run `clusteradm join`.  It happens occasionally.

**Expected behavior**
The cluster should be always approved

latest

**Additional context**

skeeey:
## Summary
Remove the accept cluster logic from the CSR controller, and put them in the managed cluster controller  

## Related issue(s)

Fixes #386

### DIFF
--- a/pkg/registration/hub/csr/reconciler.go
+++ b/pkg/registration/hub/csr/reconciler.go
@@ -11,15 +11,11 @@ import (
 	authorizationv1 "k8s.io/api/authorization/v1"
 	certificatesv1 "k8s.io/api/certificates/v1"
 	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
-	clusterclientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
-	clusterv1listers "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 
 	"open-cluster-management.io/ocm/pkg/registration/hub/user"
@@ -95,21 +91,15 @@ func (r *csrRenewalReconciler) Reconcile(ctx context.Context, csr csrInfo, appro
 
 type csrBootstrapReconciler struct {
 	kubeClient    kubernetes.Interface
-	clusterClient clusterclientset.Interface
-	clusterLister clusterv1listers.ManagedClusterLister
 	approvalUsers sets.Set[string]
 	eventRecorder events.Recorder
 }
 
 func NewCSRBootstrapReconciler(kubeClient kubernetes.Interface,
-	clusterClient clusterclientset.Interface,
-	clusterLister clusterv1listers.ManagedClusterLister,
 	approvalUsers []string,
 	recorder events.Recorder) Reconciler {
 	return &csrBootstrapReconciler{
 		kubeClient:    kubeClient,
-		clusterClient: clusterClient,
-		clusterLister: clusterLister,
 		approvalUsers: sets.New(approvalUsers...),
 		eventRecorder: recorder.WithComponentSuffix("csr-approving-controller"),
 	}
@@ -129,37 +119,12 @@ func (b *csrBootstrapReconciler) Reconcile(ctx context.Context, csr csrInfo, app
 		return reconcileContinue, nil
 	}
 
-	err := b.accpetCluster(ctx, clusterName)
-	if errors.IsNotFound(err) {
-		// Current spoke cluster not found, could have been deleted, do nothing.
-		return reconcileStop, nil
-	}
-	if err != nil {
-		return reconcileContinue, err
-	}
-
 	if err := approveCSR(b.kubeClient); err != nil {
 		return reconcileContinue, err
 	}
 
 	b.eventRecorder.Eventf("ManagedClusterAutoApproved", "spoke cluster %q is auto approved.", clusterName)
 	return reconcileStop, nil
-}
-
-func (b *csrBootstrapReconciler) accpetCluster(ctx context.Context, managedClusterName string) error {
-	managedCluster, err := b.clusterLister.Get(managedClusterName)
-	if err != nil {
-		return err
-	}
-
-	if managedCluster.Spec.HubAcceptsClient {
-		return nil
-	}
-
-	patch := []byte("{\"spec\": {\"hubAcceptsClient\": true}}")
-	_, err = b.clusterClient.ClusterV1().ManagedClusters().Patch(
-		ctx, managedCluster.Name, types.MergePatchType, patch, metav1.PatchOptions{})
-	return err
 }
 
 // To validate a managed cluster csr, we check
--- a/pkg/registration/hub/managedcluster/controller.go
+++ b/pkg/registration/hub/managedcluster/controller.go
@@ -3,6 +3,7 @@ package managedcluster
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -12,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	rbacv1informers "k8s.io/client-go/informers/rbac/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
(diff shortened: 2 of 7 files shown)


pkg/registration/hub/csr/reconciler.go
```diff
@@ -129,24 +137,25 @@ func (b *csrBootstrapReconciler) Reconcile(ctx context.Context, csr csrInfo, app
 		return reconcileContinue, nil
 	}
 
-	err := b.accpetCluster(ctx, clusterName)
+	err := b.acceptCluster(ctx, clusterName)
```
nirs, line 140: Mixing typos fixes with a real change make it harder to review and make it hard to maintain the code later. If you revert this fix because it caused a regression, we will lose also the typo fixes.

pkg/registration/hub/csr/reconciler.go
```diff
@@ -129,24 +137,25 @@ func (b *csrBootstrapReconciler) Reconcile(ctx context.Context, csr csrInfo, app
...
 	if errors.IsNotFound(err) {
-		// Current spoke cluster not found, could have been deleted, do nothing.
-		return reconcileStop, nil
+		// Current spoke cluster not found, it may not be created yet, requeue it until we cannot find its related CSR
+		b.eventRecorder.Warningf("ManagedClusterNotApproved", "spoke cluster %q is not found", clusterName)
+		return reconcileRequeue, nil
```
nirs, line 144: So if we have a CSR but we cannot find the managed cluster we queue the request and will check again after some delay?

I expect that if this reconciler depends on a CSR and ManagedCluster, it will watch both resources. If the ManagedCluster is missing, you can do nothing, and once the ManagedCluster appears, you can approve the CSR.

pkg/registration/hub/csr/reconciler.go
```diff
@@ -129,24 +137,25 @@ func (b *csrBootstrapReconciler) Reconcile(ctx context.Context, csr csrInfo, app
 		return reconcileContinue, nil
 	}
 
-	err := b.accpetCluster(ctx, clusterName)
+	err := b.acceptCluster(ctx, clusterName)
 	if errors.IsNotFound(err) {
-		// Current spoke cluster not found, could have been deleted, do nothing.
-		return reconcileStop, nil
+		// Current spoke cluster not found, it may not be created yet, requeue it until we cannot find its related CSR
+		b.eventRecorder.Warningf("ManagedClusterNotApproved", "spoke cluster %q is not found", clusterName)
```
nirs, line 143: This warning will be useful!

pkg/registration/hub/csr/reconciler.go
```diff
@@ -129,24 +137,25 @@ func (b *csrBootstrapReconciler) Reconcile(ctx context.Context, csr csrInfo, app
...
-		return reconcileContinue, err
+		return reconcileRequeue, err
 	}
 
 	b.eventRecorder.Eventf("ManagedClusterAutoApproved", "spoke cluster %q is auto approved.", clusterName)
 	return reconcileStop, nil
 }
 
-func (b *csrBootstrapReconciler) accpetCluster(ctx context.Context, managedClusterName string) error {
+func (b *csrBootstrapReconciler) acceptCluster(ctx context.Context, managedClusterName string) error {
```
nirs, line 158: Another typo that should not be part of this fix.

nirs (commented):
How can I test this fix?

pkg/registration/hub/csr/reconciler.go
skeeey: good suggestion, I will drop the typos fixes in this pr

pkg/registration/hub/csr/reconciler.go
skeeey: I will use a separated controller to handle cluster, and make this csr controller only handle the CSR ,  this will be more clear

skeeey:
> How can I test this fix?

If you want to test with this pr, you can use my pr's repo to build your own images, then use `clusteradm init --image-registry` to specify the images

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -103,6 +110,15 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 	}
 
 	if !managedCluster.Spec.HubAcceptsClient {
+		// If the ManagedClusterAutoApproval feature is enabled, we automatically accept this cluster and
+		// add an annotation to mark its initial join as accepted.
```
nirs, line 114: The comment does is not clear about the purpose of the annotation.

Based on the new test, we want to approve the cluster once on the first time we discover it, so the cluster admin can deny the cluster. This should be explained in the comment.

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -199,3 +215,11 @@ func (c *managedClusterController) removeManagedClusterResources(ctx context.Con
 	}
 	return operatorhelpers.NewMultiLineAggregate(errs)
 }
+
+func (c *managedClusterController) acceptCluster(ctx context.Context, managedClusterName string) error {
+	// TODO support patching both annotations and spec simultaneously in the patcher
+	patch := fmt.Sprintf("{\"metadata\":{\"annotations\":{\"%s\":\"\"}},\"spec\":{\"hubAcceptsClient\":true}}", initialJoinAcceptedAnnotationKey)
```
nirs, line 221: This works but hard to read and error prone. Maybe use a raw string literal to avoid the extra \" escapes?

    patch := fmt.Sprintf(`{"metadata":{"annotations":{"%s":""}},"spec":{"hubAcceptsClient":true}}`)

We don't have a value for the annotation so we use an empty string - maybe we can add a timestamp as the value to have more context? or the name of the controller adding this annotation? This can make it easier to debug issues.

test/integration/registration/spokecluster_autoapproval_test.go
```diff
@@ -88,4 +94,58 @@ var _ = ginkgo.Describe("Cluster Auto Approval", func() {
...
+		}, v1.CreateOptions{})
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		gomega.Eventually(func() error {
+			cluster, err := util.GetManagedCluster(clusterClient, clusterName)
+			if err != nil {
+				return err
+			}
+
+			if _, ok := cluster.Annotations["open-cluster-management.io/initial-join-accepted"]; !ok {
```
nirs, line 113: Based on this test, it seems that the purpose of the annotation is to approve the cluster automatically only once, so if the cluster is denied manually it will not approved automatically again. So the annotation can be renamed to "automatically-approved".

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -103,6 +110,15 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 	}
 
 	if !managedCluster.Spec.HubAcceptsClient {
+		// If the ManagedClusterAutoApproval feature is enabled, we automatically accept this cluster and
+		// add an annotation to mark its initial join as accepted.
+		if features.HubMutableFeatureGate.Enabled(ocmfeature.ManagedClusterAutoApproval) {
+			// Only accept this cluster upon its initial join.
```
nirs, line 116: This comment is not needed, the comment above can explain the cases we want to handle.

nirs (commented):
If I understand this change correctly, the only task the managedCluster controller need to do is to set hubAcceptsClient to true once?

The CSR approval is not related to this and will work correctly even if the CSR is found before the managed cluster is created?

skeeey:
> If I understand this change correctly, the only task the managedCluster controller need to do is to set hubAcceptsClient to true once?
you are correct, the managedCluster controller only set  hubAcceptsClient to true once

> The CSR approval is not related to this and will work correctly even if the CSR is found before the managed cluster is created?
yes, with this, the CSR approval process and managed cluster accept process will be independent

pkg/registration/hub/managedcluster/controller.go
skeeey: agree, I think a timestamp may be more help

test/integration/registration/spokecluster_autoapproval_test.go
skeeey: may be used this annotation
`open-cluster-management.io/automaticlly-accpected-by-managedclusterctrl=<the accept timestamp>`

pkg/registration/hub/managedcluster/controller.go
skeeey: I will update this make it more clear

pkg/registration/hub/managedcluster/controller.go
skeeey: removed

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -20,14 +22,18 @@ import (
...
 	"open-cluster-management.io/sdk-go/pkg/patcher"
 
 	"open-cluster-management.io/ocm/pkg/common/apply"
 	"open-cluster-management.io/ocm/pkg/common/queue"
+	"open-cluster-management.io/ocm/pkg/features"
 	"open-cluster-management.io/ocm/pkg/registration/helpers"
 	"open-cluster-management.io/ocm/pkg/registration/hub/manifests"
 )
 
+const clusterAcceptedAnnotationKey = "open-cluster-management.io/automatically-accepted-by-managedclusterctrl"
```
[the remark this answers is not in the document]
skeeey: yeah, let's go with "automatically-accepted-on" and timestamp firstly

nirs (approved):

nirs:
I tested the latest version, 250 deploys succeeded.

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -20,14 +22,18 @@ import (
...
 	"open-cluster-management.io/sdk-go/pkg/patcher"
 
 	"open-cluster-management.io/ocm/pkg/common/apply"
 	"open-cluster-management.io/ocm/pkg/common/queue"
+	"open-cluster-management.io/ocm/pkg/features"
 	"open-cluster-management.io/ocm/pkg/registration/helpers"
 	"open-cluster-management.io/ocm/pkg/registration/hub/manifests"
 )
 
+const clusterAcceptedAnnotationKey = "open-cluster-management.io/automatically-accepted-on"
```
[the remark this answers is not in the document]
skeeey: added

PR #389, opened 2024-03-29
🐛 create migration cr after the cluster manager cr is applied

zhujian7:
## Summary
This PR will:
- add the `ObservedGeneration` field for the cluster manager Applied condition
- wait the current generation cluster manager is applied(`Applied` condition is true and the `ObservedGeneration`==`Generation`), then create the migration CR

## Related issue(s)

Fixes #


elgnay:
To some extent, this PR https://github.com/open-cluster-management-io/ocm/pull/328 has already ensured the expected CRD DOES exist before creating a Migration CR.
- https://github.com/open-cluster-management-io/ocm/blob/main/pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go#L172
- https://github.com/open-cluster-management-io/ocm/blob/main/pkg/operator/operators/clustermanager/controllers/migrationcontroller/migration_controller.go#L221

qiujian16:
this means we cannot remove webhooks or remove an api version in the same release with migration. I am not quite sure how to control this.

PR #396, opened 2024-04-07
:bug: CI should pass when on codecov upload fail.

xuezhaojun:
## Summary

Codecov becomes because github api outage sometimes. Change to pass CI job when the codecov upload is failed.

## Related issue(s)

https://github.com/codecov/codecov-action/issues/926

### DIFF
--- a/.github/workflows/pre.yml
+++ b/.github/workflows/pre.yml
@@ -69,7 +69,7 @@ jobs:
           flags: unit
           name: unit
           verbose: true
-          fail_ci_if_error: true
+          fail_ci_if_error: false
 
   integration:
     name: integration


.github/workflows/pre.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           flags: unit
           name: unit
           verbose: true
-          fail_ci_if_error: true
+          fail_ci_if_error: false
```
zhujian7, line 72: add a TODO like change this back when we find a way to stabilize it?

xuezhaojun: That's a good point to discuss on, I prefer leave it be to false.  And if codecov continously being unstable we transfer to another tool.

PR #397, opened 2024-04-08
:sparkles: Add image pullSecret to hub controllers

zhiweiyin318:
## Summary
Support to sync the imagePullSecret `open-cluster-management-image-pull-credentials` in the operator ns to the hub namespace. 
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
+++ b/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
@@ -224,7 +224,7 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 			hubMigrationClient: hubMigrationClient, skipRemoveCRDs: n.skipRemoveCRDs},
 		&secretReconcile{cache: n.cache, recorder: n.recorder, operatorKubeClient: n.operatorKubeClient,
 			hubKubeClient: hubClient, operatorNamespace: n.operatorNamespace},
-		&hubReoncile{cache: n.cache, recorder: n.recorder, hubKubeClient: hubClient},
+		&hubReconcile{cache: n.cache, recorder: n.recorder, hubKubeClient: hubClient},
 		&runtimeReconcile{cache: n.cache, recorder: n.recorder, hubKubeConfig: hubKubeConfig, hubKubeClient: hubClient,
 			kubeClient: managementClient, ensureSAKubeconfigs: n.ensureSAKubeconfigs},
 		&webhookReconcile{cache: n.cache, recorder: n.recorder, hubKubeClient: hubClient, kubeClient: managementClient},
@@ -258,6 +258,12 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	encodedCaBundle := base64.StdEncoding.EncodeToString([]byte(caBundle))
 	config.RegistrationAPIServiceCABundle = encodedCaBundle
 	config.WorkAPIServiceCABundle = encodedCaBundle
+
+	// check imagePulSecret here because there will be a warning event FailedToRetrieveImagePullSecret
+	// if imagePullSecret does not exist.
+	if config.ImagePullSecret, err = n.getImagePullSecret(ctx); err != nil {
+		return err
+	}
 
 	for _, reconciler := range reconcilers {
 		var state reconcileState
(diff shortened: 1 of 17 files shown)


qiujian16:
we need to consider this together with #361, we basically want to sync secrets from one ns to another. I think we should handle them at the same time. cc @morvencao

morvencao:
are you referring to https://github.com/open-cluster-management-io/ocm/pull/381 for syncing work-driver-config secret?

pkg/operator/operators/clustermanager/options.go
```diff
@@ -59,6 +61,13 @@ func (o *Options) RunClusterManagerOperator(ctx context.Context, controllerConte
 		helpers.WorkWebhookSecret:         workSecretInformer.Core().V1().Secrets(),
 	}
 
+	// Read component namespace
+	operatorNamespace := constants.DefaultComponentNamespace
+	nsBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+	if err == nil {
+		operatorNamespace = string(nsBytes)
+	}
+
```
morvencao, line 70: the operator namespace can be retrieved from controllerContext of library-go:
```
controllerContext.OperatorNamespace
```

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -211,7 +214,8 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	reconcilers := []clusterManagerReconcile{
 		&crdReconcile{cache: n.cache, recorder: n.recorder, hubAPIExtensionClient: hubApiExtensionClient,
 			hubMigrationClient: hubMigrationClient, skipRemoveCRDs: n.skipRemoveCRDs},
-		&hubReoncile{cache: n.cache, recorder: n.recorder, hubKubeClient: hubClient},
+		&hubReoncile{cache: n.cache, recorder: n.recorder, operatorKubeClient: n.operatorKubeClient,
```
qiujian16, line 217: could we move the secret sync in a separated reconcile func

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -379,3 +383,15 @@ func cleanResources(ctx context.Context, kubeClient kubernetes.Interface, cm *op
 	}
 	return cm, reconcileContinue, nil
 }
+
+func (n *clusterManagerController) getImagePullSecret(ctx context.Context) (string, error) {
```
qiujian16, line 387: do you still need this?

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -259,6 +259,10 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	config.RegistrationAPIServiceCABundle = encodedCaBundle
 	config.WorkAPIServiceCABundle = encodedCaBundle
 
+	if config.ImagePullSecret, err = n.getImagePullSecret(ctx); err != nil {
```
qiujian16, line 262: why we need this step?

zhiweiyin318: will not add the pullsecretName to the sa if the imagepullSecret does not exist.
https://github.com/open-cluster-management-io/ocm/pull/397/files#diff-9f7acaff5c7d5c32d15173a975e8bed87d52c8add8e4bf8b4783854741c4cf39R6

if the imagepullSecret is not created but the name is fixed in the sa, a warning event will be output in the ns. The warning event can cause test case failure in openshift-conformance test suite.
There are 2 options here to avoid the warning event:
1. add a empty pullsecret in the operator namespace.
2. remove the pullsecret name from the sa if the pullsecret does not exist. 

I chose option 2 in this PR. 

the ApplyDirectly in libary-go does not update the imagepullsecret filed in the sa,  I also create a new func ApplyServiceAccount to apply sa in the PR.

zhiweiyin318:
@qiujian16 @morvencao  please take another look.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
zhiweiyin318: moved the imagepullsecret to deployment.

qiujian16 (commented):
/approve
/lgtm

PR #417, opened 2024-04-11
✨ Set install namespace of addonTemplate from config

qiujian16:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/addon/templateagent/values.go
+++ b/pkg/addon/templateagent/values.go
@@ -36,6 +36,15 @@ func ToAddOnRegistriesPrivateValues(config addonapiv1alpha1.AddOnDeploymentConfi
 	}, nil
 }
 
+func ToAddOnInstallNamespacePrivateValues(config addonapiv1alpha1.AddOnDeploymentConfig) (addonfactory.Values, error) {
+	if len(config.Spec.AgentInstallNamespace) == 0 {
+		return nil, nil
+	}
+	return addonfactory.Values{
+		"INSTALL_NAMESPACE": config.Spec.AgentInstallNamespace,
+	}, nil
+}
+
 type keyValuePair struct {
 	name  string
 	value string
@@ -87,7 +96,7 @@ func (a *CRDTemplateAgentAddon) getValues(
 	if err != nil {
 		return presetValues, overrideValues, privateValues, nil
 	}
-	overrideValues = addonfactory.MergeValues(overrideValues, builtinValues)
+	overrideValues = addonfactory.MergeValues(builtinValues, overrideValues)
 
 	for k, v := range overrideValues {
 		_, ok := v.(string)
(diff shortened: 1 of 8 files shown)


qiujian16:
/hold

still need e2e test.

pkg/addon/templateagent/values.go
```diff
@@ -87,7 +96,7 @@ func (a *CRDTemplateAgentAddon) getValues(
 	if err != nil {
 		return presetValues, overrideValues, privateValues, nil
 	}
-	overrideValues = addonfactory.MergeValues(overrideValues, builtinValues)
+	overrideValues = addonfactory.MergeValues(builtinValues, overrideValues)
```
zhujian7, line 99: the builtin Values are immutable

qiujian16: I thought we want overridevalue to override builtinValue?

zhujian7: we want the overridevalue to override defaultValue, but for bultinValues(ClusterName), they should be immutable.

pkg/addon/templateagent/values.go
```diff
@@ -36,6 +36,15 @@ func ToAddOnRegistriesPrivateValues(config addonapiv1alpha1.AddOnDeploymentConfi
 	}, nil
 }
 
+func ToAddOnInstallNamespacePrivateValues(config addonapiv1alpha1.AddOnDeploymentConfig) (addonfactory.Values, error) {
+	if len(config.Spec.AgentInstallNamespace) == 0 {
+		return nil, nil
+	}
+	return addonfactory.Values{
+		"INSTALL_NAMESPACE": config.Spec.AgentInstallNamespace,
```
zhujian7, line 44: Not sure if the install namespace should be a private value, if it is private, it should start with `__`, like`__INSTALL_NAMESPACE`

qiujian16: this is done

pkg/addon/templateagent/values.go
qiujian16: I am a bit confusing, I thought this func is return overrideValues that is using builtingValues as base and override by overrideValue if the key exists. The buitInvalue is immutable as input.

pkg/addon/templateagent/values.go
```diff
@@ -60,8 +69,9 @@ func (a *CRDTemplateAgentAddon) getValues(
 	overrideValues = addonfactory.MergeValues(overrideValues, defaultValues)
```
zhujian7, line 69:
```suggestion
	overrideValues = addonfactory.MergeValues(defaultValues, overrideValues)
```

pkg/addon/templateagent/values_test.go
```diff
@@ -100,3 +104,155 @@ func TestGetAddOnRegistriesPrivateValuesFromClusterAnnotation(t *testing.T) {
...
+			expectedOverride: map[string]interface{}{
+				"CLUSTER_NAME": "test-cluster",
+				"key1":         "value1",
+			},
+			expectedPrivate: map[string]interface{}{
+				InstallNamespacePrivateValueKey: "default-ns",
+			},
+		},
+		{
+			name:         "default should not be overridden",
```
zhujian7, line 183: default values can not be overridden, there are some descriptions in the tests that are not precise, I will submit another PR to refactor it after this is merged.

zhujian7:
@qiujian16 I suggest that we add a `:warning:` to the PR title: `✨ ⚠️ Set install namespace of addonTemplate from config`, and add some description to the PR summary to state the `INSTALL_NAMESPACE` is removed, and user can use downward API, etc.

PR #435, opened 2024-04-21
:seedling: Refactor Lease controller test cases.

xuezhaojun:
## Summary

Nothing change in the real code, all changes is about u-t test cases.

We want to avoid using specific times like:

```go
// wait one cycle (1 ~ 1.25s)
time.Sleep(2000 * time.Millisecond)
```

or:

```go
// wait a few milliseconds to start the lease update routine
time.Sleep(500 * time.Millisecond)
```

When u-t is running on a different environment, it may cause inaccurate results.


skeeey:
LGTM

qiujian16 (commented):
/approve
/lgtm

PR #443, opened 2024-04-25
:sparkles: Registration-agent supports multiple bootstrapkubeconfigs.

xuezhaojun:
## Summary

Klusterlet supports configure multiple bootstrapkubeconfigs and switch when hub doesn't accpet it or failed to connect to a hub.

**Special Notes:**
* The feature is depends on the auto-approve enabled and configured.
* Currently, you can't rebootstrap or restart a agent by change the items in bootstrapkubeconfigs.
* The bootstrapkubeconfigs should use long-term credentials.


pkg/registration/spoke/bootstrapkubeconfigsmanager.go
```diff
@@ -0,0 +1,237 @@
...
+type boostrapKubeConfigStatus string
+
+const (
+	boostrapKubeConfigStatusInValid boostrapKubeConfigStatus = "InValid"
+	boostrapKubeConfigStatusValid   boostrapKubeConfigStatus = "Valid"
+)
+
+// We curretnly only have the implementation of secret type boostrapKubeConfig
+type boostrapKubeConfig interface {
+	Name() string
```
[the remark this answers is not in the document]
xuezhaojun: Comment added.

pkg/registration/spoke/bootstrapkubeconfigsmanager.go
```diff
@@ -0,0 +1,237 @@
+package spoke
```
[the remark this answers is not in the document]
xuezhaojun: Sub dir created.

deploy/cluster-manager/olm-catalog/0.13.1/manifests/operator.open-cluster-management.io_clustermanagers.yaml
```diff
@@ -259,69 +259,6 @@ spec:
                 description: RegistrationImagePullSpec represents the desired image
                   of registration controller/webhook installed on hub.
                 type: string
-              resourceRequirement:
```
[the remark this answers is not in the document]
xuezhaojun: Fixed.

xuezhaojun:
> can we split this to two PRs. the first one only update the registration part.

Yes, I have split the PR and now it only contains registration part.

xuezhaojun:
Still working the intergation-test, hold the PR.
/hold

xuezhaojun:
> can we split this to two PRs. the first one only update the registration part.

Double checked, the operator part is just few lines, maybe we can keep them in this PR as well.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfig.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// KubeConfigData includes the kubeconfig and the credentials to connect to the hub cluster.
+	KubeConfigData() (map[string][]byte, error)
+
+	// Status returns the status of the bootstrap kubeconfig.
+	// A bootstrap kubeconfig has two status: Valid and InValid.
+	Status() (boostrapKubeConfigStatus, error)
+
+	// Fail means at the time t, the bootstrap kubeconfig failed to connect to the hub cluster.
+	Fail(t time.Time) error
```
qiujian16, line 32: what is the purpose of this method? When it returns error

xuezhaojun: Change to MarkFail. And remove input t.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfigsmanager.go
```diff
@@ -0,0 +1,122 @@
...
+			secretNamespace: spokeNamespace,
+			kubeClient:      kubeClient,
+		},
+		bootstrapKubeConfigs: bootstrapKubeConfigs,
+		reselected:           false,
+	}
+}
+
+// ReSelect is the only public method of a bootstrapkubeconfigManager
+func (m *bootstrapkubeconfigsManager) ReSelect(ctx context.Context) error {
```
xuezhaojun, line 53: Change to contextual Log.

xuezhaojun: Change to event recorder?

xuezhaojun: log event?

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfiginuse.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (b *bootstrapKubeConfigInUseImpl) KubeConfigData() (map[string][]byte, error) {
+	secret, err := b.kubeClient.CoreV1().Secrets(b.secretNamespace).Get(context.Background(), b.secretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("get the bootstrap kubeconfig secret failed: %v", err)
+	}
+	return secret.Data, nil
+}
+
+func (b *bootstrapKubeConfigInUseImpl) Update(ctx context.Context, kubeconfigData map[string][]byte) error {
```
xuezhaojun, line 36: Need to validate KubeConfigData, must have kubeconfig key etc

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfiginuse.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// bootstrapKubeConfigInUse is the registration spoke's current in used bootstrap kubeconfig.
+type bootstrapKubeConfigInUse interface {
```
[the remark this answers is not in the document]
xuezhaojun: ack.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfiginuse.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Update updates the kubeconfig data of the bootstrap kubeconfig in use.
+	Update(ctx context.Context, kubeconfigData map[string][]byte) error
+}
+
+var _ bootstrapKubeConfigInUse = &bootstrapKubeConfigInUseImpl{}
+
+type bootstrapKubeConfigInUseImpl struct {
+	secretName      string
+	secretNamespace string
```
xuezhaojun, line 24: remove secretName and use from constant pkg.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfigsmanager.go
```diff
@@ -0,0 +1,122 @@
...
+	"k8s.io/klog/v2"
+)
+
+// bootstrapkubeconfigsManager manages multiple bootstrapkubeconfigs.
+// The only public method is ReSelect.
+// ReSelect do 2 things:
+// 1. Call Fail method of the bootstrapkubeconfig that has the same kubeconfig data with the current bootstrapkubeconfig in use.
+// 2. Select a first valid bootstrapkubeconfig from the list and update it to the current bootstrapkubeconfig in use.
+//
+// Note: ReSelect should be called only once. Then it waits for registration-agent to restart and rebootstrap.
```
xuezhaojun, line 20: Need another controller to sync changes in secret in the list to bootstarap-hub-kubeconfig when use actively change it.

pkg/registration/spoke/lease/lease_controller.go
```diff
@@ -129,15 +139,52 @@ func (u *leaseUpdater) stop() {
...
+		if !u.lastSuccuessUpdateTime.IsZero() && now.After(u.lastSuccuessUpdateTime.Add(time.Duration(u.timeoutSeconds)*time.Second)) {
+			utilruntime.HandleError(fmt.Errorf("Timeout to update lease cluster lease %q on hub cluster", u.leaseName))
+			if err := u.handleTimeout(ctx); err != nil {
+				utilruntime.HandleError(fmt.Errorf("unable to handle timeout of cluster lease %q on hub cluster: %w", u.leaseName, err))
+			}
+			return
+		}
+	}
+
+	handleErrorAndTimeout := func(err error, now time.Time) {
```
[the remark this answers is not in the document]
xuezhaojun: Agree, wrapper moved out, and a u-t added.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfiginuse.go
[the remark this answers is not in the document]
xuezhaojun: No, the current bootstrap kubeconfig used to trigger rebootstrap is based on the secret, and if in the future we want to change to another type such as file, then the interface is a good practice to prevent that change impact other parts of code.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfigsmanager.go
```diff
@@ -0,0 +1,117 @@
...
+			kubeClient:      kubeClient,
+		},
+		bootstrapKubeConfigs: bootstrapKubeConfigs,
+		reselected:           false,
+	}
+}
+
+// ReSelect is the only public method of a bootstrapkubeconfigManager
+func (m *bootstrapkubeconfigsManager) ReSelect(ctx context.Context) error {
+	if m.reselected {
```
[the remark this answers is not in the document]
xuezhaojun: Never in the same lifecycle, the pod will restart and in the next lifecycle of container, the reselect will be set as false initially.

pkg/registration/spoke/spokeagent.go
```diff
@@ -411,6 +424,16 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 		)
 	}
 
+	hubAcceptController := registration.NewHubAcceptController(
+		o.agentOptions.SpokeClusterName,
```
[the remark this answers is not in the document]
xuezhaojun: Yes, the feature depends on the auto-approve enabled and configured, I will note this point in the doc in the future.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfig.go
xuezhaojun: Done.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfigsmanager.go
xuezhaojun: Changed.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfiginuse.go
xuezhaojun: validatioin for the key "kubeconfig" added.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfiginuse.go
xuezhaojun: @qiujian16 

From the code we know, the spoke compoennt namespace is not always the  defaultSpokeComponentNamespace, it depends on which ns the agent is installed, which means it's neccessary to keep the `secretNamespace`.

```
// NewAgentOptions returns the flags with default value set
func NewAgentOptions() *AgentOptions {
	opts := &AgentOptions{
		HubKubeconfigDir:   "/spoke/hub-kubeconfig",
		ComponentNamespace: defaultSpokeComponentNamespace,
		CommoOpts:          NewOptions(),
	}
	// get component namespace of spoke agent
	nsBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
	if err == nil {
		opts.ComponentNamespace = string(nsBytes)
	}
	return opts
}
```

pkg/registration/spoke/registration/hub_accept_controller.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (c *hubAcceptController) sync(ctx context.Context, _ factory.SyncContext) error {
+	cluster, err := c.hubClusterLister.Get(c.clusterName)
+	if err != nil {
+		return err
+	}
+	if cluster.Spec.HubAcceptsClient {
+		return nil
+	}
```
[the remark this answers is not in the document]
xuezhaojun: The related code: https://github.com/xuezhaojun/OCM/blob/fc574c1be4fcfe1dfb0e250b41ef9cd3b9c1705c/pkg/registration/hub/managedcluster/controller.go#L132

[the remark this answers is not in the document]
xuezhaojun: Comment added: the hubAccpetController can only handle the case that "true" -> "false".

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfigsmanager.go
xuezhaojun: Done, another controller added.

pkg/registration/spoke/bootstrapkubeconfigsmanager/bootstrapkubeconfiginuse.go
xuezhaojun: Not always true that secret contains the key of kubeconfig.

pkg/registration/spoke/registration/hub_accept_controller.go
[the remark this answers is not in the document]
xuezhaojun: PS: from [this line of code](https://github.com/kubernetes/apimachinery/blob/1da46c3f5a5b4a0cc756cb6050df0cf6f06b64c2/pkg/watch/mux.go#L290-L291) we can see that an event is send to all watchers at the same time. So if the informer on the hub side and informer on the agent side will both get the `update` event.

.codecov.yml
```diff
@@ -8,3 +8,4 @@ coverage:
 
 ignore:
   - "**/*generated*.go"
+  - "**/ssar.go"
```
qiujian16, line 11: why ignore this file?

pkg/common/helpers/ssar.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return true, nil, nil
+}
+
+func GetBootstrapSSARs() []authorizationv1.SelfSubjectAccessReview {
+	var reviews []authorizationv1.SelfSubjectAccessReview
+	clusterResource := authorizationv1.ResourceAttributes{
+		Group:    "cluster.open-cluster-management.io",
+		Resource: "managedclusters",
+	}
```
qiujian16, line 35: I think we should also add the resourceName here.

Update: should at least add a TODO here, we should add resourceName in the future.

pkg/registration/hub/manifests/rbac/managedcluster-registration-clusterrole.yaml
```diff
@@ -11,7 +11,7 @@ rules:
   # TODO for backward compatible, we do not limit the resource name
   # remove this after we no longer support lower versions kubernetes (less than 1.14)
   #resourceNames: ["managed-cluster-lease"]
-  verbs: ["get", "update"]
+  verbs: ["get", "update", "list", "watch"]
```
qiujian16, line 14: why need list/watch here?

pkg/registration/spoke/bootstrapkubeconfigs.go
```diff
@@ -0,0 +1,75 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	clusterv1client "open-cluster-management.io/api/client/cluster/clientset/versioned"
+
+	commonhelpers "open-cluster-management.io/ocm/pkg/common/helpers"
+)
+
+func selectBootstrapKubeConfigs(ctx context.Context, managedCluster string, bootstrapKubeConfigs []*rest.Config) (int, error) {
```
qiujian16, line 19: add some comment. You are selecting based on order, and find the first available hub cluster.

pkg/registration/spoke/options.go
```diff
@@ -38,12 +51,17 @@ func NewSpokeAgentOptions() *SpokeAgentOptions {
...
 			interval: ClientCertHealthCheckInterval,
 		},
+		HubConnectionTimeoutSeconds: 600, // by default, the timeout is 10 minutes
 	}
 
 	options.bootstrapKubeconfigHealthChecker = &bootstrapKubeconfigHealthChecker{
 		bootstrapKubeconfigSecretName: &options.BootstrapKubeconfigSecret,
 	}
 
+	options.reSelectChecker = &reSelectChecker{
```
qiujian16, line 61: no need to declare separately, just put this under line 54

pkg/registration/spoke/bootstrapkubeconfigs.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		if mc.Spec.HubAcceptsClient {
+			return index, nil
+		}
+	}
+
+	return -1, fmt.Errorf("no bootstrap kubeconfig is available for managed cluster %s", managedCluster)
+}
+
+type reSelectChecker struct {
```
qiujian16, line 62: comments is needed. What is for? and how shouldReSelect should be set.

pkg/registration/spoke/options.go
```diff
@@ -70,8 +92,15 @@ func (o *SpokeAgentOptions) AddFlags(fs *pflag.FlagSet) {
...
 func (o *SpokeAgentOptions) Validate() error {
-	if o.BootstrapKubeconfig == "" {
-		return errors.New("bootstrap-kubeconfig is required")
+	if features.SpokeMutableFeatureGate.Enabled(ocmfeature.MultipleHubs) {
+		// expect BootstrapKubeconfig is empty and BootstrapKubeconfigs has at least 2 items
+		if len(o.BootstrapKubeconfigs) < 2 {
+			return errors.New("expect at least 2 bootstrap kubeconfigs")
+		}
+	} else if !features.SpokeMutableFeatureGate.Enabled(ocmfeature.MultipleHubs) {
+		if o.BootstrapKubeconfig == "" {
```
qiujian16, line 101: I do not think we need this check. if it is empty, we will just check the kubeconfig in the default location.

pkg/registration/spoke/registration/bootstrapkubeconfig_secret_controller.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/apimachinery/pkg/api/meta"
+	corev1informers "k8s.io/client-go/informers/core/v1"
+
+	"open-cluster-management.io/ocm/pkg/common/queue"
+)
+
+// The controller watches the BootstrapKubeConfig secrets, and handle the secret data change.
+type bootstrapKubeConfigSecretsController struct {
```
qiujian16, line 15: why we still need this controller?

pkg/registration/spoke/registration/hub_timeout_controller.go
```diff
@@ -0,0 +1,70 @@
...
+		ToController("HubTimeoutController", recorder)
+}
+
+func (c *hubTimeoutController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	logger := klog.FromContext(ctx)
+	if c.handleTimeout == nil {
+		return nil
+	}
+
+	lease, err := c.leaseLister.Leases(c.clusterName).Get(leaseName)
```
qiujian16, line 46: Can we just get lease using client? user lister does not make sense here, it works only when you watch instead of polling.

pkg/registration/spoke/registration/hub_timeout_controller.go
```diff
@@ -0,0 +1,70 @@
...
+		if c.lastLeaseRenewTime.IsZero() {
+			c.lastLeaseRenewTime = time.Now()
+		}
+	} else {
+		c.lastLeaseRenewTime = lease.Spec.RenewTime.Time
+	}
+
+	if isTimeout(time.Now(), c.lastLeaseRenewTime, c.timeoutSeconds) {
+		logger.Info("Lease timeout", "cluster", c.clusterName, "lease", leaseName)
+		err := c.handleTimeout(ctx)
```
qiujian16, line 59: I think I understand what you are trying to do...So the you get from local cache, and if the network is disconnected, the local cache will not be updated? If so I would rather use client to get and maintain in this controller, using informer and lister is too heavy.

Also we should add a TODO. These code should finally be put into the existing lease controller.

pkg/registration/spoke/spokeagent.go
```diff
@@ -170,10 +170,52 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
...
+		var bootstrapKubeConfigs []*rest.Config
+		for _, fp := range o.registrationOption.BootstrapKubeconfigs {
+			kc, err := clientcmd.BuildConfigFromFlags("", fp)
+			if err != nil {
+				return fmt.Errorf("unable to load bootstrap kubeconfig from file %q: %w", fp, err)
+			}
+			bootstrapKubeConfigs = append(bootstrapKubeConfigs, kc)
+		}
+
+		index, err := selectBootstrapKubeConfigs(ctx, o.agentOptions.SpokeClusterName, bootstrapKubeConfigs)
```
qiujian16, line 190: if so, why just let select fun return a *rest.Config?

pkg/registration/spoke/spokeagent.go
```diff
@@ -418,6 +459,54 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
...
 	}
 
+	var hubAcceptController, hubTimeoutController factory.Controller
+	hubKubeConfigSecretAutoGenCtx, cancelHubKubeConfgiSecretAutoGen := context.WithCancel(ctx)
+	defer cancelHubKubeConfgiSecretAutoGen()
+	if features.SpokeMutableFeatureGate.Enabled(ocmfeature.MultipleHubs) {
+		hubAcceptController = registration.NewHubAcceptController(
+			o.agentOptions.SpokeClusterName,
+			hubClusterInformerFactory.Cluster().V1().ManagedClusters(),
+			func(ctx context.Context) error {
```
qiujian16, line 469: this is duplicate with line 492, should build the same one.

pkg/registration/spoke/spokeagent.go
```diff
@@ -418,6 +459,54 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
...
+				logger.Info("Failed to connect to hub because of hubAcceptClient set to false, restart agent to reselect a new bootstrap kubeconfig")
+				cancelHubKubeConfgiSecretAutoGen()
+				// remove hub kubeconfig secret
+				if err := managementKubeClient.CoreV1().Secrets(o.agentOptions.ComponentNamespace).Delete(ctx,
```
qiujian16, line 473: do we really need to remove the secret? it will be restarted with the changed bootstrap which makes the existing hubkubeconfig invalid, right?

pkg/registration/spoke/spokeagent.go
```diff
@@ -418,6 +459,54 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 		)
 	}
 
+	var hubAcceptController, hubTimeoutController factory.Controller
+	hubKubeConfigSecretAutoGenCtx, cancelHubKubeConfgiSecretAutoGen := context.WithCancel(ctx)
+	defer cancelHubKubeConfgiSecretAutoGen()
```
qiujian16, line 464: what is this for?

pkg/registration/spoke/spokeagent.go
```diff
@@ -170,10 +170,52 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
...
+		if err != nil {
+			return fmt.Errorf("failed to select a bootstrap kubeconfig: %w", err)
+		}
+		logger.Info("Select bootstrap kubeconfig", "index", index, "file", o.registrationOption.BootstrapKubeconfigs[index])
+
+		bootstrapKubeConfig = o.registrationOption.BootstrapKubeconfigs[index]
+
+		// hubValiedConfig
+		hubValidConfig = func(ctx context.Context) (bool, error) {
+			return o.HasValidHubClientConfig(ctx)
```
qiujian16, line 200: why we do not compare hub and boostratp config here?

pkg/registration/hub/manifests/rbac/managedcluster-registration-clusterrole.yaml
xuezhaojun: the hubtimeout controller needs to check the lease every minutes since we mentioned the apiserver call pressure, I think we should using lister for this.

.codecov.yml
xuezhaojun: The code in this file mostly consists of construction functions. It not ignored then I need to improve coverage somewhere else.

pkg/common/helpers/ssar.go
xuezhaojun: Added.

pkg/registration/spoke/bootstrapkubeconfigs.go
xuezhaojun: Comment added.

pkg/registration/spoke/options.go
xuezhaojun: Updated.

pkg/registration/spoke/bootstrapkubeconfigs.go
xuezhaojun: Added.

pkg/registration/spoke/options.go
xuezhaojun: Does "this check" means the check of the featureGates?

pkg/registration/spoke/registration/bootstrapkubeconfig_secret_controller.go
xuezhaojun: Forget to remove, no removed.

pkg/registration/spoke/registration/hub_timeout_controller.go
xuezhaojun: Sure, that makes sense.

pkg/registration/spoke/spokeagent.go
xuezhaojun: Some legacy code is relaying on the file type of bootstrapkubeconfig:
```
server, proxyURL, caData, err := parseKubeconfig(bootstrapKubeConfig)
```

Can I add a todo for this? I will to refactor the core logic of registration to depending on *rest.Config.

pkg/registration/spoke/spokeagent.go
xuezhaojun: Agree, merged.

pkg/registration/spoke/spokeagent.go
xuezhaojun: because the hub-kubeconfig and the bootstrap-kubeconfig are not always should be equal.

For example, we have a [hub1, hub2],both hub1 and hub2 are available now, currently the agent connect to hub2. So the hub-kubeconfig is point to hub2.

Now we restart the agent, with our select logic, the bootstrapkubeconfig we choose will point to hub1. But we still want to hub-kubeconfig points to hub2.

If we compare the bootstrapkubeconfig and hub-kubeconfig, the agent will fail on this line:
```
        // check if there already exists a valid client config for hub
	ok, err := hubValidConfig(ctx)
	if err != nil {
		return err
	}
```

pkg/registration/spoke/spokeagent.go
xuezhaojun: Because we using checker and health probe to perform restart, we need to pervent the hubKubeConfigSecret to created by controller after we delete it.

The reason why we need to delete hub-kubeconfig is list up.

pkg/registration/spoke/spokeagent.go
xuezhaojun: No, as I mentioned in https://github.com/open-cluster-management-io/ocm/pull/443#discussion_r1609224184, in MultipleHubs enabled cause, we don't compare hub-kubeconfig and bootstrap-kubeconfig.

pkg/registration/hub/manifests/rbac/managedcluster-registration-clusterrole.yaml
xuezhaojun: Removed, use kubeclient instead of lister in hubtimeout controller.

.codecov.yml
qiujian16: The code here is just move from other files? I think the existing ut has covered them already, why this movement will impact coverage?

pkg/registration/spoke/options.go
qiujian16: if o.BootstrapKubeconfig == ""

pkg/registration/spoke/spokeagent.go
qiujian16: yes, that is ok. Let's add a TODO

pkg/registration/spoke/spokeagent.go
qiujian16: So that is about how we select bootstrap right? I think we can do like this:

1. if there is kubeKubeconfig, get the part we want to compare with bootstrap, and check the matched bootstrap at first.
2 if the matched boostratp is found and check passed. skip the rebootstrap
3. otherwise, find the bootstrap based on order.

.codecov.yml
xuezhaojun: It moves from one package to another, so when caculate the average code coverage score, it's 80% to (0%+80%)/2 and cause a drop.

I'm working add u-t in other files, I will improve the score back.

pkg/registration/spoke/spokeagent.go
xuezhaojun: It's a great point, updated accordingly.

pkg/registration/spoke/spokeagent.go
xuezhaojun: Remove.

pkg/registration/spoke/spokeagent.go
xuezhaojun: Remove.

pkg/registration/spoke/options.go
```diff
@@ -8,15 +8,27 @@ import (
...
+	// The differences among BootstrapKubeconfig, BootstrapKubeconfigSecret, BootstrapKubeconfigSecrets are:
+	// 1. BootstrapKubeconfig is a file path, the controller uses it to build the client.
+	// 2. BootstrapKubeconfigSecret is the secret, a eventhandler will watch it, if the secret is changed, threbootstrap.
+	// 3. BootstrapKubeconfigs is a list of file path, the controller uses one of its item to build the client.
+	// BootstrapKubeconfigs can only be used when MultipleHubs is enabled.
+	BootstrapKubeconfig       string
+	BootstrapKubeconfigSecret string
+	BootstrapKubeconfigs      []string
+
+	HubConnectionTimeoutSeconds int32
```
qiujian16, line 30: What this value should be set, is this related to the leaseDurationSecond in managedcluster spec?

pkg/registration/spoke/bootstrapkubeconfigs.go
```diff
@@ -0,0 +1,119 @@
...
+// 1. if find the matched bootrapkubeconfig then check it first, if it's valid then continue to use it,
+// no need to select another bootstrapkubeconfig.
+// 2. selects the first available bootstrap kubeconfig from the given list of configurations.
+// If no suitable kubeconfig is found, it returns -1 and an error indicating that no bootstrap kubeconfig is available for the specified managed cluster.
+func selectBootstrapKubeConfigs(ctx context.Context,
+	managedCluster, hubKubeConfigFilePath string, bootstrapKubeConfigFilePaths []string) (int, error) {
+	logger := log.FromContext(ctx)
+
+	for index, fp := range bootstrapKubeConfigFilePaths {
+		equal, err := areHubAndBootstrapKubeConfigsEqual(hubKubeConfigFilePath, fp)
```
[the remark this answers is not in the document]
xuezhaojun: This function renamed from  https://github.com/open-cluster-management-io/ocm/blob/a148049f2dd753d5a8f9874280de0e1b9b4f24f9/pkg/registration/spoke/spokeagent.go#L507 

If hub-kubeconfig is always a copy from one of the bootstrapkuebconfigs, I think it's ok to keep it the same?

pkg/registration/spoke/options.go
xuezhaojun: The value is not necessarily related to leaseDurationSecond; by default, we set the value to 10 minutes and at least 3 minutes:
```yaml
                          hubConnectionTimeoutSeconds:
                            default: 600
                            description: HubConnectionTimeoutSeconds is used to set
                              the timeout of connecting to the hub cluster. When agent
                              loses the connection to the hub over the timeout seconds,
                              the agent do a rebootstrap. By default is 10 mins.
                            format: int32
                            minimum: 180
```

qiujian16: why it is not related? It will not make sense that leaseDurationSecond is set to 5 mins and hubConnectionTimeoutSeconds is 1 mins.

pkg/registration/spoke/bootstrapkubeconfigs.go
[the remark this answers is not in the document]
qiujian16: This func is to check if we need to rebootstrap, but not switch. I think we only switch when apiserver endpoint is not the same, we do not care CA or other config.

pkg/registration/spoke/options.go
xuezhaojun: If the leaseDurationSecond is 5min and hubConnnectionTimeoutSeconds is 1 min I prefer consider it as a configuration error of user.

Do you mean we should set the timeout seconds a number times the leaseDurationSecond? I'm not sure if it can suite every customer's case. For example, if the leaseDurationSecond set to 1min is a common thing, but in different cases, customer may want to set hubTimeoutSeconds to 3mins, 5mins or 10mins.

pkg/registration/spoke/bootstrapkubeconfigs.go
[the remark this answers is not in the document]
xuezhaojun: Got it. Will fix.

pkg/registration/spoke/options.go
xuezhaojun: Do you think we can add a step in the validate to check hubTimeoutSeconds must bigger than the leaseDurationSeconds?

qiujian16: leaseDurationSeconds can be configured in ManagedCluster, so add such step won't work. I think we need to at least add some doc here. It is ok as the first step to have hubConnnectionTimeoutSeconds as an flag, but we should have a TODO to retrieve it from leaseDurationSeconds.

xuezhaojun: Another reason is `leaseDurationSeconds` is a field per managedcluster. 
```go
observedLeaseDurationSeconds := cluster.Spec.LeaseDurationSeconds
```

But we set hubTimetoutSeconds as a spoke-agent configuration field. We can't compare those 2 fields in the Validate function before agent run.

xuezhaojun: I think there is another way to ensure timeout always greater than leaseDuration, we already has a min value(180 seconds) set for hubTimeoutSeconds in api definenation: https://github.com/open-cluster-management-io/api/blob/9d94703b9eba006e618a68588000722bb945112f/operator/v1/types_klusterlet.go#L209

We can also set a max value(60 seconds) for the leaseDurationSeconds(), so that one is always greater than another?

qiujian16: we cannot set maximum, one of the reason to allow configuration is in the pool network scenario, we allow user to set it longer to tolerate temp network disruption.

xuezhaojun: Agree, the TODO added.

test/integration/registration/spokeagent_switch_hub_test.go
```diff
@@ -0,0 +1,371 @@
...
+	ginkgo.By("Accept and approve the ManagedCluster")
+	// Simulate hub cluster admin to accept the managedcluster and approve the csr
+	gomega.Eventually(func() error {
+		return util.AcceptManagedCluster(hubClusterClient, managedClusterName)
+	}, eventuallyTimeout, eventuallyInterval).Should(gomega.Succeed())
+	err := auth.ApproveSpokeClusterCSR(hubKubeClient, managedClusterName, certAget)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+}
+
+func assertManagedClusterSuccessfullyJoined(testNamespace, managedClusterName, hubKubeconfigSecret string,
```
qiujian16, line 277: do we have an existing assert that already cover this ?

test/integration/registration/spokeagent_switch_hub_test.go
```diff
@@ -0,0 +1,371 @@
...
+	commOptions.HubKubeconfigDir = hubKubeconfigDir
+	commOptions.SpokeClusterName = managedClusterName
+
+	agentCtx, stopAgent := context.WithCancel(ctx)
+	runAgentWithContext(agentCtx, "switch-hub", agentOptions, commOptions, spokeCfg)
+
+	return agentCtx, stopAgent
+}
+
+func approveAndAcceptManagedCluster(managedClusterName string,
```
qiujian16, line 232: do we have an existing helper that already did this?

qiujian16 (commented):
/approve

2 nit comment on test, but should not block merge. This is a good start. Great work!

pkg/registration/spoke/options.go
```diff
@@ -8,15 +8,30 @@ import (
...
 
 var ClientCertHealthCheckInterval = 30 * time.Second
 
 // SpokeAgentOptions holds configuration for spoke cluster agent
 type SpokeAgentOptions struct {
-	BootstrapKubeconfig         string
-	BootstrapKubeconfigSecret   string
+	// The differences among BootstrapKubeconfig, BootstrapKubeconfigSecret, BootstrapKubeconfigSecrets are:
+	// 1. BootstrapKubeconfig is a file path, the controller uses it to build the client.
+	// 2. BootstrapKubeconfigSecret is the secret, a eventhandler will watch it, if the secret is changed, threbootstrap.
```
zhujian7, line 23:
```suggestion
	// 2. BootstrapKubeconfigSecret is the secret, an event handler will watch it, if the secret is changed, then rebootstrap.
```
?

pkg/registration/spoke/registration/hub_timeout_controller.go
```diff
@@ -0,0 +1,73 @@
...
+		if err != nil {
+			logger.Error(err, "Failed to handle lease timeout", "cluster", c.clusterName, "lease", leaseName)
```
zhujian7, line 64: should we return the error here?

zhujian7:
@xuezhaojun great work!
/lgtm
/hold for other reviewers

pkg/registration/spoke/options.go
xuezhaojun: Good catch, fixed.

pkg/registration/spoke/registration/hub_timeout_controller.go
xuezhaojun: The controller will reconcile every minute, it could be a timer actually. whether reconcile again on the error case doesn't matter, we have a TODO to implement the timeout feature in leaseController later on.

xuezhaojun:
> @xuezhaojun great work! /lgtm /hold for other reviewers

Thanks for the review! Code updated according to the comments, also need a /lgtm again.

zhujian7:
I think we should have more UTs to cover https://github.com/open-cluster-management-io/ocm/pull/443#issuecomment-2077576620

xuezhaojun:
> I think we should have more UTs to cover [#443 (comment)](https://github.com/open-cluster-management-io/ocm/pull/443#issuecomment-2077576620)

The score doesn't perfectly reflect the coverage, we have many lines of uncovered code are imports or construction functions like:

```
func GetBootstrapSSARs() []authorizationv1.SelfSubjectAccessReview {
	var reviews []authorizationv1.SelfSubjectAccessReview
	clusterResource := authorizationv1.ResourceAttributes{
		Group:    "cluster.open-cluster-management.io",
		Resource: "managedclusters",
		// TODO: add the resourceName @xuezhaojun https://github.com/open-cluster-management-io/ocm/pull/443#discussion_r1609202000
	}
	reviews = append(reviews, generateSelfSubjectAccessReviews(clusterResource, "create", "get")...)

	certResource := authorizationv1.ResourceAttributes{
		Group:    "certificates.k8s.io",
		Resource: "certificatesigningrequests",
	}
	return append(reviews, generateSelfSubjectAccessReviews(certResource, "create", "get", "list", "watch")...)
}
```

And intergration-test also covers some logic of the controller at the high level.

PR #448, opened 2024-04-28
✨ Send importing events for hosted mode klusterlet

zhujian7:
## Related issue(s)

Fixes #


zhujian7:
we can not send this event from the hosting cluster to the hub cluster by the bootstrap kubeconfig, since at that time, the managed cluster may not been approved on the hub cluster, so the managedcluster namespace may not been created on the hub cluster. sending the event to the hub cluster in the managedcluster namespace may fail.
/close

PR #457, opened 2024-05-13
 🐛 Always create default addon namespace

qiujian16:
The addon namespace should always be default on.
The operator will not create addon ns based
on klusterlet install namespace.

## Related issue(s)

Fixes #


pkg/operator/helpers/helpers.go
```diff
@@ -49,6 +49,9 @@ const (
 	FeatureGatesTypeValid             = "ValidFeatureGates"
 	FeatureGatesReasonAllValid        = "FeatureGatesAllValid"
 	FeatureGatesReasonInvalidExisting = "InvalidFeatureGatesExisting"
+
+	// DefaultAddonNamespace is the default namespace for agent addon
+	DefaultAddonNamespace = "open-cluster-management-addon"
```
zhujian7, line 54: open-cluster-management-agent-addon?

PR #458, opened 2024-05-13
🐛 Refresh external managed token secret if service account ns changes

zhujian7:
## Related issue(s)

Fixes #


qiujian16 (commented):
/approve

PR #463, opened 2024-05-15
:bug: igonre failed to get pullsecret error

zhiweiyin318:
## Related issue(s)

Fixes #


pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -262,7 +262,7 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	// check imagePulSecret here because there will be a warning event FailedToRetrieveImagePullSecret
 	// if imagePullSecret does not exist.
 	if config.ImagePullSecret, err = n.getImagePullSecret(ctx); err != nil {
-		return err
+		klog.Errorf("failed to get image pull secret: %v", err)
```
zhujian7, line 265: if this error will not result in running to fail, it should be Info or Warning level.

qiujian16: ditto, also add a todo here that we should return err in the next release.

zhiweiyin318: done

qiujian16 (commented):
/approve
/lgtm

PR #468, opened 2024-05-20
✨ configurable controller replicas and master node selector

promid:
## Summary
In production, operators should be able to controll the replicas of the deployments. Making replicas larger than 1 is very important. So this PR is submitted to make controller replicas and master node selector configurable.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cmd/hub/operator.go
+++ b/pkg/cmd/hub/operator.go
@@ -22,6 +22,8 @@ func NewHubOperatorCmd() *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.BoolVar(&cmOptions.SkipRemoveCRDs, "skip-remove-crds", false, "Skip removing CRDs while ClusterManager is deleting.")
+	flags.StringToStringVar(&cmOptions.MasterNodeLabelSelector, "master-node-label-selector", map[string]string{"node-role.kubernetes.io/master": ""}, "Label selector for master nodes, e.g. 'node-role.kubernetes.io/master='")
+	flags.Int32Var(&cmOptions.ControllerReplicas, "controller-replicas", 0, "Number of controller replicas, operator will automatically determine replicas if not set")
 	opts.AddFlags(flags)
 	return cmd
 }
--- a/pkg/operator/helpers/helpers.go
+++ b/pkg/operator/helpers/helpers.go
@@ -37,7 +37,6 @@ import (
 	"k8s.io/klog/v2"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	apiregistrationclient "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
-
 	"open-cluster-management.io/api/feature"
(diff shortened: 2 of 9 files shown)


pkg/cmd/hub/operator.go
```diff
@@ -22,6 +22,10 @@ func NewHubOperatorCmd() *cobra.Command {
 
 	flags := cmd.Flags()
 	flags.BoolVar(&cmOptions.SkipRemoveCRDs, "skip-remove-crds", false, "Skip removing CRDs while ClusterManager is deleting.")
+	flags.StringToStringVar(&cmOptions.MasterNodeLabelSelector, "master-node-label-selector",
```
zhiweiyin318, line 25: The flag is named `label-selector`, so I think it should have 2 parts : MatchLabels and MatchExpressions, like the definition of LabelSelector  https://github.com/open-cluster-management-io/ocm/blob/92a2c91f15947d30627bf2652c5155cf81971437/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1213  So I think string type is better than map here.
Otherwise maybe we need to rename the flag name, for example  control-plane-node labels ?

pkg/operator/helpers/helpers.go
```diff
@@ -451,14 +452,15 @@ func DetermineReplica(ctx context.Context, kubeClient kubernetes.Interface, mode
...
+	return DetermineReplicaByNodes(ctx, kubeClient, masterNodeLabelSelector)
 }
 
 // DetermineReplicaByNodes determines the replica of deployment based on:
 // list master nodes in the cluster and return 1 if
 // the number of master nodes is equal or less than 1. Return 3 otherwise.
-func DetermineReplicaByNodes(ctx context.Context, kubeClient kubernetes.Interface) int32 {
-	nodes, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{LabelSelector: "node-role.kubernetes.io/master="})
+func DetermineReplicaByNodes(ctx context.Context, kubeClient kubernetes.Interface, masterNodeLabelSelector map[string]string) int32 {
+	labelSelector := mapToString(masterNodeLabelSelector)
```
zhiweiyin318, line 462: suggest to use `FormatLabelSelector`: https://github.com/open-cluster-management-io/ocm/blob/92a2c91f15947d30627bf2652c5155cf81971437/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/helpers.go#L169

pkg/cmd/hub/operator.go
promid: renamed to `control-plane-node-labels`

pkg/operator/helpers/helpers.go
promid: updated

promid:
@zhiweiyin318 I changed `controlPlaneNodeLabels` to `controlPlaneNodeLabelSelector`, please review this PR again. Thanks

zhiweiyin318:
@promid the value of the flag `controlPlaneNodeLabelSelector`  is label not labelSelector, so is there a reason why use selector here ?

Issue #467, opened 2024-05-17
Apply all labels that klusterlet has to all the resources it will create and keep labels in sync


EricaJ6:
**Describe the enhancement**
We observed that resources created by klusterlet do not have labels that klusterlet has. For example, `klusterlet-agent` in the `open-cluster-management-agent` namespace is missing labels. We’re using these labels for cost tracking purposes. As discussed with @qiujian16, and agreed to apply all labels that klusterlet has to all the resources it will create. Additionally, if labels on the klusterlet got changed (either updated or added new ones), all the resources created by it should be in sync as well.

qiujian16:
@zhiweiyin318 will you be able to take a look at this one?

zhiweiyin318:
@EricaJ6 @qiujian16  I have some questions about this requirement.
1. just double check that we need to sync the label to the all agent resources or only deployments since the description mentioned ` cost tracking purposes`.
2. do we need delete the label from the agent if it  is removed from the klusterlet ? the reason why I ask is currently the map update strategy is merge.

EricaJ6:
Thanks @zhiweiyin318! For your questions,
1. We should sync the labels to all resources created by the klusterlet. The goal is all resources should have the necessary labels. I looked into the above PR, it looks like currently, only deployments have labels added, could you please add labels to all resources (such as `spec. template. metadata.labels` and so on)?
2. If the label got removed from the klusterlet, it should also be removed from all resources created by the klusterlet.

zhiweiyin318:
@EricaJ6  currently there are 2 labels `app: klusterlet-agent`  and `createdBy: klusterlet `on the deployment created by klusterlet.
if we add the 2 labels to all agent resources, not sure if can meet your requirements for cost tracking purpose?
if the 2 labels can meet your requirements, I think it is not necessary to sync labels from klusterlet.

EricaJ6:
@zhiweiyin318 Other than the 2 default labels, we are adding additional customization labels to klusterlet, and the organization is using these customized labels for cost-tracking purpose. That's why we would like to sync the labels to all resources created by the klusterlet. Because of the use of additional customization labels, the 2 default labels do not meet the requirements.

EricaJ6:
Thanks @zhiweiyin318 on sharing the information and updating the PR! 

1.  To make sure all pods created by the deployments also have the labels, how about adding labels to the `spec.template.metadata.labels` of the `*deployment.yaml` files as well?
[...]

PR #475, opened 2024-05-22
:sparkles:  sync the labels from klusterlet to agent

zhiweiyin318:
## Summary
sync the labels from klusterlet to all agent resouces, the default labels `createdBy: klusterlet`

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/467

### DIFF
--- a/manifests/klusterlet/managed/klusterlet-registration-clusterrole-addon-management.yaml
+++ b/manifests/klusterlet/managed/klusterlet-registration-clusterrole-addon-management.yaml
@@ -7,6 +7,12 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: open-cluster-management:{{ .KlusterletName }}-registration:addon-management
+  labels:
+    {{ if gt (len .Labels) 0 }}
+    {{ range $key, $value := .Labels }}
+    {{ $key }}: {{ $value }}
+    {{ end }}
+    {{ end }}
 rules:
 # Allow agent to get/list/watch/create/delete/update/patch secrets.
 - apiGroups: [""]
--- a/pkg/operator/helpers/helpers.go
+++ b/pkg/operator/helpers/helpers.go
@@ -52,6 +52,9 @@ const (
 
 	// DefaultAddonNamespace is the default namespace for agent addon
 	DefaultAddonNamespace = "open-cluster-management-agent-addon"
+
+	// AgentLabelKey is used to filter resources in informers
+	AgentLabelKey = "createdByKlusterlet"
 )
 
 const (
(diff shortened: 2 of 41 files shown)


pkg/operator/helpers/helpers.go
```diff
@@ -821,3 +822,18 @@ func GetOperatorNamespace() string {
 	}
 	return operatorNamespace
 }
+
+func MergeLabels(destination map[string]string, source map[string]string) map[string]string {
```
qiujian16, line 826: can we just use MergeMap in library-go?

pkg/operator/helpers/helpers.go
```diff
@@ -821,3 +822,18 @@ func GetOperatorNamespace() string {
...
+	}
+
+	for k, v := range source {
+		destination[k] = v
+	}
+	return destination
+}
+
+func KlusterletAgentDefaultLabels() map[string]string {
+	return map[string]string{"createdBy": "klusterlet"}
```
qiujian16, line 838: could be just a var rather than a func?

pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller_test.go
```diff
@@ -463,6 +463,11 @@ func ensureObject(t *testing.T, object runtime.Object, klusterlet *operatorapiv1
 		return
 	}
 
+	if !mapCompare(helpers.MergeLabels(helpers.KlusterletAgentDefaultLabels(), klusterlet.GetLabels()), access.GetLabels()) {
```
qiujian16, line 466: I think you can just use deepEqual?

qiujian16 (commented):
let's add an integration test.

manifests/klusterlet/managed/klusterlet-registration-clusterrole-addon-management.yaml
```diff
@@ -7,6 +7,13 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: open-cluster-management:{{ .KlusterletName }}-registration:addon-management
+  labels:
+    createdBy: klusterlet
```
zhujian7, line 11: considering there might be multiple klusterlet in hosted mode, should we put the klusterlet as the real klusterlet name?

pkg/operator/helpers/helpers.go
zhiweiyin318: done

pkg/operator/helpers/helpers.go
zhiweiyin318: change this func to this:
```
func GetKlusterletAgentLabels(klusterlet *operatorapiv1.Klusterlet) map[string]string {
	labels := klusterlet.GetLabels()
	if labels == nil {
		labels = map[string]string{}
	}
	labels["klusterletName"] = klusterlet.Name
	// This label is used to filter in deployment informer
	labels["createdBy"] = "klusterlet"

	return labels
}
```

pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller_test.go
zhiweiyin318: the synced klusterlet labels is a part of the exiting labels, for example the deployments have label `app=klusterlet-registration-agent`, namespace has the label `workload.openshift.io/allowed=management`

manifests/klusterlet/managed/klusterlet-registration-clusterrole-addon-management.yaml
zhiweiyin318: the label `createdBy: klusterlet` is used to filter in deployment informer, I added another label klusterletName to set the real klusterlet name.

qiujian16: TBH maybe not necessary to create both label right now...why do we still need createdBy: klusterlet if we have klusterletName label? We can change the filter in deployment informer.

zhiweiyin318: how about to change the label to `createdByKlusterlet: <klusterlet name>` , in deployment informer, changed to check if the key `createdByKlusterlet` exists?

qiujian16:
@zhiweiyin318 can we merge this?

PR #481, opened 2024-05-29
📖 Update contributing guide

zhujian7:
## Related issue(s)

Fixes #

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,13 @@
   - [DCO Sign Off](#dco-sign-off)
   - [Code of Conduct](#code-of-conduct)
   - [Contributing a patch](#contributing-a-patch)
+  - [Setting up your dev environment, coding, and debugging](#setting-up-your-dev-environment-coding-and-debugging)
+    - [Prerequisites](#prerequisites)
+    - [Setting up dev environment](#setting-up-dev-environment)
+    - [Building Images](#building-images)
+    - [Testing the changes in the kind cluster](#testing-the-changes-in-the-kind-cluster)
+    - [Integration tests](#integration-tests)
+    - [E2E tests](#e2e-tests)
   - [Issue and pull request management](#issue-and-pull-request-management)
 - [References](#references)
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,40 +1,24 @@
...
-| work | staging |
-| registration-operator | staging |
-
-Status:
-* staging: the component is having all code in `staging` folder.
-* mix: the component is having code in both `staging` and outside.
-* done: the component is having all code outside `staging` folder and done of the migration.
-
-If a component is in the `staging` status, you need to open project in the `/staging` folder and contribute there.
+Before 0.11.0, the 5 components has independent repos, now we are going through [the task of consolidating code](https://github.com/open-cluster-management-io/OCM/issues/128) to merge all code into this repo to gain all kinds of benifits.
```
qiujian16, line 21: we have consolidate the code into ocm repo.

The task is done, so we should not it is going through.

zhujian7: removed this line

xuezhaojun:
👍lgtm

qiujian16:
do we need to apply this for all repos? do we also need to add a start dev enviornment section in this or in the website?

zhujian7:
@qiujian16 @xuezhaojun PTAL

lenicatko:
> do we need to apply this for all repos? do we also need to add a start dev enviornment section in this or in the website?

As a new comer to OCM, I noticed that each repo has its own contributing guide, but they are mostly the same. If it make sense and each repo requires something different, it is ok. But if not, it would be good to consolidate it. 

You have also a community repo, where you updated a contributor ladder, but it is not linked in any other repo. Try to focus on that repo, and make sure you make reference to it in each repo, so you don't need to update everything multiple times.

xuezhaojun:
lgtm to me

qiujian16:
/hold

should we move this to the community repo and just put a link here?

zhujian7:
@qiujian16 I think we can merge this into the ocm repo, since different repos have different developing process, (e.g API, addon-framework, clusteradm).

PR #482, opened 2024-05-29
:bug: Add event when stop creating CSR because of too many.

xuezhaojun:
## Summary

There are times when we do troubleshooting, find CSR is not triggered but no error msg in registration-agent pod logs.
It will be more friendly if we can have a event for this case.


qiujian16 (commented):
/approve
/lgtm

Issue #474, opened 2024-05-22
open-cluster-management-addon namespace should be optionally created


qiujian16:
**Describe the enhancement**
Today we alway create the open-cluster-management-addon ns for default addon ns. In some case, it should not be created at all, otherwise it will violate policy for certain environment.

We should provide a flag in the klusterlet to not create open-cluster-management-addon namespace.

PR #484, opened 2024-05-30
✨ Add a disable-default-addon-namespace flag

qiujian16:
if the flag is set, default addon ns will not be created by the operator.

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/474

### DIFF
--- a/pkg/cmd/spoke/operator.go
+++ b/pkg/cmd/spoke/operator.go
@@ -31,14 +31,20 @@ func NewKlusterletOperatorCmd() *cobra.Command {
 
 	// add disable leader election flag
 	flags := cmd.Flags()
-	cmd.Flags().BoolVar(&klOptions.SkipPlaceholderHubSecret, "skip-placeholder-hub-secret", false,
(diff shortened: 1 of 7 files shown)


clyang82 (commented):
what is behaviour if the `disable-default-addon-namespace` is set as `false` at the beginning, then update to `true`?

elgnay:
Should this line be updated as well to remove the the default add-on ns? https://github.com/qiujian16/ocm/blob/disable-addon-ns/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_namespace_reconcile.go#L42

qiujian16:
> what is behaviour if the disable-default-addon-namespace is set as false at the beginning, then update to true?

the namespace will be leftover, it is actually is safer.

qiujian16:
> Should this line be updated as well to remove the the default add-on ns? https://github.com/qiujian16/ocm/blob/disable-addon-ns/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_namespace_reconcile.go#L42

updated

PR #502, opened 2024-06-05
:bug: create an empty image pull secret if the secret is not existed

zhiweiyin318:
## Related issue(s)
https://github.com/open-cluster-management-io/clusteradm/issues/433

Fixes #


qiujian16:
can we do it in clusteradm instead of here?

We can also have flag to let user set docker credential when init and join

zhiweiyin318:
@qiujian16  good suggestion.

PR #505, opened 2024-06-06
:bug: add enable-sync-labels flag to klusterlet operator

zhiweiyin318:
## Summary
in gitops case, klusterlet is deployed by ArgoCD, so klusterlet has a label [argocd.argoproj.io/instance=xxx](http://argocd.argoproj.io/instance=xxx). If everything created klusterlet inherits this label, these objects will be triggered as being managed by ArgoCD but actually they are not.

fix: to add a flag to enable sync labels feature, the default is disable.

## Related issue(s)

Fixes #


zhiweiyin318:
cc @qiujian16 @EricaJ6

PR #506, opened 2024-06-06
🌱 Debug e2e with an existing cluster

zhujian7:
## Summary

Add an environment `USE_EXISTING_CLUSTER` for debugging e2e with act locally.
With this set true, we can create a kind cluster, and use `act -j e2e` to debug repeatedly, even if the test fails, the kind cluster will not be destroyed.
This will not affect the e2e running on the GitHub.

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -15,6 +15,7 @@ on:
 env:
   GO_VERSION: '1.21'
   GO_REQUIRED_MIN_VERSION: ''
+  USE_EXISTING_CLUSTER: false # set to true to use an existing kind cluster for debugging with act
 
 permissions:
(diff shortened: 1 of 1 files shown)


qiujian16:
might worth adding this in the readme on how to debug on act.

zhujian7:
> might worth adding this in the readme on how to debug on act.

@qiujian16 this is mentioned in another PR https://github.com/open-cluster-management-io/ocm/pull/481/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R187

qiujian16:
ah cool, thanks.
/approve
/lgtm

PR #499, opened 2024-06-05
:seedling: simplify the addon progressing message

haoqing0110:
## Related issue(s)

Fixes #

PR #513, opened 2024-06-12
:bug:  update mca progressing condition when desired and last applied matches

haoqing0110:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/addon/controllers/addonprogressing/controller.go
+++ b/pkg/addon/controllers/addonprogressing/controller.go
@@ -283,20 +283,12 @@ func workIsReady(work *workapiv1.ManifestWork) bool {
 	return true
 }
 
-// set addon progressing condition and last applied
+// Set addon progressing condition and last applied
+// Once completed, the LastAppliedConfig will set to DesiredConfig. Even the condition status
+// changes later, eg, spoke cluster becomes unhealthy, status becomes progressing or failed after completed,
+// the LastAppliedConfig and DesiredConfig won't change. this is to avoid back and forth in rollout.
+// The setRolloutStatus() func in graph.go only look at the LastAppliedConfig and DesiredConfig.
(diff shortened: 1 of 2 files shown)


pkg/addon/controllers/addonprogressing/controller.go
```diff
@@ -285,18 +285,6 @@ func workIsReady(work *workapiv1.ManifestWork) bool {
 
 // set addon progressing condition and last applied
 func setAddOnProgressingAndLastApplied(reason string, message string, addon *addonapiv1alpha1.ManagedClusterAddOn) {
-	// always update progressing condition when there is no config
```
qiujian16, line 288: why do we need to remove this skip?

haoqing0110: Meet an issue: when addon applied configs successfully and then configs become unaccessible (eg, removed), the desired spec hash will be empty and report "mca and work configs mismatch", and when the configs added back, since lastapplied already the same as desired, so the error message will leave in the condition and can never trigger an update. 

2 reasons feel save to remove the skip: 
- In the previous code, add a skip to avoid "install" message changes to "upgrade" after a fresh install. With PR https://github.com/open-cluster-management-io/ocm/pull/499/files merged, won't have such issue. 
- Another reason is once the mca applied config successfully (lastapplied = desired), do not want to change the "Progressing" condition any more, to avoid back and forth in rollout. Feel save to remove the skip because the [setRolloutStatus](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/addon/controllers/addonconfiguration/graph.go#L54) only look at the lastapplied and desired, not message.

qiujian16: could we add some comments here?

haoqing0110: added.

PR #518, opened 2024-06-13
✨ Add well known status rule for daemonsets

zhujian7:
## Related issue(s)

Fixes #

### DIFF
--- a/test/integration/cloudevents/statusfeedback_test.go
+++ b/test/integration/cloudevents/statusfeedback_test.go
@@ -541,4 +541,186 @@ var _ = ginkgo.Describe("ManifestWork Status Feedback", func() {
 			}, eventuallyTimeout, eventuallyInterval).ShouldNot(gomega.HaveOccurred())
 		})
 	})
+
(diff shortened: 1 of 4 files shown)


test/integration/cloudevents/statusfeedback_test.go
```diff
@@ -541,4 +541,186 @@ var _ = ginkgo.Describe("ManifestWork Status Feedback", func() {
 			}, eventuallyTimeout, eventuallyInterval).ShouldNot(gomega.HaveOccurred())
 		})
 	})
+
+	ginkgo.Context("DaemonSet Status feedback", func() {
```
qiujian16, line 545: @skeeey why do we still need to create separated integration test case for cloudevent?

qiujian16 (commented):
/approve
/lgtm

test/integration/cloudevents/statusfeedback_test.go
skeeey: the could events supports single manifest, for single manifest, some integration cannot pass

PR #524, opened 2024-06-16
:sparkles:Operator support MultipleHubs.

xuezhaojun:
## Summary

The PR is to support MultiHubs also in the operator.

The PR is related to: https://github.com/open-cluster-management-io/ocm/pull/443

### DIFF
--- a/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
+++ b/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
@@ -135,7 +135,10 @@ type klusterletConfig struct {
 	ClusterName                                 string
 	ExternalServerURL                           string
 	HubKubeConfigSecret                         string
+	MultipleHubs                                bool
 	BootStrapKubeConfigSecret                   string
+	BootStrapKubeConfigSecrets                  []string
+	HubConnectionTimeoutSeconds                 int32
 	OperatorNamespace                           string
 	Replica                                     int32
 	ClientCertExpirationSeconds                 int32
@@ -174,6 +177,26 @@ type klusterletConfig struct {
 	Labels map[string]string
 }
 
+// If multiplehubs feature gate is enabled, using the bootstrapkubeconfigs from klusterlet CR.
+// Otherwise, using the default bootstrapkubeconfig.
+func (config *klusterletConfig) populateBootstrap(klusterlet *operatorapiv1.Klusterlet) {
+	config.MultipleHubs = klusterlet.Spec.RegistrationConfiguration != nil &&
+		helpers.FeatureGateEnabled(klusterlet.Spec.RegistrationConfiguration.FeatureGates, ocmfeature.DefaultSpokeRegistrationFeatureGates, ocmfeature.MultipleHubs)
+
+	if config.MultipleHubs {
+		var bootstapKubeconfigSecrets []string
+		if klusterlet.Spec.RegistrationConfiguration.BootstrapKubeConfigs.Type == operatorapiv1.LocalSecrets {
+			for _, secret := range klusterlet.Spec.RegistrationConfiguration.BootstrapKubeConfigs.LocalSecrets.KubeConfigSecrets {
+				bootstapKubeconfigSecrets = append(bootstapKubeconfigSecrets, secret.Name)
+			}
+		}
+		config.BootStrapKubeConfigSecrets = bootstapKubeconfigSecrets
+		config.HubConnectionTimeoutSeconds = klusterlet.Spec.RegistrationConfiguration.BootstrapKubeConfigs.LocalSecrets.HubConnectionTimeoutSeconds
+	} else {
+		config.BootStrapKubeConfigSecret = helpers.BootstrapHubKubeConfig
+	}
+}
+
 func (n *klusterletController) sync(ctx context.Context, controllerContext factory.SyncContext) error {
 	klusterletName := controllerContext.QueueKey()
 	klog.V(4).Infof("Reconciling Klusterlet %q", klusterletName)
@@ -207,7 +230,6 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		WorkImage:                              klusterlet.Spec.WorkImagePullSpec,
 		ClusterName:                            klusterlet.Spec.ClusterName,
 		SingletonImage:                         klusterlet.Spec.ImagePullSpec,
-		BootStrapKubeConfigSecret:              helpers.BootstrapHubKubeConfig,
 		HubKubeConfigSecret:                    helpers.HubKubeConfig,
 		ExternalServerURL:                      getServersFromKlusterlet(klusterlet),
 		OperatorNamespace:                      n.operatorNamespace,
@@ -228,6 +250,8 @@ func (n *klusterletController) sync(ctx context.Context, controllerContext facto
 		ResourceRequirements:            resourceRequirements,
 		DisableAddonNamespace:           n.disableAddonNamespace,
 	}
+
+	config.populateBootstrap(klusterlet)
 
 	if n.enableSyncLabels {
 		config.Labels = helpers.GetKlusterletAgentLabels(klusterlet)
--- a/pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
+++ b/pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
@@ -19,6 +19,7 @@ import (
 	operatorv1client "open-cluster-management.io/api/client/operator/clientset/versioned/typed/operator/v1"
 	operatorinformer "open-cluster-management.io/api/client/operator/informers/externalversions/operator/v1"
 	operatorlister "open-cluster-management.io/api/client/operator/listers/operator/v1"
+	ocmfeature "open-cluster-management.io/api/feature"
 	operatorapiv1 "open-cluster-management.io/api/operator/v1"
 	"open-cluster-management.io/sdk-go/pkg/patcher"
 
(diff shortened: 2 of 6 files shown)


pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
```diff
@@ -143,30 +144,34 @@ func (c *ssarController) sync(ctx context.Context, controllerContext factory.Syn
...
-			ObservedGeneration: klusterlet.Generation,
-			Reason:             bootstrapDegradedCondition.Reason + "," + hubConfigDegradedCondition.Reason,
-			Message:            bootstrapDegradedCondition.Message + "\n" + hubConfigDegradedCondition.Message,
-		})
-		_, err := c.patcher.PatchStatus(ctx, newKlusterlet, newKlusterlet.Status, klusterlet.Status)
-		if err != nil {
-			klog.Errorf("Update Klusterlet Status Failed: %v", err)
-			controllerContext.Queue().AddAfter(klusterletName, SSARReSyncTime)
+		// If the MultipleHubs feature is enabled, there is no need to check the bootstrap kubeconfig secret.
+		// The ssar check to bootstrapkubeconfigs is done at the start of the registration agent.
```
qiujian16, line 148: I am not so sure why we need to do this. What is the impact if we alway check the bootstrap secret?

test/e2emultiplehubs/multiplehubs_test.go
```diff
@@ -0,0 +1,230 @@
...
+	flag.StringVar(&registrationImage, "registration-image", "", "The image of the registration")
+	flag.StringVar(&workImage, "work-image", "", "The image of the work")
+	flag.StringVar(&singletonImage, "singleton-image", "", "The image of the klusterlet agent")
+
+	flag.StringVar(&hub1Kubeconfig, "hub1-kubeconfig", "", "The kubeconfig of the hub cluster 1")
+	flag.StringVar(&hub2Kubeconfig, "hub2-kubeconfig", "", "The kubeconfig of the hub cluster 2")
+	flag.StringVar(&spokeKubeconfig, "spoke-kubeconfig", "", "The kubeconfig of the spoke cluster")
+}
+
+func TestE2E(t *testing.T) {
```
qiujian16, line 56: put test main in the separated files

test/testhelper/common.go
```diff
@@ -0,0 +1,141 @@
+package testhelper
```
qiujian16, line 1: this is like a refactor of the existing test util funcs. We should put this in the separated PR and refactor the existing code if needed.

test/e2emultiplehubs/multiplehubs_test.go
```diff
@@ -0,0 +1,230 @@
...
+				clusterv1.ManagedClusterConditionAvailable:   metav1.ConditionTrue,
+				clusterv1.ManagedClusterConditionJoined:      metav1.ConditionTrue,
+				clusterv1.ManagedClusterConditionHubAccepted: metav1.ConditionTrue,
+			})
+		}, 5*time.Minute, 5*time.Second).Should(Succeed())
+
+		// TODO: @xuezhaojun after the agent switch to the hub2, the 'managedcluster' on the hub1 keep showing `Available` status
+		// This is because:
+		// * the lease contoller only watch managedcluster resource, no periodical check on the managedcluster lease
+		// * if the `hubAcceptsClient` is false, the controller will skip the reconciliation
```
qiujian16, line 200: this looks like a bug cc @elgnay

.github/workflows/e2e.yml
```diff
@@ -135,3 +135,63 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: install imagebuilder
+        run: go install github.com/openshift/imagebuilder/cmd/imagebuilder@v1.2.3
+      - name: Build images
+        run: IMAGE_TAG=e2e make images
+      - name: Setup kind cluster
```
qiujian16, line 154: setup environment in action would be a bit hard for local dev testing. We should probably prepare a script in hack to setup the local multi-hubs environment.

pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -174,6 +177,26 @@ type klusterletConfig struct {
 	Labels map[string]string
 }
 
+// If multiplehubs feature gate is enabled, using the bootstrapkubeconfigs from klusterlet CR.
+// Otherwise, using the default bootstrapkubeconfig.
+func (config *klusterletConfig) populateBootstrap(klusterlet *operatorapiv1.Klusterlet) {
+	klusterlet = klusterlet.DeepCopy()
```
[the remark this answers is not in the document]
xuezhaojun: You're right, there is `klusterlet := originalKlusterlet.DeepCopy()` at the begining of the sync. Line removed.

pkg/operator/helpers/helpers.go
```diff
@@ -790,6 +790,20 @@ func ConvertToFeatureGateFlags(component string, features []operatorapiv1.Featur
 	return flags, ""
 }
 
+func IsRegistrationFeatureGateEnabled(rc *operatorapiv1.RegistrationConfiguration, featureName featuregate.Feature) bool {
```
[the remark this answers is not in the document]
xuezhaojun: The `ConvertToFeatureGateFlags` is conveient for registration and work, but the `IsRegistrationFeatureGateEnabled ` is used in the `**operator**` part.

I'm making the operator and registartion-agent using the same featureGate which is the MultipleHubs in the `ClusterManager.Spec.RegistrataionConfiguration.FeatureGates`.

pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_controller.go
```diff
@@ -174,6 +177,26 @@ type klusterletConfig struct {
...
+	if helpers.IsRegistrationFeatureGateEnabled(klusterlet.Spec.RegistrationConfiguration, ocmfeature.MultipleHubs) {
```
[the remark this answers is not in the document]
xuezhaojun: Done.

pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
xuezhaojun: If MultipleHubs is enabled, the legacy bootstrap-kubeconfig then is not required to be provided.

And we will get a `BootstrapSecretMissing` condition.

test/e2emultiplehubs/multiplehubs_test.go
xuezhaojun: Done.

test/testhelper/common.go
xuezhaojun: Agree, the refactor on legacy e2e is on the way. I will have several little PRs to do that. But I don't think this will be a blocker for this PR, we can do 2 in parallel.

.github/workflows/e2e.yml
xuezhaojun: Agree, the script for local-test is added in hack dir.

pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
qiujian16: can we set bootstrap-kubeconfig file as the var for this controller, so we can specify the bootstrap kubeconfig when starting the controller.

xuezhaojun: I don't think klusterlet_ssar_controller as part of the operator can depend bootstrap-kubeconfig file. 

Using the boostrapkubeconfigs localSecrets is possible, but we should consider new condition, for example: when hub1 bootstrapkubeconfig is incorrect, but hub2 bootstrapkubeconfig is fine, do we want to consider it as a "HubConnectionDegraded" False.

xuezhaojun:
@qiujian16 The e2e part is removed and will join the refactor PR, the intergration test added, please take another look, thanks!

pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
qiujian16: if it is not file, we can use secret? The gap is we do not know which secret is picked?

pkg/operator/helpers/helpers.go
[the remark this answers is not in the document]
qiujian16: can we just set config.MultipleHubs = helpers.FeatureGateEnabled(...

and later use config.MultipleHubs in other places?

pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
qiujian16: should we just check all the secrets in this case? It will give user the info on which hub is not working right now.

xuezhaojun: We can get the secrets from the Klusterelt.Spec, the gap is even we have the secrets, it can not act the same as the default bootstrap-hub-kubeconfig, for example:

We have a informer running against the default bootstrap-hub-kubeconfig, if the default bootstrap-hub-kubeconfig become invalid, the reconcile will trigger to update the `HubConnectionDegraded` condition: https://github.com/open-cluster-management-io/ocm/blob/a14450aa9346782a75bc4d102379715b3810740c/pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go#L67 

But for the bootstrapkubeconfigs case, there is no informer for these secrets. We can add one, that means the controller will go back to watch all secrets.

---

And the `HubConnectionDegraded` condition is depended by klusterlet-controller to decide whether to deploy the work-agent: https://github.com/open-cluster-management-io/ocm/blob/a14450aa9346782a75bc4d102379715b3810740c/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_runtime_reconcile.go#L94-L106
```go
// Deploy work agent.
	// * work agent is scaled to 0 only when degrade is true with the reason is HubKubeConfigSecretMissing.
	//   It is to ensure a fast startup of work agent when the klusterlet is bootstrapped at the first time.
	// * The work agent should not be scaled to 0 in degraded condition with other reasons,
	//   because we still need work agent running even though the hub kubconfig is missing some certain permission.
	//   It can ensure work agent to clean up the resources defined in manifestworks when cluster is detaching from the hub.
	hubConnectionDegradedCondition := meta.FindStatusCondition(klusterlet.Status.Conditions, operatorapiv1.ConditionHubConnectionDegraded)
	if hubConnectionDegradedCondition == nil {
		workConfig.Replica = 0
	} else if hubConnectionDegradedCondition.Status == metav1.ConditionTrue &&
		strings.Contains(hubConnectionDegradedCondition.Reason, operatorapiv1.ReasonHubKubeConfigSecretMissing) {
		workConfig.Replica = 0
	}
```

So, if we want to let user get the connectivity status from the condition, I suggest to add a new Condition instead of adding it to the HubConnectionDegradedCondition.

xuezhaojun: So the questions are:
* Do we want to watch all secrets or bootstrapkubeconfigs secrets(means operator needs to add a label to the secrets configured in Klusterlet.Spec.RegistrationConfiguration.BootstrapKubeConfig.LocalSecrets)?
* Do we add want to add a new Condition to represent the connectivity status in multiplehubs case?

pkg/operator/helpers/helpers.go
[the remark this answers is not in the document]
xuezhaojun: The function is used by 2 controllers:
* the klusterlet_controller.go: used to render the manifests
* the klusterlet_ssar_controller.go:  used to skip bootstrapDegradedCondition when multiplehubs.

Where we set `config.MultipleHubs = helpers.FeatureGateEnabled` is in the klusterlet_controller.go, but the config here is a struct only used in render, I don't think it can also be used in the klusterlet_ssar_controller.

---

The `IsRegistrationFeatureGateEnabled ` is replace to `FeatureGateEnabled` to align with other usages.

pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go
qiujian16: > Do we want to watch all secrets or bootstrapkubeconfigs secrets(needs operator to add label on them)? 
we cannot watch all secrets. But the operator should be able to read the secrets in a certain namespace. 

> Do we want to add a new condition represent the connectivity status of bootstrapkubeconfigs?
existing status is good enough. The invalid bootstrap will not make HubConnectionDegraded to true

xuezhaojun: > we cannot watch all secrets. But the operator should be able to read the secrets in a certain namespace.

Yes, but we get the agentNamespace from the Klusterlet by:

https://github.com/open-cluster-management-io/ocm/blob/fc9b2d96e1f49b16e03115c0d0dfa952f374dad3/pkg/operator/helpers/helpers.go#L608C1-L626C2

That means we don't know which agentnamespace the klusterlet will use at the start of the operator.

>  existing status is good enough. The invalid bootstrap will not make HubConnectionDegraded to true

Get it, just need confirm the details of the the condition, we mean:

**case1**: when multiplehubs is enabled, **hub-kubeconfig works fine**, and **none** of the bootstraps in the list is invalid, we want the HubConnectionDegraded be:

```go
metav1.Condition{
      Status: metav1.ConditionFalse,
      Reason: operatorapiv1.ReasonHubConnectionFunctional,
      Message: fmt.Sprintf("Hub kubeconfig secret %s/%s to apiserver %s is working",
			agent.namespace, helpers.HubKubeConfig, host),
}
```

**case2**: when multiplehubs is enabled, **hub-kubeconfig works fine**, and **some** of the bootstraps in the list are invalid, we want the HubConnectionDegraded be:

```go
metav1.Condition{
      Status: metav1.ConditionFalse,
      Reason: operatorapiv1.ReasonHubConnectionFunctional,
      Message: fmt.Sprintf("BootstrapKubeConfig Secrets:[%v] are getting error, %v", invaildBootstrapSecrets, err),
}
```

**case3**: when multiplehubs is enabled, **hub-kubeconfig doesn't function**, and **some** of the bootstraps in the list are invalid, we want the HubConnectionDegraded be: 

```go
metav1.Condition{
      Status: metav1.ConditionTrue,
      Reason: "BootstrapSecretsError",
      Message: fmt.Sprintf("BootstrapKubeConfig Secrets:[%v] are getting error, %v", invaildBootstrapSecrets, errs),
}
```

To me case2 is a bit odd because the reason is Functional, the status is False, but the message shows some of the BootstrapKubeConfigSecrets have trouble.

qiujian16: case 2 is helpful for user to know which bootstrap kubeconfig is not working.

xuezhaojun: That's fair, another concern is: we skip the ssar check for default bootstrapkubeconfig when hub kubeconfig is functional to avoid sending too many requests to apiserver, case2 means we are going to regression this point in the multiplehubs enabled case.

```
		// the hub kubeconfig is functional, the bootstrap kubeconfig check is not needed,
		// ignore it to avoid sending additional sar requests
		if hubConfigDegradedCondition.Status == metav1.ConditionFalse {
			meta.SetStatusCondition(&newKlusterlet.Status.Conditions, hubConfigDegradedCondition)
			_, err := c.patcher.PatchStatus(ctx, newKlusterlet, newKlusterlet.Status, klusterlet.Status)
			if err != nil {
				klog.Errorf("Update Klusterlet Status Failed: %v", err)
				controllerContext.Queue().AddAfter(klusterletName, SSARReSyncTime)
			}

			return
		}
```
https://github.com/open-cluster-management-io/ocm/blob/fc9b2d96e1f49b16e03115c0d0dfa952f374dad3/pkg/operator/operators/klusterlet/controllers/ssarcontroller/klusterlet_ssar_controller.go#L133C1-L144C4

---

If our ultimate intention is to let user know which bootstrap kubeconfig is not working, how about do it periodically? So we can 2r regression:
* Avoid watching all secrets
* Avoid sending addtional ssar when hub-kubeconfig is functional.

xuezhaojun: To fix the first issue, how about adding a controller to add label on bootstrapkubeconfig secret, so that we can create a informer against this label?

xuezhaojun: Decide to delay the design at the moment. 
Adding a TODO, we need to find a way to reflect the connectivity status of bootstrapkubeconfigs.

PR #525, opened 2024-06-17
🌱 make additional secret data always sensitive

elgnay:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/spoke/spokeagent.go
+++ b/pkg/registration/spoke/spokeagent.go
@@ -11,7 +11,6 @@ import (
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


pkg/registration/spoke/spokeagent.go
```diff
@@ -244,12 +243,6 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 	// in scenario #2 and #3, which results in an error message in log: 'Observed a panic: timeout waiting for
 	// informer cache'
 	if !ok {
-		// delete the hub kubeconfig secret if exits
```
zhujian7, line 247: Should we add a note in line 290 that if the secret is not valid, the xxx controller will update it to a valid one?

zhujian7: And maybe more note that waiting forever in line 290 is because the cluster should be approved on the hub side?

elgnay: Comments added as suggested.

qiujian16:
/hold
we should merge this after release of 0.14.0

PR #532, opened 2024-06-19
🌱 Update deps of api and library to 0.14.0

qiujian16:
## Related issue(s)

Fixes #


test/integration/registration/addon_registration_test.go
```diff
@@ -395,8 +384,6 @@ var _ = ginkgo.Describe("Addon Registration", func() {
 		ginkgo.It("should register addon failed with invalid custom signer", func() {
```
zhujian7, line 384: Is this ginkgo.It case necessary now? Should we remove it?

qiujian16: removed

PR #533, opened 2024-06-19
🐛 Add install namespace to built in values

zhujian7:
## Summary

Add `INSTALL_NAMESPACE` to the built in values if the private value `__INSTALL_NAMESPACE` is set.

## Related issue(s)

Fixes #


qiujian16 (commented):
do we still need to keep the private install namespace value?

zhujian7:
> do we still need to keep the private install namespace value?

@qiujian16 I'm not sure, but I just want to keep the changes as small as possible here.

qiujian16:
OK I think we need to refactor this a bit in the next release.

/approve
/lgtm

PR #534, opened 2024-06-19
:seedling:Refactor makefile e2e commands.

xuezhaojun:
## Summary

Except for the standard e2e prepare process: `test-e2e`.

We may also want to customized the process, for example:

```
# deploy ocm hub on hub1
make deploy-hub KUBECONFIG=/home/runner/.kube/hub1 HUB_KUBECONFIG=/home/runner/.kube/hub1-kubeconfig

# deploy ocm hub on hub2
make deploy-hub KUBECONFIG=/home/runner/.kube/hub2 HUB_KUBECONFIG=/home/runner/.kube/hub2-kubeconfig

# deploy ocm agent on spoke
make deploy-spoke-operator KUBECONFIG=/home/runner/.kube/spoke

run-e2e
```

The changes intend to make `deploy-hub` and `run-e2e` commands more independent, so we can reuse them in different e2e prepare processes.

### DIFF
--- a/test/e2e-test.mk
+++ b/test/e2e-test.mk
@@ -18,7 +18,7 @@ endif
 hub-kubeconfig:
 	$(KUBECTL) config view --minify --flatten > $(HUB_KUBECONFIG)
 
-deploy-hub: deploy-hub-operator apply-hub-cr hub-kubeconfig
+deploy-hub: deploy-hub-operator apply-hub-cr hub-kubeconfig cluster-ip
 
 deploy-hub-operator: ensure-kustomize
 	cp deploy/cluster-manager/config/kustomization.yaml deploy/cluster-manager/config/kustomization.yaml.tmp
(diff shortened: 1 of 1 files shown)


test/e2e-test.mk
```diff
@@ -42,11 +45,6 @@ clean-spoke: clean-spoke-cr clean-spoke-operator
 clean-e2e:
 	$(RM) ./e2e.test
 
-cluster-ip:
```
xuezhaojun, line 45: The only usage of `cluster-ip` is before `run-e2e`. 
And the only usage of `run-e2e` is the last step of `test-e2e`.

That's mean all HUB_KUBECONFIG must be processed by `cluster-ip` in our e2e. That's the reason I conbine these 2 commands into 1.

And I also move `bootstrap-secret` to `test-e2e` to make `run-e2e` more indenpendent. The reason is test-e2e represents a default/standard way of preparing e2e testing env.   But for cases that we want to do some motifications, we can reuse the `run-e2e`.

PR #535, opened 2024-06-20
✨ Refactor registration

qiujian16:
## Related issue(s)

Fixes #


pkg/registration/spoke/spokeagent.go
```diff
@@ -246,55 +268,27 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 		// create a ClientCertForHubController for spoke agent bootstrap
 		// the bootstrap informers are supposed to be terminated after completing the bootstrap process.
 		bootstrapInformerFactory := informers.NewSharedInformerFactory(bootstrapKubeClient, 10*time.Minute)
```
zhujian7, line 270: the comments should be updated. there is no `ClientCertForHubController` now. And please check comments in other places.

pkg/registration/register/interface.go
```diff
@@ -0,0 +1,96 @@
...
+type StatusUpdateFunc func(ctx context.Context, cond metav1.Condition) error
+
+// RegisterDriver is the interface that each driver should implement
+type RegisterDriver interface {
+	// IsHubKubeConfigValid is to check if the current hube-kubeconfig is valid. It is called before
+	// and after bootstrap to confirm if the bootstrap is finished.
+	IsHubKubeConfigValid(ctx context.Context, secretOption SecretOption) (bool, error)
+
+	// BuildKubeConfigFromTemplate builds the kubeconfig from the template kubeconfig
+	BuildKubeConfigFromTemplate(config *clientcmdapi.Config) *clientcmdapi.Config
```
zhujian7, line 62:
```suggestion
	BuildKubeConfigFromTemplate(template *clientcmdapi.Config) *clientcmdapi.Config
```

zhujian7: Or is it better to make the different driver only provide the auth-info? so the `default-auth` key will always match the key defined in the `BaseKubeConfigFromBootStrap` template kubeconfig context.

qiujian16: some implementation may also add additional field rather than authinfo, such as execconfig or execvar.

PR #556, opened 2024-07-05
:seedling: Explicitly rename and use universal klusterlet/managed cluster.

xuezhaojun:
## Summary

Cases are using a shared, universal klusterlet/managed cluster, rename the field to expilicitly hight this point.

Also:
* Remove the clusterName, klusterletName, agentNamespace from flags since they will always be named by: 

https://github.com/open-cluster-management-io/ocm/blob/b3f15c14f483884bb12fee7528321ef6e1d53608/test/e2e/e2e_suite_test.go#L85-L87

* Remove deploy-klusterlet from the flags since it's always true in our test.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"testing"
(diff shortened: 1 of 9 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -1,48 +1,67 @@
...
+// but also pay attention, because the klusterlet is shared, so the developers should not delete the klusterlet.
+// And there might be some side effects on other cases if the developers change the klusterlet's spec for their cases.
+var (
+	universalClusterName    string
+	universalKlusterletName string
+	universalAgentNamespace string
+)
+
+const (
+	UNIVERSAL_CLUSTERSET = "universal"
```
qiujian16, line 62: why this is public?

test/e2e/e2e_suite_test.go
```diff
@@ -1,48 +1,67 @@
...
-	flag.StringVar(&klusterletDeployMode, "klusterlet-deploy-mode", string(operatorapiv1.InstallModeDefault), "The image of the work")
+
+	flag.DurationVar(&eventuallyTimeout, "eventually-timeout", 60*time.Second, "The timeout of Gomega's Eventually (default 60 seconds)")
 }
 
+// The e2e will always create one universal klusterlet, the developers can reuse this klusterlet in their case
+// but also pay attention, because the klusterlet is shared, so the developers should not delete the klusterlet.
+// And there might be some side effects on other cases if the developers change the klusterlet's spec for their cases.
+var (
+	universalClusterName    string
```
qiujian16, line 56: can we also define the default value here?

test/e2e/e2e_suite_test.go
xuezhaojun: Right, it shouldn't be.

test/e2e/e2e_suite_test.go
xuezhaojun: Yes we can, updated.

qiujian16 (commented):
/approve
/lgtm

PR #565, opened 2024-07-10
:seedling: Refactor Tester to e2e framework pkg.

xuezhaojun:
## Summary

In this PR, we refactor the Tester into a e2e framework pkg. The pkg includes:
* 2 main abstraction Hub and Spoke, they holds the all clients that need to interact with a hub or a spoke, the metadata, and the runtime data.
* Files that contains functions of checking, creating, and etc of a specific aspect. For example, under the `managedcluster.go` file, there are `GetManagedCluster` and `CheckManagedClusterStatus`.

The default `e2e` has 1 hub and 1 spoke, the `hub` and `spoke` are global varible so developer can easily use them, for example, if in your case, you want to do something on the ManagedCluster, you can do:
```go
mc, err := hub.ClusterClient.ClusterV1().ManagedClusters().Get(context.TODO(),
		clusterName, metav1.GetOptions{})
```

You can also check the file `test/e2e/framework/managedcluster.go` to see if there is already a function implement your intention. All functions are seperated by different aspect so it's easier to find.


xuezhaojun:
/assign @qiujian16 
PR updated, please take another look, thanks!

test/framework/featuregate.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	ocmfeature "open-cluster-management.io/api/feature"
+	operatorapiv1 "open-cluster-management.io/api/operator/v1"
+)
+
+func EnableWorkFeature(hub *Hub, feature string) error {
```
qiujian16, line 13: can make this as enableHubFeature, work feature could be on hub and spoke.

test/framework/hub.go
```diff
@@ -0,0 +1,95 @@
...
+	return &Hub{
+		OCMClients: clients,
+		// the name of the ClusterManager object is constantly "cluster-manager" at the moment; The same name as deploy/cluster-manager/config/samples
+		ClusterManagerName:      "cluster-manager",
+		ClusterManagerNamespace: helpers.ClusterManagerDefaultNamespace,
+		ClusterCfg:              clusterCfg,
+	}, nil
+}
+
+func CheckHubReady(hub *Hub) error {
```
qiujian16, line 45: could this be a func in hub struct rather than using hub as an input.

test/framework/klusterlet.go
```diff
@@ -0,0 +1,259 @@
...
+		klusterlet, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		klog.Errorf("failed to create klusterlet %v . %v", klusterlet.Name, err)
+		return nil, err
+	}
+
+	return realKlusterlet, nil
+}
+
+func CreatePureHostedKlusterlet(spoke *Spoke, name, clusterName string) (*operatorapiv1.Klusterlet, error) {
```
qiujian16, line 157: why this func, and why this is different from CreateKlusterlet. When should we use this?

test/framework/klusterlet.go
```diff
@@ -0,0 +1,259 @@
...
+		klusterlet, metav1.CreateOptions{})
+	if err != nil {
+		klog.Errorf("failed to create klusterlet %v . %v", klusterlet.Name, err)
+		return nil, err
+	}
+
+	return realKlusterlet, nil
+}
+
+func CheckKlusterletStatus(spoke *Spoke, klusterletName, condType, reason string, status metav1.ConditionStatus) error {
```
qiujian16, line 192: spoke already has the klusterName set in it, shouldn't this as a func under spoke struc?

test/framework/klusterlet.go
```diff
@@ -0,0 +1,259 @@
...
+
+	if cond.Status != status {
+		return fmt.Errorf("condition status is not matched, expect %s, got %s", status, cond.Status)
+	}
+
+	return nil
+}
+
+// CleanKlusterletRelatedResources needs both hub side and spoke side operations
+func CleanKlusterletRelatedResources(
```
qiujian16, line 215: we actually should have a clean func for spoke and hub. Something like spoke.Cleanup() and hub.Cleaup, and call these two in this func.

test/framework/managedcluster.go
```diff
@@ -0,0 +1,190 @@
...
+	certificatesv1 "k8s.io/api/certificates/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+func GetManagedCluster(hub *Hub, clusterName string) (*clusterv1.ManagedCluster, error) {
```
qiujian16, line 17: If these are all hub side operation, shouldn't we make it as func in hub struct?

test/framework/klusterlet.go
```diff
@@ -0,0 +1,259 @@
...
+
+	operatorapiv1 "open-cluster-management.io/api/operator/v1"
+
+	"open-cluster-management.io/ocm/pkg/operator/helpers"
+)
+
+// CreateAndApproveKlusterlet requires operations on both hub side and spoke side
+func CreateAndApproveKlusterlet(
+	hub *Hub, spoke *Spoke,
+	klusterletName, managedClusterName, klusterletNamespace string,
```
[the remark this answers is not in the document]
xuezhaojun: That's a good point.

The spoke has:
* `KlusterletOperatorNamespace`
* `KlusterletOperator`

Which is different from `KlusterletNamespace` and `Klusterlet`.

One represent the operator deployment, one represent a Klusterlet.

test/framework/spoke.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	clientcmd "k8s.io/client-go/tools/clientcmd"
+)
+
+// Spoke represents a spoke cluster, it holds the clients to interact with the spoke cluster
+type Spoke struct {
+	*OCMClients
+	KlusterletOperatorNamespace string
+	KlusterletOperator          string
```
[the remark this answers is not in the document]
xuezhaojun: The spoke cluster will be deployed multiple Klusterlet as well as managed cluster in different namespace during the e2e.  The spoke to managedcluster is 1-N so we shouldn't have clusterName in Spoke.

test/framework/klusterlet.go
xuezhaojun: That's a good question but may beyond this PR's scope, the function is directly copied from Tester and I will try to answer.

Comparing with the `CreateKlusterlet` which not only create Klusterlet but also create the related reasources like namespace, bootstrapkubeconfigsecret, externalKubeconfigSecret, the `CreatePureHostedKlusterlet ` only create Klusterlet, no another thing.

It seems there is a case that requires only Hosted mode Klusterlet, but don't want the external managed kubeconfig, that's why we want a 'Pure' function.
```
	It("Delete klusterlet CR in Hosted mode without external managed kubeconfig", func() {
		By(fmt.Sprintf("create klusterlet %v with managed cluster name %v in Hosted mode", klusterletName, clusterName))
		_, err := CreatePureHostedKlusterlet(spoke, klusterletName, clusterName)
		Expect(err).ToNot(HaveOccurred())
```

I can add a TODO on this function and the `CreateKlusterlet` to try to merge them into one if we need.

test/framework/klusterlet.go
xuezhaojun: The spoke has the name of `KlusterletOperator` instead of `Klusterlet`. The same reason in https://github.com/open-cluster-management-io/ocm/pull/565#discussion_r1675258355

test/framework/klusterlet.go
xuezhaojun: Agree, this is something simply copied and pasted in the PR.

But in this case, it's not like hub.Cleanup is suitable, because hub.Cleanup is like a entire hub Cleanup, in this function, we only want to clean up the managedcluster created by the klusterlet on the hub.

And also `spoke != klustelret`, spoke has multiple klusterlet during the e2e.

I think we can have hub.Cleanup and spoke.Cleanup but not for this case.

test/framework/managedcluster.go
xuezhaojun: That's a good question, I think a lot on this, whether the `GetManagedCluster` and other similar code blocks be a method of hub or a function.

The reason 1 is to keep the defination and scope of a `hub` clear, hub is only:
* the client
* the metadata
* the runtime data
It provide information for us to do things, but not do things for us.

The reason 2 is,  the `GetManagedCluster` actually doesn't need the whole hub, it only needs the `clusterclient`,  and if we have:

```go
func GetManagedCluster(clusterclient clusterclient.Interface, clusterName string) (*clusterv1.ManagedCluster, error) {
	if clusterName == "" {
		return nil, fmt.Errorf("the name of managedcluster should not be null")
	}
	return clusterclient.ClusterV1().ManagedClusters().Get(context.TODO(), clusterName, metav1.GetOptions{})
}
```

That means we can reuse this function and this framework not only in e2e, but also the intergration test, the addon repo's e2e and integration, the downstream repos that replies on ocm.

Currently if we develop a addon or a component that relaies on ocm, we write integration and e2e part(code and scripts) indenependly but if we can provide a ocm-test-framework, I think it can reduce some effort in write and maintaining these tests.

test/framework/hub.go
xuezhaojun: also answered in https://github.com/open-cluster-management-io/ocm/pull/565#discussion_r1675316792

test/framework/klusterlet.go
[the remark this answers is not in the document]
qiujian16: what does KlusterletOperator mean?

test/framework/klusterlet.go
qiujian16: this could be confusing, because to me a spoke == a klusterlet

qiujian16: and for hub, hub.Cleanup(cluster1) should be fine.

test/framework/managedcluster.go
qiujian16: To my undstanding, hub defines a hub environment. A test case could manage hub environment with the hub struct.

test/framework/klusterlet.go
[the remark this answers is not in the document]
xuezhaojun: It's a copy of:

https://github.com/open-cluster-management-io/ocm/blob/7a8e444a52a360a6425bb9a2b01746e0e0c2f7c9/test/e2e/common.go#L73

I think it represent the deployment of of klusterlet operator:

https://github.com/open-cluster-management-io/ocm/blob/7a8e444a52a360a6425bb9a2b01746e0e0c2f7c9/deploy/klusterlet/config/operator/operator.yaml#L6

test/framework/klusterlet.go
xuezhaojun: During the e2e we are deploying multiple klusterlet on the same spoke: 

https://github.com/open-cluster-management-io/ocm/blob/7a8e444a52a360a6425bb9a2b01746e0e0c2f7c9/test/e2e/klusterlet_test.go#L38

In our e2e, a spoke should be a cluster that has klusterlet operator on it, and we deploy multiple klusterlet on one spoke.

xuezhaojun: Should we use `CleanUpManagedCluster`? Because if we use `CleanUp` specifically represent to clean up a managedcluster, then clean up manifestwork, clustermanager and other resources will be confusing.

test/framework/managedcluster.go
xuezhaojun: Sure, I can change to `hub.GetManagedCluster` since currently we have no plan for a ultimate test-framework for al l ocm related repos. But I still think `GetManagedCluster` is not limit to a hub environment, in integration cases, there is not hub just controllers and fake kube-apiserver, we can still reuse `GetManagedCluster`.

xuezhaojun: Updated, all functions are moved under hub and spoke as method.

test/framework/klusterlet.go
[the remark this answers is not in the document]
qiujian16: I do not think this needs to be a variable.

test/framework/klusterlet.go
qiujian16: I think you will need:
1: just clean klusterlet CR,
2. clean klusterlet CR and klusterlet operator
3. just clean a managedcluster CR
4. clean them all

qiujian16 (commented):
/approve

looks good in general, just some minor comments. And I would like everyone to take a look on the change. @open-cluster-management-io/core-dev

test/framework/klusterlet.go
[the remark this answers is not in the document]
xuezhaojun: That's true, maybe we can change these 2 to const values?

test/framework/klusterlet.go
xuezhaojun: Yes, can I list it as a TODO in the future, because "clean the klusterlet operator" part feels need more discussion, it's a behavior change comparing with previous code.

xuezhaojun:
> /approve
> 
> looks good in general, just some minor comments. And I would like everyone to take a look on the change. @open-cluster-management-io/core-dev

Hi, @qiujian16 I have synced the changes with @open-cluster-management-io/core-dev in a internal weekly meeting, is it ok now to merge the PR?

PR #577, opened 2024-07-30
:bug:  Fix: miss handle the err lease-controller.

xuezhaojun:
## Summary

When the lease controller get hub lease failed, the error is not handled in any way, change to reporting it via condition msg.

### DIFF
--- a/pkg/registration/spoke/addon/lease_controller.go
+++ b/pkg/registration/spoke/addon/lease_controller.go
@@ -160,7 +160,7 @@ func (c *managedClusterAddOnLeaseController) syncSingle(ctx context.Context,
 			Type:    addonv1alpha1.ManagedClusterAddOnConditionAvailable,
 			Status:  metav1.ConditionUnknown,
 			Reason:  "ManagedClusterAddOnLeaseNotFound",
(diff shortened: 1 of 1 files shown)


pkg/registration/spoke/addon/lease_controller.go
```diff
@@ -160,7 +160,7 @@ func (c *managedClusterAddOnLeaseController) syncSingle(ctx context.Context,
 			Type:    addonv1alpha1.ManagedClusterAddOnConditionAvailable,
 			Status:  metav1.ConditionUnknown,
 			Reason:  "ManagedClusterAddOnLeaseNotFound",
-			Message: fmt.Sprintf("The status of %s add-on is unknown.", addOn.Name),
+			Message: fmt.Sprintf("The status of %s add-on is unknown because get lease from hub failed: %v", addOn.Name, err),
```
qiujian16, line 163: isn't the err in this case is always not found?

xuezhaojun: Could it be a network eror from agent to access hub?

xuezhaojun: That make sense to me now, close the PR.

PR #578, opened 2024-07-30
🌱 Add chart action into release

qiujian16:
## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,6 +62,15 @@ jobs:
         run: |
           python hack/changelog.py ${{ secrets.GITHUB_TOKEN }} ${{ needs.env.outputs.RELEASE_VERSION }} > /home/runner/work/changelog.txt
           cat /home/runner/work/changelog.txt
+      - name: setup helm
+        uses: azure/setup-helm@v1
+      - name: chart package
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -62,6 +62,15 @@ jobs:
         run: |
           python hack/changelog.py ${{ secrets.GITHUB_TOKEN }} ${{ needs.env.outputs.RELEASE_VERSION }} > /home/runner/work/changelog.txt
           cat /home/runner/work/changelog.txt
+      - name: setup helm
+        uses: azure/setup-helm@v1
+      - name: chart package
+        run: |
+          mkdir -p release
+          pushd release
+          helm package ../deploy/cluster-manager/chart/cluster-manager
```
zhiweiyin318, line 71: should we change the appVersion to the release version  https://github.com/open-cluster-management-io/ocm/blob/5623a4b2a23048186590a9ad386c937d84a40a38/deploy/cluster-manager/chart/cluster-manager/Chart.yaml#L6 ? 
if not the helm chart will use the latest images by default.

zhujian7:
/lgtm
/hold
@zhiweiyin318 PTAL

PR #580, opened 2024-07-31
🌱 Move hub side csr code to csr driver

qiujian16:
## Related issue(s)

Fixes #


pkg/registration/register/interface.go
```diff
@@ -74,26 +73,14 @@ type RegisterDriver interface {
...
-		}
-
-		return driver.IsHubKubeConfigValid(ctx, secretOption)
-	}
+// Approvers is the inteface that each driver should implement on hub side. The hub controller will use this driver
+// to check the registration request from the agent and cleanup.
+type Approver interface {
+	// Run starts a reconciler on the hub side to monitor the registration request and approve the request
+	// if necessary.
+	Run(ctx context.Context, workers int)
```
zhujian7, line 81: should we mention that the `Run` method will block and not return immediately?
For me looking at the interface it is easy to think of Run and Cleanup are similar as one-time operations.

qiujian16: updated

zhujian7:
@qiujian16 the integration test failed, please take a look.

PR #593, opened 2024-08-13
:seedling: return raw data of chart render results

zhiweiyin318:
## Summary
update operator charts render function to return raw data of the results.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/helpers/chart/render.go
+++ b/pkg/operator/helpers/chart/render.go
@@ -1,10 +1,8 @@
-package chartrender
+package chart
 
 import (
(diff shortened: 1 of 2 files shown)


xuezhaojun:
Can we add the motivation of this PR? Like improve performance?

zhiweiyin318:
> Can we add the motivation of this PR? Like improve performance?

@xuezhaojun  to return raw data will be easy for other components to consume, like clusteradm .

qiujian16 (commented):
/approve
/lgtm

PR #594, opened 2024-08-13
🐛 Ignore appliedmanifestwork crd not found error when checking managed cluster connectivity

zhujian7:
## Summary
This PR will fix the bug that detaching cluster gets stuck when appliedManifestWork CRD does not exist on the hosted managed cluster
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go
+++ b/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go
@@ -220,6 +220,10 @@ func (n *klusterletCleanupController) checkConnectivity(ctx context.Context,
 	if err == nil {
 		return true, nil
 	}
+	if errors.IsNotFound(err) {
+		klog.Infof("AppliedManifestWork not found, klusterlet %s", klusterlet.Name)
+		return true, nil
+	}
 
 	// if the managed cluster is destroyed, the returned err is TCP timeout or TCP no such host,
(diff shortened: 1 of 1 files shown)


pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go
```diff
@@ -220,6 +220,10 @@ func (n *klusterletCleanupController) checkConnectivity(ctx context.Context,
 	if err == nil {
 		return true, nil
 	}
+	if errors.IsNotFound(err) {
+		klog.Infof("AppliedManifestWork not found, klusterlet %s, ignore the resources", klusterlet.Name)
+		return false, nil
```
elgnay, line 225: Return `true` or `false` in this situation?

zhujian7: returning true means we need to [clean up resources](https://github.com/open-cluster-management-io/ocm/blob/a08afd659bedae688f2eecdc2c5a0585591e4c8c/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go#L173) on the managed cluster(appliedmanifestworks, namespaces,crds), here if the appliedmanifestwork CRD is not found, I think we do not need to do the cleanup on the managed cluster.

zhujian7: discussed with Le offline, As the func checkConnectivity's name suggests, not found means the connectivity is OK, so I changed to returning true here.

PR #599, opened 2024-08-16
:seedling: Rename the solution "Multiple Hubs" to "Multiple Agents".

xuezhaojun:
## Summary

The purpose is to diff this solution with the [MultipleHubs](https://github.com/open-cluster-management-io/api/pull/331) feature we're working on.


qiujian16 (commented):
/approve
/lgtm

PR #603, opened 2024-08-27
:seedling: add createNamespace option to chart render

zhiweiyin318:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/helpers/chart/render.go
+++ b/pkg/operator/helpers/chart/render.go
@@ -11,10 +11,12 @@ import (
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/engine"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/json"
-	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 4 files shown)


pkg/operator/helpers/chart/render.go
```diff
@@ -172,3 +201,17 @@ func renderManifests(chart *chart.Chart, values chartutil.Values) ([][]byte, err
...
+
+func newNamespaceRawObject(namespace string) ([]byte, error) {
+	ns := &corev1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespace,
+		},
```
dongbeiqing91, line 213: In some enterprise environments, annotations or labels are sometimes required in Namespace. I would be better if optional annotations and labels can be supported here.

zhiweiyin318: The release ns can be created automatically without any customized annotations or labels if use Helm to install the operators . 
This PR only adds an option to render the operator helm charts to determine if add the ns objects into rendered object slice or not, like using `helm install --create-namepace `flag. 
By default the release ns will not be included into the result, and user needs to create it manually, and in the default case user can add any annotations or labels to the user's ns.

There are 2 kinds of ns in OCM, the ns in which the operators are running is created or named by user, the ns like `open-cluster-management-hub`, `open-cluster-management-agent` and `open-cluster-management-agent-addon` are created by cluster-manager or klusterlet operator.
I think we could consider to support add optional annotations and labels to these namespaces created by OCM operators.

dongbeiqing91: I see
/lgtm

PR #605, opened 2024-08-29
:seedling: support set imagepullsecret credential in helm chart and run e2e using helm chart

zhiweiyin318:
## Summary
1. support set imagepullsecret credential in the helm chart.
2. set the token-secret length to 16 in the chart.
3. use helm chart to deploy in e2e.
4. add a flag to control if create operator CR, because do not need create klusterlet CR in e2e deploy stage.  the default is to create. 
5. move the chart config struct to the helper dir. the consumer will only import the helper dir.

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ update: copy-crd update-csv
 
 test-unit: ensure-kubebuilder-tools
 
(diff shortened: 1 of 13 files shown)


qiujian16 (commented):
/approve
/lgtm

PR #611, opened 2024-09-05
:bug: remove the creationTimestamp from metadata when using SSA in work

skeeey:
## Summary
Currently, if the required object has zero creationTime in metadata, it will cause kube-apiserver to increment generation even if nothing else changes when using SSA to apply work's manifests

## Related issue(s)
https://github.com/kubernetes/kubernetes/issues/67610

### DIFF
--- a/pkg/work/spoke/apply/server_side_apply.go
+++ b/pkg/work/spoke/apply/server_side_apply.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
 
 	workapiv1 "open-cluster-management.io/api/work/v1"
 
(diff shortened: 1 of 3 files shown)


pkg/work/spoke/apply/server_side_apply.go
```diff
@@ -72,3 +80,27 @@ func (c *ServerSideApply) Apply(
 
 	return obj, err
 }
+
+func removeCreationTimeFromMetadata(obj map[string]interface{}) {
+	if metadata, found := obj["metadata"]; found {
+		if metaObj, ok := metadata.(map[string]interface{}); ok {
+			klog.V(4).Infof("remove `metadata.creationTimestamp`")
```
qiujian16, line 87: should we also check the value of creationTimestamp is nil

qiujian16 (commented):
I think we should also add an integration test

pkg/work/spoke/apply/server_side_apply.go
skeeey: we should

machi1990: +1 to @qiujian16 comment.

PR #630, opened 2024-09-26
:bug: monitor the bootstrap kubeconfig and restart immediately when changes

haoqing0110:
## Related issue(s)

Fixes #


qiujian16:
could you explain why we need to do this, I was always trying to avoid monitoring bootstrap secret

pkg/registration/spoke/event_handler.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	oldSecret, ok := oldObj.(*corev1.Secret)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("invalid secret object: %v", oldObj))
+		return
+	}
+
+	if !reflect.DeepEqual(newSecret.Data, oldSecret.Data) {
+		// Restart immediately if the bootstrap kubeconfig changes. Otherwise, in the backup restore scenario,
```
elgnay, line 46: backup restore scenario -> switch hub scenario
the restored one -> the new one

pkg/registration/spoke/options.go
```diff
@@ -22,9 +23,10 @@ type SpokeAgentOptions struct {
 	// 2. BootstrapKubeconfigSecret is the secret, a event handler will watch it, if the secret is changed, then rebootstrap.
 	// 3. BootstrapKubeconfigs is a list of file path, the controller uses one of its item to build the client.
 	// BootstrapKubeconfigs can only be used when MultipleHubs is enabled.
-	BootstrapKubeconfig       string
-	BootstrapKubeconfigSecret string
-	BootstrapKubeconfigs      []string
+	BootstrapKubeconfig             string
+	BootstrapKubeconfigSecret       string
+	BootstrapKubeconfigs            []string
+	BootstrapKubeconfigEventHandler *bootstrapKubeconfigEventHandler
```
elgnay, line 29: The SpokeAgentConfig might be a more suitable location for the BootstrapKubeconfigEventHandler, as it seems that SpokeAgentOptions primarily contains options and configurations

qiujian16:
I do not think it full resolve this issue. It could reduce the chance it happens, but cannot fully resolve it. The root cause for this is we use manifestwork to sync bootstrap-hubkubeconfig secret which generates a weird loop... I think one option we could resolve this is in the manifetwork, we could specify an option that "when some thing changes on the spoke, do not update while when it changes in mw, update it."

pkg/registration/spoke/event_handler.go
```diff
@@ -0,0 +1,68 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+type bootstrapKubeconfigEventHandler struct {
+	bootstrapKubeconfigSecretName *string
+	cancel                        context.CancelFunc
```
xuezhaojun, line 16: This cancel is specifically point to the context in:
```
func NewRegistrationAgent() *cobra.Command {
	ctx, cancel := context.WithCancel(context.TODO())
```

It feels it should have a more specific name or some commemnts to help readers understand the case, like:

```go
shutdownAgent context.CancelFunc
```

or 

```go
// This cancel function point to the root ctx of the agent, call it will shutdown the agent pod.
cancel context.Context
```

pkg/registration/spoke/spokeagent.go
```diff
@@ -54,9 +54,8 @@ type SpokeAgentConfig struct {
 
 	internalHubConfigValidFunc wait.ConditionWithContextFunc
 
-	hubKubeConfigChecker             *hubKubeConfigHealthChecker
-	bootstrapKubeconfigHealthChecker *bootstrapKubeconfigHealthChecker
-	reSelectChecker                  *reSelectChecker
+	hubKubeConfigChecker *hubKubeConfigHealthChecker
+	reSelectChecker      *reSelectChecker
```
xuezhaojun, line 58: @xuezhaojun

Just thought in the `multiplehubs` case, we also need an immediate restart as well. For example, when hub1 is down for 3 min (the threshold we set), we expect a restart and reselection to happen at that time point, not a few minutes later based on the health checker.

Maybe we should refactor these checkers all into using this `cancel` approach(but not in this PR, a follow-up PR in the future)? @qiujian16 any thoughts on this?

pkg/registration/spoke/event_handler.go
haoqing0110: Comments updated.

pkg/registration/spoke/options.go
haoqing0110: Moved to `SpokeAgentConfig`.

pkg/registration/spoke/event_handler.go
haoqing0110: Comments updated.

pkg/registration/spoke/spokeagent.go
qiujian16: yes, we could create an issue for followup

qiujian16 (commented):
/approve

pkg/registration/spoke/event_handler.go
```diff
@@ -0,0 +1,70 @@
...
+type bootstrapKubeconfigEventHandler struct {
+	bootstrapKubeconfigSecretName *string
+	// The cancel function allows the agent to restart immediately if the bootstrap kubeconfig changes.
+	// It should be point to the root context of the agent.
+	cancel context.CancelFunc
+}
+
+func (hc *bootstrapKubeconfigEventHandler) Name() string {
+	return "bootstrap-hub-kubeconfig"
+}
```
elgnay, line 23: The name function is no longer needed. It's part of the HealthChecker interface.

haoqing0110: removed.

PR #635, opened 2024-09-30
:sparkles: set configured condition in mca

haoqing0110:
## Related issue(s)

Fixes #


pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler.go
```diff
@@ -19,16 +22,55 @@ type managedClusterAddonConfigurationReconciler struct {
...
+		}
+	}
+
+	// Update mca configured condition to true for addons with rollout status Succeeded.
+	// Including:
+	// 1. Addons without configurations, the rollout status is set to success in func setRolloutStatus(),
+	// should set condition to true.
+	// 2. Addons with configurations and already rollout successfully. In upgrade scenario, when the
+	// addon configurations do not change while addon components upgrade, should set condition to true.
+	for _, addon := range graph.getAddonsSucceeded() {
```
qiujian16, line 64: this is not clear to me, should toUpdate and toApply include all addons need to be set? ToSuccess always comes from toUpdate.

pkg/addon/templateagent/template_agent.go
```diff
@@ -96,6 +97,11 @@ func (a *CRDTemplateAgentAddon) Manifests(
 	cluster *clusterv1.ManagedCluster,
 	addon *addonapiv1alpha1.ManagedClusterAddOn) ([]runtime.Object, error) {
 
+	if !meta.IsStatusConditionTrue(addon.Status.Conditions, addonapiv1alpha1.ManagedClusterAddOnConditionConfigured) {
```
qiujian16, line 100: if we have addon-framework check the conds, do we still need this?

test/integration/addon/addon_manager_upgrade_test.go
```diff
@@ -793,6 +841,12 @@ var _ = ginkgo.Describe("Addon upgrade", func() {
 					Reason:  addonapiv1alpha1.ProgressingReasonCompleted,
 					Message: "completed with no errors.",
 				})
+				assertManagedClusterAddOnConditions(testAddOnConfigsImpl.name, clusterNames[i], metav1.Condition{
+					Type:    addonapiv1alpha1.ManagedClusterAddOnConditionConfigured,
+					Status:  metav1.ConditionTrue,
+					Reason:  "ConfigurationsConfigured",
+					Message: "Configurations configured",
+				})
```
qiujian16, line 849: we need an integration test for the case that only when condition is set, the addon-template will start render.

pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler.go
haoqing0110: For addons that don't define any configs, the rollout status will be set to success in[setRolloutStatus](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/addon/controllers/addonconfiguration/graph.go#L99) directly, and will never appear in the toUpdate or toApply.
Another case is for example, addon defines config A, and in next release, both addon-framework and addon-manager upgrades while it keeps using config A, since the desired config already matches the actual, will not have change to add configured condition in toUpdate or toApply.

pkg/addon/templateagent/template_agent.go
haoqing0110: Currently, the code to "check the conds" is added at the beginning of each implementation of `Manifests()`, so both template and addon-framework have this piece of code.

test/integration/addon/addon_manager_upgrade_test.go
haoqing0110: will do.

pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler.go
```diff
@@ -19,16 +22,55 @@ type managedClusterAddonConfigurationReconciler struct {
...
+
+		configured.Insert(addon.mca.Namespace)
+	}
+
+	// Update mca configured condition to false for addons with rollout status ToApply.
+	for _, addon := range graph.getAddonsToApply() {
+		if configured.Has(addon.mca.Namespace) {
+			continue
+		}
+		if !meta.IsStatusConditionFalse(addon.mca.Status.Conditions, addonv1alpha1.ManagedClusterAddOnConditionConfigured) {
```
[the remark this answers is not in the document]
haoqing0110: fixed.

pkg/addon/templateagent/template_agent.go
haoqing0110: will try to find a better place to add the check.

haoqing0110:
Have updated the code based on comments and will create another PR to update the framework and add testing case for addon template.

qiujian16 (commented):
/approve
/lgtm

PR #636, opened 2024-09-30
🐛 Fix manifestwork and appliedmanifestwork unsync issue

qiujian16:
merge the two controller as reconcilers for one controller

## Related issue(s)

Fixes #


pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -137,96 +136,45 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
...
 
 	if len(errs) > 0 {
-		err = utilerrors.NewAggregate(errs)
-		klog.Errorf("Reconcile work %s fails with err: %v", manifestWorkName, err)
+		klog.Errorf("Reconcile work %s fails with err: %v", manifestWorkName, errs)
+		return utilerrors.NewAggregate(errs)
 	}
 
-	return err
+	if !mwUpdated && !amwUpdated && requeueTime < MaxRequeueDuration {
```
elgnay, line 173: In a case where a ManifestWork includes more than 1 manifest and one of them is failed to apply due to lack of permission, `wmUpdated` and `amwUpdated` will be `true` and `requeueTime` will be less than `MaxRequeueDuration`, but the requeue is still needed.

qiujian16: I think so

PR #637, opened 2024-09-30
:seedling: upgrade api and fix integration test issues

haoqing0110:
## Related issue(s)

Fixes #


test/integration/operator/klusterlet_test.go
```diff
@@ -77,6 +77,11 @@ var _ = ginkgo.Describe("Klusterlet", func() {
 				},
 				ClusterName: "testcluster",
 				Namespace:   klusterletNamespace,
+				RegistrationConfiguration: &operatorapiv1.RegistrationConfiguration{
+					RegistrationDriver: operatorapiv1.RegistrationDriver{
```
qiujian16, line 81: why do we need to set this? it is the default I think?

haoqing0110: it has a default value when creating yaml file but reports the below error when using the client:
```
              Message: "Klusterlet.operator.open-cluster-management.io \"klusterlet-multiplehubs-enabled\" is invalid: spec.registrationConfiguration.registrationDriver.authType: Unsupported value: \"\": supported values: \"csr\", \"awsirsa\"",
```

qiujian16: hrm....this will have compatible issue when other consumer is using this API. Should we make the RegistrationDriver as pointer?

qiujian16: or we need to ensure when other consumer update the api deps, they also need to specify this in their code.

haoqing0110: The field [`BootstrapKubeConfigs`](https://github.com/open-cluster-management-io/api/blob/main/operator/v1/types_klusterlet.go#L174) is also not a pointer, but seems the client never report error. Let me further check.

haoqing0110:
close this PR and https://github.com/open-cluster-management-io/api/pull/343 should fix the issues.

Issue #651, opened 2024-10-16
flaky test in addon template


qiujian16:
**Describe the bug**
this test fails sometimes
https://github.com/open-cluster-management-io/ocm/actions/runs/11360842169/job/31599378353?pr=650

qiujian16:
seems this is fixed already.

/close

PR #654, opened 2024-10-17
🐛 Ensure one cma is processed by one addon worker at the same time

zhujian7:
## Summary
Currently, there are [2 workers started for the addonConfigurationController](https://github.com/open-cluster-management-io/ocm/blob/33b409dae87e4b3854726e3d3cb5b2d7af95c166/pkg/addon/manager.go#L201), and this controller watches the ManagedClusterAddOn resource, when mca changes, it enqueues the managedClusterAddon [namespace and name](https://github.com/open-cluster-management-io/ocm/blob/33b409dae87e4b3854726e3d3cb5b2d7af95c166/pkg/addon/controllers/addonconfiguration/controller.go#L93) in the format "{namespace}/{name}", but in the sync func, it only [uses the name to get clusterManagementAddOn](https://github.com/open-cluster-management-io/ocm/blob/33b409dae87e4b3854726e3d3cb5b2d7af95c166/pkg/addon/controllers/addonconfiguration/controller.go#L105-L112).

Think about the case, there is an addon named `addon1`, and we created 2 managedClusterAddon, one is in the `cluster1` namespace, and the other one is in the `cluster2` namespace, so there will be 2 keys enqueued, `cluster1/addon1` and `cluster2/addon1`, because these 2 keys are not the same, the k8s cache queue will not merge them to one, so 2 workers will process these 2 keys at the same time. But finally, these 2 workers will get the same clusterManagementAddon `addon1` simultaneously. This may cause some race condition problems.

This PR changes the enqueue rule for the managedClusterAddon resources only to enqueue the name. so `cluster1/addon1` and `cluster2/addon1` changing will result in only enqueue one key `addon1`, only one worker will be assigned to process it.

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/651

Fixes #

### DIFF
--- a/pkg/addon/controllers/addonconfiguration/controller.go
+++ b/pkg/addon/controllers/addonconfiguration/controller.go
@@ -90,7 +90,7 @@ func NewAddonConfigurationController(
 		queue.QueueKeyByMetaNamespaceName,
 		c.addonFilterFunc,
 		clusterManagementAddonInformers.Informer()).
-		WithInformersQueueKeysFunc(queue.QueueKeyByMetaNamespaceName, addonInformers.Informer()).
+		WithInformersQueueKeysFunc(queue.QueueKeyByMetaName, addonInformers.Informer()).
 		WithInformersQueueKeysFunc(
 			index.ClusterManagementAddonByPlacementDecisionQueueKey(clusterManagementAddonInformers), placementDecisionInformer.Informer()).
 		WithInformersQueueKeysFunc(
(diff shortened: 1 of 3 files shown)


pkg/addon/controllers/addonconfiguration/controller.go
```diff
@@ -90,7 +90,7 @@ func NewAddonConfigurationController(
 		queue.QueueKeyByMetaNamespaceName,
 		c.addonFilterFunc,
 		clusterManagementAddonInformers.Informer()).
-		WithInformersQueueKeysFunc(queue.QueueKeyByMetaNamespaceName, addonInformers.Informer()).
+		WithInformersQueueKeysFunc(queue.QueueKeyByMetaName, addonInformers.Informer()).
```
zhujian7, line 93: this is the key point of this PR.

PR #667, opened 2024-10-22
🐛  WIP Fix: should `AcceptFalse` when get cluster forbidden.

xuezhaojun:
## Summary

After set `hubAcceptsClient` to `false`, there are chances the rbac resources removed before the event is watched by the agent. If that happens, the change will never captured by the agent side.

So change to use client not lister, and check if error is forbidden, if it's forbidden, treat it as 'accept=false`.

### DIFF
--- a/pkg/registration/spoke/registration/hub_accept_controller.go
+++ b/pkg/registration/spoke/registration/hub_accept_controller.go
@@ -2,44 +2,53 @@ package registration
 
 import (
 	"context"
+	"time"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
 	clusterv1informer "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
-	clusterv1listers "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
 )
 
 // hubAcceptController watch ManagedCluster CR on hub after spoke bootstrap done
 // If the managedCluster.Spec.HubAccetpsClient is false, then mark the connection as failed.
 //
 // Note that: when the HubAcceptsClient is false, the clusterrole and clusterrolebinding will be removed.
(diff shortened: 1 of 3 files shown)


xuezhaojun:
/assign @qiujian16 
cc @clyang82

pkg/registration/spoke/registration/hub_accept_controller.go
```diff
@@ -2,44 +2,53 @@ package registration
...
 	c := &hubAcceptController{
 		clusterName:       clusterName,
-		hubClusterLister:  hubClusterInformer.Lister(),
+		hubClusterClient:  hubClusterClient,
 		handleAcceptFalse: handleAcceptFalse,
 		recorder:          recorder,
 	}
 	return factory.New().
 		WithInformers(hubClusterInformer.Informer()).
+		ResyncEvery(3*time.Minute).
```
xuezhaojun, line 40: @qiujian16 I wonder if 3 min is too long in this case? Should we use 1 min?

qiujian16:
I am not sure we should treat it as a problem, Event with using client, it is still possible that agent is unauthorized before notifying accept=false. Why don't we just treat unauthorized and disconnection on the agent side.

pkg/registration/spoke/registration/hub_accept_controller.go
```diff
@@ -2,44 +2,53 @@ package registration
...
 		WithSync(c.sync).
 		ToController("HubAcceptController", recorder)
 }
 
 func (c *hubAcceptController) sync(ctx context.Context, _ factory.SyncContext) error {
-	cluster, err := c.hubClusterLister.Get(c.clusterName)
+	cluster, err := c.hubClusterClient.ClusterV1().ManagedClusters().Get(ctx, c.clusterName, metav1.GetOptions{})
 	if err != nil {
+		if apierrors.IsForbidden(err) {
```
xuezhaojun, line 48: @qiujian16  Here, when rbac on hub side is removed, client will return forbidden error and captured by this controller then `handle` as accpet False.

xuezhaojun:
> I am not sure we should treat it as a problem, Event with using client, it is still possible that agent is unauthorized before notifying accept=false. Why don't we just treat unauthorized and disconnection on the agent side.

I comment on the line of code that can capture the error when `rbac` resources are removed.

xuezhaojun:
Move to antoher way of fix: https://github.com/open-cluster-management-io/ocm/pull/675

PR #671, opened 2024-10-25
🐛 Fix duplicate deployment names in klusterlet-agent availability message

RokibulHasan7:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
+++ b/pkg/operator/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	appsinformer "k8s.io/client-go/informers/apps/v1"
 	"k8s.io/client-go/kubernetes"
 	appslister "k8s.io/client-go/listers/apps/v1"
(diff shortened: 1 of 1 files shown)


pkg/operator/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
```diff
@@ -78,24 +78,27 @@ func (k *klusterletStatusController) sync(ctx context.Context, controllerContext
 	registrationDeploymentName := fmt.Sprintf("%s-registration-agent", klusterlet.Name)
 	workDeploymentName := fmt.Sprintf("%s-work-agent", klusterlet.Name)
 
-	if helpers.IsSingleton(klusterlet.Spec.DeployOption.Mode) {
+	IsSingleton := helpers.IsSingleton(klusterlet.Spec.DeployOption.Mode)
```
zhujian7, line 81:
```suggestion
	isSingleton := helpers.IsSingleton(klusterlet.Spec.DeployOption.Mode)
```

zhujian7:
Another option is to deduplicate the agents in the `checkAgentsDeploymentAvailable` func, like:
```
// Check agent deployments, if both of them have at least 1 available replicas, return available condition
func checkAgentsDeploymentAvailable(ctx context.Context, kubeClient kubernetes.Interface, agents []klusterletAgent) metav1.Condition {
	var availableMessages []string
	components := sets.Set[string]{}  ### import "k8s.io/apimachinery/pkg/util/sets"
	for _, agent := range agents {
		componentID := fmt.Sprintf("%s-%s", agent.namespace, agent.deploymentName)
		if components.Has(componentID) {
			continue
		}
		components.Insert(componentID)
		deployment, err := kubeClient.AppsV1().Deployments(agent.namespace).Get(ctx, agent.deploymentName, metav1.GetOptions{})
		...
}

```

pkg/operator/operators/klusterlet/controllers/statuscontroller/klusterlet_status_controller.go
RokibulHasan7: updated

xuezhaojun:
@RokibulHasan7 Can we have a comparison of how the message looks before and after the change. This will help followers and code readers understand the fix better. Thank you!

RokibulHasan7:
> @RokibulHasan7 Can we have a comparison of how the message looks before and after the change. This will help followers and code readers understand the fix better. Thank you!

Before:
```
  Type:			Available
  Status:		True
  LastTransitionTime:	2024-10-27 13:25:38 +0600 +06
  Reason:		KlusterletAvailable
  Message:		deployments are ready: klusterlet-agent,klusterlet-agent
  ```
  
After:
```
  Type:			Available
  Status:		True
  LastTransitionTime:	2024-10-27 13:25:38 +0600 +06
  Reason:		KlusterletAvailable
  Message:		deployments are ready: klusterlet-agent
```

qiujian16:
good catch, thanks!
/approve

PR #673, opened 2024-10-26
✨ Register ManifestWorkReplicaSet webhook when feature is enabled

RokibulHasan7:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/work/webhook/v1alpha1/manifestworkreplicaset_validating.go
+++ b/pkg/work/webhook/v1alpha1/manifestworkreplicaset_validating.go
@@ -6,13 +6,13 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
-	utilfeature "k8s.io/apiserver/pkg/util/feature"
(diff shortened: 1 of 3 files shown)


zhiweiyin318:
/lgtm
need approve from @qiujian16

pkg/work/webhook/v1alpha1/manifestworkreplicaset_validating.go
```diff
@@ -78,7 +78,7 @@ func validatePlaceManifests(mwrSet *workv1alpha1.ManifestWorkReplicaSet) error {
 }
 
 func checkFeatureEnabled() error {
-	if !utilfeature.DefaultMutableFeatureGate.Enabled(ocmfeature.ManifestWorkReplicaSet) {
+	if !features.HubMutableFeatureGate.Enabled(ocmfeature.ManifestWorkReplicaSet) {
```
zhujian7, line 81: nit: because we already checked if the feature is enabled https://github.com/open-cluster-management-io/ocm/pull/673/files#diff-d26f37502c2c85732485a395caad60404fad094e6eac48fe2df438b2f24871d1R78 , do we still need to check again here?

qiujian16 (commented):
/approve

PR #675, opened 2024-10-27
:bug: Fix accetp false add rbac

xuezhaojun:
## Summary

After set hubAcceptsClient to false, there are chances the rbac resources removed before the event is watched by the agent. If that happens, the change will never captured by the agent side.

So this change is to keep a `watch` role of managed cluster to let mangaed cluster watch events when it's not accepted.

### DIFF
--- a/pkg/registration/hub/managedcluster/controller.go
+++ b/pkg/registration/hub/managedcluster/controller.go
@@ -126,8 +126,33 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 		// Hub cluster-admin denies the current spoke cluster, we remove its related resources and update its condition.
 		c.eventRecorder.Eventf("ManagedClusterDenied", "managed cluster %s is denied by hub cluster admin", managedClusterName)
 
-		if err := c.removeManagedClusterResources(ctx, managedClusterName); err != nil {
-			return err
+		// Apply(Update) the cluster specific rbac resources for this spoke cluster with hubAcceptsClient=false.
+		var errs []error
+		applyResults := c.applier.Apply(ctx, syncCtx.Recorder(),
+			helpers.ManagedClusterAssetFnWithAccepted(manifests.RBACManifests, managedClusterName, managedCluster.Spec.HubAcceptsClient),
+			manifests.ClusterSpecificRBACFiles...)
+		for _, result := range applyResults {
+			if result.Error != nil {
+				errs = append(errs, result.Error)
+			}
+		}
+		if aggErr := operatorhelpers.NewMultiLineAggregate(errs); aggErr != nil {
+			return aggErr
+		}
+
+		// Remove the cluster role binding files for registration-agent and work-agent.
+		removeResults := resourceapply.DeleteAll(ctx,
+			resourceapply.NewKubeClientHolder(c.kubeClient),
+			c.eventRecorder,
+			helpers.ManagedClusterAssetFn(manifests.RBACManifests, managedClusterName),
+			manifests.ClusterSpecificRoleBindings...)
+		for _, result := range removeResults {
+			if result.Error != nil {
+				errs = append(errs, fmt.Errorf("%q (%T): %v", result.File, result.Type, result.Error))
+			}
+		}
+		if aggErr := operatorhelpers.NewMultiLineAggregate(errs); aggErr != nil {
+			return aggErr
 		}
 
 		if err = c.approver.Cleanup(ctx, managedCluster); err != nil {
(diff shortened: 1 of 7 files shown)


pkg/registration/hub/manifests/rbac/managedcluster-watch-clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: open-cluster-management:managedcluster:{{ .ManagedClusterName }}-watch
+  labels:
```
qiujian16, line 5: can we just update the existing clusterrole? So we do not recreate another two additional clusterroles

xuezhaojun: Yes, that's also an approach, the PR updated, some original concerns are:
1. the following four rbac files used to have the same lifecycle, change the `clusterrole.yaml` need to break this set into 2 sets, the different is clusterrole doe
```go
var ClusterSpecificRBACFiles = []string{
	"rbac/managedcluster-clusterrole.yaml",
	"rbac/managedcluster-clusterrolebinding.yaml",
	"rbac/managedcluster-registration-rolebinding.yaml",
	"rbac/managedcluster-work-rolebinding.yaml",
}
```

2. the `update` need to introduce another varible in process of `ManagedClusterAssetFn`, and then also affect the upper caller functions:

```go
func ManagedClusterAssetFn(fs embed.FS, managedClusterName string) resourceapply.AssetFunc {
	return func(name string) ([]byte, error) {
		config := struct {
			ManagedClusterName string
		}{
			ManagedClusterName: managedClusterName,
		}

		template, err := fs.ReadFile(name)
		if err != nil {
			return nil, err
		}
		return assets.MustCreateAssetFromTemplate(name, template, config).Data, nil
	}
}
```

xuezhaojun: PR updated, if the new changes are accepted, I will squash the commits and continue on.

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -126,8 +126,37 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 		// Hub cluster-admin denies the current spoke cluster, we remove its related resources and update its condition.
 		c.eventRecorder.Eventf("ManagedClusterDenied", "managed cluster %s is denied by hub cluster admin", managedClusterName)
 
-		if err := c.removeManagedClusterResources(ctx, managedClusterName); err != nil {
-			return err
+		// Apply(Update) the cluster specific rbac resources for this spoke cluster with hubAcceptsClient=false.
+		var errs []error
+		{
```
qiujian16, line 131: what this bracket for?

qiujian16 (commented):
add at least some unit test?

pkg/registration/hub/managedcluster/controller.go
xuezhaojun: U-T added,
adding brackets can make `resourceResults` only affect inside it. It seems confused,  changed,

test/integration/registration/managedcluster_accept_test.go
```diff
@@ -0,0 +1,202 @@
...
+			}
+
+			// Should have rules for certificate rotation
+			hasCertRule := false
+			hasClientCertRule := false
+			hasManagedClusterRule := false
+			hasManagedClusterStatusRule := false
+
+			for _, rule := range clusterRole.Rules {
+				// Check certificate signing request rule
```
qiujian16, line 50: this looks a bit complicated, and when some rule changes, the test will fail. Can we just read the file from manifests and compare with the output?

xuezhaojun: I have some concerns about this one. I think when a rule changes, the developer should modify the checks here. For example, if the future developer adds the update to the cluster role when accept=false, the integration test case should fail to let the developer know they added a permission that we don't want. 

If we configure a wrong manifest, `reading from manifests and comparing with output` may repeat the mistake twice.

qiujian16: If so can we send a SAR instead of comparing the rules?

xuezhaojun: If check every rule is too complicated, how about change to only check on rule when `accept=false`? To make sure when `accept=false`, no `write` permission on `read`

test/integration/registration/managedcluster_accept_test.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+
+			rule := clusterRole.Rules[0]
+			expectedRule := rbacv1.PolicyRule{
+				APIGroups:     []string{"cluster.open-cluster-management.io"},
+				Resources:     []string{"managedclusters"},
+				ResourceNames: []string{managedCluster.Name},
+				Verbs:         []string{"get", "list", "watch"},
+			}
+			if !reflect.DeepEqual(rule, expectedRule) {
```
xuezhaojun, line 105: Compare with an expected rule structure. @qiujian16

test/integration/registration/managedcluster_accept_test.go
```diff
@@ -0,0 +1,169 @@
...
+			if err != nil {
+				return err
+			}
+
+			// check clusterrole is correct
+			clusterRole, err := kubeClient.RbacV1().ClusterRoles().Get(context.Background(), mclClusterRoleName(managedCluster.Name), metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			if len(clusterRole.Rules) != 4 {
```
xuezhaojun, line 45: Less restrictive check, only check rule numbers, so that when we have a slight update, the case will still pass. @qiujian16

xuezhaojun, line 45: Less restrictive check, only check rule numbers, so that when we have a slight update, the case will still pass. @qiujian16

test/integration/registration/managedcluster_accept_test.go
xuezhaojun: I updated to use `DeepEqual`, as we are going to `get` clusterrole anyway,  compare between structure maybe a little simpler than SAR.

qiujian16 (commented):
/approve
/lgtm

PR #677, opened 2024-10-27
:sparkles: Add AWS IAM support

suvaanshkumar:
## Summary

The PR adds changes for  AWS IAM based authentication support for Open Cluster Management (OCM) for Amazon EKS clusters. This change updates the Klusterletconfig object by adding the registration driver for awsirsa when the value of registration-auth flag is set as awsirsa from clusteradm join command.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/helpers/helpers.go
+++ b/pkg/registration/helpers/helpers.go
@@ -160,6 +160,8 @@ func IsCSRSupported(nativeClient kubernetes.Interface) (bool, bool, error) {
 	return v1CSRSupported, v1beta1CSRSupported, nil
 }
 
+// IsEksArnWellFormed checks if the EKS cluster ARN is well-formed
+// Example of a well-formed ARN: arn:aws:eks:us-west-2:123456789012:cluster/my-cluster
 func IsEksArnWellFormed(eksArn string) bool {
 	pattern := "^arn:aws:eks:([a-zA-Z0-9-]+):(\\d{12}):cluster/([a-zA-Z0-9-]+)$"
 	matched, err := regexp.MatchString(pattern, eksArn)
--- a/pkg/registration/helpers/helpers.go
+++ b/pkg/registration/helpers/helpers.go
@@ -3,6 +3,7 @@ package helpers
 import (
 	"embed"
 	"net/url"
+	"regexp"
 
 	"github.com/openshift/library-go/pkg/assets"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
@@ -157,4 +158,13 @@ func IsCSRSupported(nativeClient kubernetes.Interface) (bool, bool, error) {
 		}
 	}
 	return v1CSRSupported, v1beta1CSRSupported, nil
+}
+
+func IsEksArnWellFormed(eksArn string) bool {
+	pattern := "^arn:aws:eks:([a-zA-Z0-9-]+):(\\d{12}):cluster/([a-zA-Z0-9-]+)$"
+	matched, err := regexp.MatchString(pattern, eksArn)
+	if err != nil {
+		return false
+	}
+	return matched
 }
--- a/pkg/registration/helpers/testing/testinghelpers.go
+++ b/pkg/registration/helpers/testing/testinghelpers.go
@@ -421,6 +421,7 @@ type TestCert struct {
 	Key  []byte
 }
 
+// TODO: Remove this struct once we have the function fully implemented for the AWSIRSADriver
 type TestIrsaRequest struct {
 }
 
--- a/pkg/registration/helpers/testing/testinghelpers.go
+++ b/pkg/registration/helpers/testing/testinghelpers.go
@@ -421,6 +421,9 @@ type TestCert struct {
 	Key  []byte
 }
 
+type TestIrsaRequest struct {
+}
+
 func NewHubKubeconfigSecret(namespace, name, resourceVersion string, cert *TestCert, data map[string][]byte) *corev1.Secret {
 	secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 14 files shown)


mikeshng (commented):
Thanks for the PR!

It's currently failing at this check. Could you please make sure it works locally `make update-gofmt`?

Also, could you please update the PR summary with more details?

mikeshng:
@dtclxy64 @suvaanshkumar could you please take a look at the merge conflict and rebase? 

I think you could try deleting the following, based on the PR https://github.com/open-cluster-management-io/ocm/pull/658
```

		reSelectChecker: &reSelectChecker{shouldReSelect: false},
		bootstrapKubeconfigEventHandler: &bootstrapKubeconfigEventHandler{
			bootstrapKubeconfigSecretName: &opts.BootstrapKubeconfigSecret,
			cancel:                        cancel,
		},
```

mikeshng:
@dtclxy64 thanks for the rebase. It's now failing at the DCO check and `make update-gofmt` check. 
For DCO you could try:
`git commit -s --amend` then force push.

pkg/registration/spoke/options.go
```diff
@@ -74,6 +76,10 @@ func (o *SpokeAgentOptions) AddFlags(fs *pflag.FlagSet) {
 			"the value of --cluster-signing-duration command-line flag of the kube-controller-manager will be used.")
 	fs.StringToStringVar(&o.ClusterAnnotations, "cluster-annotations", o.ClusterAnnotations, `the annotations with the reserve
 	 prefix "agent.open-cluster-management.io" set on ManagedCluster when creating only, other actors can update it afterwards.`)
+	fs.StringVar(&o.RegistrationAuth, "registration-auth", o.RegistrationAuth,
```
qiujian16, line 79: I think we should have a drivers Options and we just need to AddFlags here?

pkg/registration/register/aws_irsa/aws_irsa.go
```diff
@@ -0,0 +1,105 @@
...
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/cache"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/klog/v2"
+
+	"open-cluster-management.io/ocm/pkg/registration/register"
+)
+
+const (
+	// TLSKeyFile is the name of tls key file in kubeconfigSecret
```
qiujian16, line 20: nit you would not need these files. This are only for mtls

pkg/registration/helpers/testing/testinghelpers.go
```diff
@@ -421,6 +421,9 @@ type TestCert struct {
 	Key  []byte
 }
 
+type TestIrsaRequest struct {
+}
```
qiujian16, line 425: might add some TODO or comment here,  I am not quite sure why this is added.

pkg/registration/helpers/helpers.go
```diff
@@ -158,3 +159,12 @@ func IsCSRSupported(nativeClient kubernetes.Interface) (bool, bool, error) {
 	}
 	return v1CSRSupported, v1beta1CSRSupported, nil
 }
+
+func IsEksArnWellFormed(eksArn string) bool {
```
qiujian16, line 163: add some comment for this func, it will be easier to understand what is the format for the arn.

pkg/registration/spoke/spokeagent.go
```diff
@@ -54,7 +59,8 @@ type SpokeAgentConfig struct {
 
 	internalHubConfigValidFunc wait.ConditionWithContextFunc
 
-	hubKubeConfigChecker *hubKubeConfigHealthChecker
+	hubKubeConfigChecker            *hubKubeConfigHealthChecker
+	bootstrapKubeconfigEventHandler *bootstrapKubeconfigEventHandler
```
qiujian16, line 63: why this is added? I did not see where it is used

qiujian16 (commented):
looks good in general, just some minor comments

pkg/registration/spoke/options.go
dtclxy64: @qiujian16 We are currently using the `registration-auth` flag to determine the type of `registerDriver` to set up. See lines 199-202 in pkg/registration/spoke/spokeagent.go. 

Are you suggesting that we should add another flag to capture the type of the registration driver as well?

pkg/registration/spoke/spokeagent.go
dtclxy64: We got the change from the main branch which was introduced as part of PR #630, but we see in the latest main branch, it's no longer used. We removed it to match the main branch as requested.

pkg/registration/helpers/testing/testinghelpers.go
dtclxy64: We are using the struct in `aws_irsa_test.go` on lines 38 and 64, which is setting up a test irsa driver for each test case.

pkg/registration/spoke/options.go
qiujian16: the flag name is fine. I think you could put all these flags adding func to the register directory with a new Option struct. And the code here just to call driverOptions.AddFlags(fs)

pkg/registration/helpers/testing/testinghelpers.go
qiujian16: I see, I think we could add a //TODO here so we will not forget the followup task later :)

pkg/registration/register/aws_irsa/aws_irsa.go
qiujian16: Add a //TODO here so we do not forget to remove these later.

PR #678, opened 2024-10-28
🌱  Remove ManagedClusterSet webhook registration part

RokibulHasan7:
## Summary
We are not implementing any mutating or validating webhooks for `ManagedClusterSet` here, so we can remove this webhook registration section.


zhujian7:
@RokibulHasan7 could you add some description for this PR, for example, why do we need this?

RokibulHasan7:
> @RokibulHasan7 could you add some description for this PR, for example, why do we need this?

done. https://github.com/open-cluster-management-io/ocm/pull/678#issue-2617308890

PR #680, opened 2024-10-31
:bug: Add auto approve annotation for an accepted cluster when auto approval enabled

skeeey:
## Summary

This will fix the below problem

1. The hub has an accepted cluster.
2. Enable the auto approval feature
3. Set the cluster hubAcceptsClient to false

The cluster hubAcceptsClient should be able to set to false


qiujian16 (commented):
/approve
/lgtm

PR #682, opened 2024-10-31
:seedling: Check `error` instead of `bool` to see error msg when case fails.

xuezhaojun:
## Description:
This pull request updates the test cases in `managedcluster_deletion_test.go` to check for errors directly instead of returning boolean values. This change enhances the clarity of error reporting when test cases fail, allowing for more informative messages regarding the specific issues encountered.

## Changes:
* Modified gomega.Eventually assertions to return error instead of bool.
* Updated error handling to provide detailed messages when conditions are not met, improving the debugging experience


qiujian16 (commented):
/approve
/lgtm

PR #686, opened 2024-11-05
✨ [WIP] Manifestwork update resource only when it changes in mw

qiujian16:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/work/spoke/apply/update_apply.go
+++ b/pkg/work/spoke/apply/update_apply.go
@@ -2,7 +2,9 @@ package apply
 
 import (
 	"context"
+	"crypto/md5"
 	"fmt"
+	"io"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/work/spoke/apply/update_apply.go
```diff
@@ -172,3 +201,46 @@ func isSameUnstructured(obj1, obj2 *unstructured.Unstructured) bool {
...
+	if err != nil {
+		return true
+	}
+	// existing could be nil in some cases.
+	if existing == nil {
+		return false
+	}
+	existingAccessor, err := meta.Accessor(existing)
+	if err != nil {
+		return true
```
zhujian7, line 232: if building accessor for the existing object failed, should we apply the object?

pkg/work/spoke/apply/update_apply.go
```diff
@@ -172,3 +201,46 @@ func isSameUnstructured(obj1, obj2 *unstructured.Unstructured) bool {
...
+	}
+	fmt.Printf("2 existing object is %v\n", required)
```
zhujian7, line 234:
```suggestion
```

pkg/work/spoke/apply/update_apply.go
```diff
@@ -172,3 +201,46 @@ func isSameUnstructured(obj1, obj2 *unstructured.Unstructured) bool {
...
+	if len(requiredAccessor.GetAnnotations()) == 0 {
+		return true
+	}
+	requiredObjHash := requiredAccessor.GetAnnotations()[objHashAnnotation]
+	fmt.Printf("existing object is %v\n", existing)
```
zhujian7, line 239:
```suggestion
```

PR #687, opened 2024-11-06
🌱 Update `api` dependency and remove unnecessary featureGate enabling part 

RokibulHasan7:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/operator/helpers/helpers.go
+++ b/pkg/operator/helpers/helpers.go
@@ -38,7 +38,6 @@ import (
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	apiregistrationclient "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
 
-	"open-cluster-management.io/api/feature"
 	operatorapiv1 "open-cluster-management.io/api/operator/v1"
 )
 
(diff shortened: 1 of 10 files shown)


pkg/operator/helpers/helpers.go
```diff
@@ -775,11 +767,23 @@ func ConvertToFeatureGateFlags(component string, features []operatorapiv1.Featur
 			continue
 		}
 
-		if feature.Mode == operatorapiv1.FeatureGateModeTypeDisable && defaultFeature.Default {
+		if feature.Mode == operatorapiv1.FeatureGateModeTypeEnable {
```
zhujian7, line 770: why do we need to change this?

pkg/operator/helpers/helpers_test.go
```diff
@@ -1572,15 +1573,15 @@ func TestConvertToFeatureGateFlags(t *testing.T) {
 		{
 			name:         "unset",
 			features:     []operatorapiv1.FeatureGate{},
-			desiredFlags: []string{},
+			desiredFlags: []string{"--feature-gates=ClusterClaim=true", "--feature-gates=AddonManagement=true"},
```
zhujian7, line 1576: if it is unset, should we keep the desireFlags unset?

RokibulHasan7: If the user's provided featureGate is nil, the default features will be enabled automatically. So, the desiredFlags set to the default featureGate

zhujian7: I am thinking of the case: if we set the default value in the flags explicitly(TestFeature=false), once we change the default value in the future(TestFeature=true), will the TestFeature be changed(false to true)? This is going to be hard to decide because we do not know whether the old value(TestFeature=false) is set by default or users. what is our expectation in this case?

RokibulHasan7: If the user explicitly sets a value for a feature gate, we’re always taking that as the defined value. If the user doesn’t set any value for feature gates where Default=True, then we’re enabling those feature gates.

zhujian7: If the user doesn’t set any value for feature gates(TestFeature) where Default=False, so this PR will set the flag TestFeature=False right? If we change the TestFeature Default=True in the next release, when upgrading from this release to the next release, how do we know if the old flag `TestFeature=False` is set by users on purpose or set by us by default? I am asking because I think if the user intentionally disabled a feature by setting TestFeature=False, we should not enable it when upgrading.

RokibulHasan7: Got it, updated. PTAL @zhujian7

qiujian16:
/approve

leaves to @zhujian7 for the final review.

PR #693, opened 2024-11-11
:sparkles: support cel selector

haoqing0110:
## Related issue(s)

https://github.com/open-cluster-management-io/ocm/issues/843 
https://github.com/open-cluster-management-io/ocm/issues/923


vendor/open-cluster-management.io/api/cluster/v1beta1/types_placement.go
```diff
@@ -168,19 +168,30 @@ type ClusterPredicate struct {
...
 	// +optional
 	ClaimSelector ClusterClaimSelector `json:"claimSelector,omitempty"`
 }
 
+type ClusterLabelSelector struct {
+	// +optional
+	metav1.LabelSelector `json:",inline"`
+
+	// +optional
+	RegexSelector map[string]string `json:"regexSelector,omitempty"`
```
qiujian16, line 183: I would consider a CEL based selector, which could be more flexible.
e.g. could reference a score value

PR #703, opened 2024-11-14
:sparkles: support wildcard in ManifestConfigs

zhiweiyin318:
## Summary
Support wildcard in the name and namespace of ManifestConfigs.ResourceIdentifier for feedbackRule and updateStrategy.
if the a resource matches more than 1 options, only the first one works.
## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/705
Fixes #

### DIFF
--- a/pkg/work/helper/helpers.go
+++ b/pkg/work/helper/helpers.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 	"time"
 
@@ -119,7 +120,7 @@ func resetOrdinal(meta workapiv1.ManifestResourceMeta) workapiv1.ManifestResourc
 func mergeManifestCondition(condition, newCondition workapiv1.ManifestCondition) workapiv1.ManifestCondition {
 	return workapiv1.ManifestCondition{
 		ResourceMeta: newCondition.ResourceMeta,
-		//Note this func is only used for merging status conditions, the statusFeedbacks should keep the old one.
+		// Note this func is only used for merging status conditions, the statusFeedbacks should keep the old one.
 		StatusFeedbacks: condition.StatusFeedbacks,
 		Conditions:      MergeStatusConditions(condition.Conditions, newCondition.Conditions),
 	}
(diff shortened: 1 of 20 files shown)


pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
```diff
@@ -195,7 +195,7 @@ func (m *manifestworkReconciler) applyOneManifest(
 	requiredOwner := manageOwnerRef(ownedByTheWork, owner)
 
 	// find update strategy option.
-	option := helper.FindManifestConiguration(resMeta, workSpec.ManifestConfigs)
+	option := helper.FindManifestConfiguration(resMeta, workSpec.ManifestConfigs)
```
qiujian16, line 198: I would want updateStrategy also support wildcard match. I think we should have helper.FindManifestConfigurations(resMeta, workSpec.ManifestConfigs) returns a list. For updateStrategy, we pick the first or latest item in the list, while for feedback rules, we pick every item.

test/integration/work/statusfeedback_test.go
```diff
@@ -242,8 +242,8 @@ var _ = ginkgo.Describe("ManifestWork Status Feedback", func() {
 					ResourceIdentifier: workapiv1.ResourceIdentifier{
 						Group:     "apps",
 						Resource:  "deployments",
-						Namespace: commOptions.SpokeClusterName,
-						Name:      "deploy1",
+						Namespace: "*",
+						Name:      "*1",
```
qiujian16, line 246: we should have some more tests for wildcard case.

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
zhiweiyin318: returned a list of options .

test/integration/work/statusfeedback_test.go
zhiweiyin318: added

qiujian16: We should also have another test with multiple matched options.

qiujian16:
/approve

@elgnay or @zhujian7 please also take a look

zhiweiyin318:
added more integration tests in feedBackRule and upgradeStrategy : 
1. a wildcard option matches multi resources 2. a resource is matched by multi options. 
@elgnay @zhujian7 please take a look. cc @qiujian16

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
zhujian7: > while for feedback rules, we pick every item.

@zhiweiyin318 want to confirm for the current implementation, do we only pick up the first one?

zhiweiyin318: fixed. for feedbackRule, will pick all matched options.

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
```diff
@@ -194,16 +194,25 @@ func (m *manifestworkReconciler) applyOneManifest(
...
+	// strategy is updated by default
 	strategy := workapiv1.UpdateStrategy{Type: workapiv1.UpdateStrategyTypeUpdate}
-	if option != nil && option.UpdateStrategy != nil {
-		strategy = *option.UpdateStrategy
+
+	// only handle the first option if the resource matches more than one option.
+	// need avoid a resource from being matched to more options.
+	option := workapiv1.ManifestConfigOption{}
+	if len(options) > 0 {
+		option = options[0]
```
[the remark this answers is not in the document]
zhiweiyin318: fixed. for feedbackRule, will pick all matched options.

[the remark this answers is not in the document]
qiujian16: I am not sure it works. Each feedback result have a unique name, and it should not be overlapped.

qiujian16 (commented):
/approve
/lgtm

test/integration/work/statusfeedback_test.go
zhiweiyin318: done

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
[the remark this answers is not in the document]
zhiweiyin318: changed to only pick up the first one for feedbackRule and installStrategy.

zhiweiyin318:
@zhujian7 @qiujian16  please take a look again.

pkg/work/helper/helpers.go
```diff
@@ -364,20 +365,18 @@ func NewAppliedManifestWorkOwner(appliedWork *workapiv1.AppliedManifestWork) *me
...
+func FindManifestConfiguration(resourceMeta workapiv1.ManifestResourceMeta,
+	manifestOptions []workapiv1.ManifestConfigOption) *workapiv1.ManifestConfigOption {
+	// return the first matched option
+	for i := 0; i < len(manifestOptions); i++ {
+		option := manifestOptions[i]
+		if resourceMeta.Group == option.ResourceIdentifier.Group &&
+			resourceMeta.Resource == option.ResourceIdentifier.Resource &&
+			wildcardMatch(resourceMeta.Namespace, option.ResourceIdentifier.Namespace) &&
+			wildcardMatch(resourceMeta.Name, option.ResourceIdentifier.Name) {
+			return &option
```
zhujian7, line 377: 1. If two options match, the first configured the FeedbackRules but no UpdateStrategy, and the second configured the UpdateStrategy but no FeedbackRules, will returning the first option cause the FeedbackRules to be lost? 
2. I am still considering if we should make the word-by-word matching take precedence over wildcard match?

zhiweiyin318: > If two options matched, the first configured the FeedbackRules but no UpdateStrategy, and the second configured the UpdateStrategy but no FeedbackRules, will returning the first option cause the FeedbackRules to be lost?

fixed by splitting the FindManifestConfiguration into 2 functions.

> I am still considering if we should make the word-by-word matching take precedence over wildcard matching?
I think the users only use the first option to filter resources in the most cases, so word-by-word matching is the same level as wildcard matching.
@qiujian16 WDYT?

qiujian16: could we just build a new option finding the first available rule and the first updateStrategy in the FindManifestConfiguration?

zhiweiyin318: done. changed to one FindManifestConfiguration and return an option with the first matched updateStrategy and feedbackRule.

qiujian16 (commented):
/lgtm

PR #708, opened 2024-11-19
:seedling: Revert "enable leader election of klusterlet-agent on single node man…

xuezhaojun:
…aged clusters (#695)"

This reverts commit 8544ff1e2999ad774193186945aaa58938f56278.

## Summary

In practice, we found the commit affect on some concern cases, revert the PR util we confirm the effect.


qiujian16:
could you explain what cases is impacted? so that we can add tests in the future.

xuezhaojun:
> could you explain what cases is impacted? so that we can add tests in the future.

We have some e2e cases that detach and cleanup klusterlet resources many times. From my test, when using the latest images, there are about 3-6 cases often fail.

Issue #690, opened 2024-11-11
Work supports ignore resources's fileds differents


ivan-cai:
**Describe the enhancement**
A clear and concise description of what the enhancement request is.
If a deployment is applied by manifestwork, and I use a HPA for this deployment, will have a conflict with replicas. Work agent will override the replicas that changed by HPA.
If I use HPA, I would to control replicas by HPA, not work.

ivan-cai:
@qiujian16

PR #726, opened 2024-11-28
✨ Implement ignoreFields in server side apply

qiujian16:
## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/690


zhujian7:
LGTM

pkg/work/spoke/apply/server_side_apply_test.go
```diff
@@ -279,3 +294,251 @@ func newManifestWork() *unstructured.Unstructured {
...
+				removeFieldByJSONPath(desired.UnstructuredContent(), jsonPath, logger)
+			}
+			if !equality.Semantic.DeepEqual(c.exp, desired) {
+				t.Errorf("expected %v, got %v", c.exp, desired)
+			}
+		})
+	}
+}
+
+func TestServerSideApplyWithIgnoreFields(t *testing.T) {
```
haoqing0110, line 426: Does it make sense if ignorefields is .metadata.xxx? For example, what if ignorefields is .metadata.annotations and type is IgnoreOnSpokePresent, will the existing annotations be kept on the spoke or be override by object-hash?

qiujian16: it will be kept. We compute the hash for the whole object including metadata.

PR #727, opened 2024-11-29
:seedling: enable leader election of klusterlet-agent on single node managed clusters

haoqing0110:
## Summary
The leader election of the klusterlet-agent is disabled on the single node managed cluster to speed up the restart procedure (https://github.com/open-cluster-management-io/registration-operator/pull/193). While it might cause that two pods with different configuration, such as images or bootstrap config, running at the same time. Some unexpected behavior happens in this situation, for example, when old and new agents both exist, they override hub-kubeconfig-secret with different value and trigger each other create csr constantly.

To fix those issues, it is necessary to enable the leader election of the klusterlet-agent on the single node managed cluster.

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/pull/695

### DIFF
--- a/pkg/registration/spoke/spokeagent.go
+++ b/pkg/registration/spoke/spokeagent.go
@@ -354,12 +354,23 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 		go bootstrapInformerFactory.Start(bootstrapCtx.Done())
 		go secretController.Run(bootstrapCtx, 1)
 
-		// wait for the hub client config is ready.
+		// Wait for the hub client config is ready.
+		// PollUntilContextCancel periodically executes the condition func `o.internalHubConfigValidFunc`
+		// until one of the following conditions is met:
+		// - condition returns `true`: Indicates the hub client configuration
+		//   is ready, and the polling stops successfully.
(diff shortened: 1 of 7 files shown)


pkg/registration/spoke/spokeagent.go
```diff
@@ -359,7 +359,10 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 		if err := wait.PollUntilContextCancel(bootstrapCtx, 1*time.Second, true, o.internalHubConfigValidFunc); err != nil {
 			// TODO need run the bootstrap CSR forever to re-establish the client-cert if it is ever lost.
 			stopBootstrap()
-			return fmt.Errorf("failed to wait for hub client config for managed cluster to be ready: %w", err)
+			// DO NOT return error if context is canceled, allows the leader to release the leadership.
```
qiujian16, line 362: looks good in general.

Just a question: do we need to doc in which case internalHubConfigValidFunc should return err and it means the agent return with error, and will start a new leader election?

qiujian16 (commented):
/approve

pkg/registration/spoke/spokeagent.go
haoqing0110: yes, more comments added.

Issue #514, opened 2024-06-12
Natively support OCM on EKS


jaswalkiranavtar:
**Describe the enhancement**
Currently OCM supports x509 certificates based authenticated only which is not supported by EKS. Hence, we [request](https://kubernetes.slack.com/archives/C01GE7YSUUF/p1718036433801529?thread_ts=1706477864.905629&cid=C01GE7YSUUF)  the csr controller on the hub and certificateController on the managed cluster to be extracted such that those controllers can be swapped by different vendor specific controllers leading to support on all platforms without making further changes to OCM code.

@qiujian16 has [proposed](https://kubernetes.slack.com/archives/C01GE7YSUUF/p1718070037638259?thread_ts=1706477864.905629&cid=C01GE7YSUUF) a regsitration-driver based solution.

mikeshng:
@qiujian16  shared this initial design doc:
https://docs.google.com/document/d/1HZG1mfHyAmjETjLK5ccE0tspCiXkTBmTzwMxzp6alL8/edit

mikeshng:
Qiu Jian's registration refactor WIP PR that is related to this issue: https://github.com/open-cluster-management-io/ocm/pull/535

jaswalkiranavtar:
Created a formal enhancement proposal for supporting ocm natively on eks

https://github.com/open-cluster-management-io/enhancements/pull/125/files

jaswalkiranavtar:
This issue has been marked stale. Can we do something to stop this from automatically closing?

bhperry:
It seems this is primarily aimed at supporting EKS hub with other clusters on AWS. What about using an EKS hub with spoke clusters that are in different clouds? I assume you would just need to provide AWS credentials to the registration agent. Would everything else work the same otherwise?

jaswalkiranavtar:
> It seems this is primarily aimed at supporting EKS hub with other clusters on AWS (using cross-account IAM, etc.). What about using an EKS hub with spoke clusters that are in different clouds? I assume you would just need to provide AWS credentials to the registration agent. Would everything else work the same otherwise?

I can think of a few things we may have to do extra in order to make it work in no EKS managed cluster:
1. Hub cluster creates IAM role for every spoke. For non EKS spoke, you can create spoke IAM role on hub IAM. You will also have to create an IAM user for every spoke and assign spoke IAM role to it. This will allow you to assume role  using IAM user creds on spoke. 
[...]

PR #729, opened 2024-11-29
:sparkles: Adding aws-cli to docker image

jaswalkiranavtar:
## Summary
This image will be use as initContainer to copy the aws-cli to klusterlet-agent when initialized with awsirsa auth type.
This will allow us to keep registration-operator docker image clean.

## Related issue(s)

### DIFF
--- a/build/Dockerfile.aws-cli
+++ b/build/Dockerfile.aws-cli
@@ -1,13 +1,13 @@
 FROM golang:1.22-bullseye AS builder
 ARG OS=linux
-ARG ARCH=amd64
+ARG ARCH=x86_64
 
 # Downloading aws-cli
 WORKDIR /tmp
 RUN apt-get update
 RUN apt-get install unzip
 RUN unzip -v
-RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+RUN curl "https://awscli.amazonaws.com/awscli-exe-${OS}-${ARCH}.zip" -o "awscliv2.zip"
 RUN unzip awscliv2.zip
 
 
--- a/build/Dockerfile.aws-cli
+++ b/build/Dockerfile.aws-cli
@@ -1,13 +1,14 @@
 FROM golang:1.22-bullseye AS builder
 ARG OS=linux
 ARG ARCH=x86_64
+ARG VERSION=2.22.8
 
 # Downloading aws-cli
 WORKDIR /tmp
 RUN apt-get update
 RUN apt-get install unzip
 RUN unzip -v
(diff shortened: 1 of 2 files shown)


build/Dockerfile.aws-cli
```diff
@@ -0,0 +1,25 @@
+FROM golang:1.22-bullseye AS builder
+ARG OS=linux
+ARG ARCH=amd64
+
+# Downloading aws-cli
+WORKDIR /tmp
+RUN apt-get update
+RUN apt-get install unzip
+RUN unzip -v
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
```
qiujian16, line 10: can we make this as arch agnostic? we would also need to support other arch.

qiujian16:
also would you help to remind me on why we do not use aws go sdk? https://pkg.go.dev/github.com/aws/aws-sdk-go-v2. I remember we discussed this before. It will be good to leave some comments so we do not miss the reason.

jaswalkiranavtar:
> also would you help to remind me on why we do not use aws go sdk? https://pkg.go.dev/github.com/aws/aws-sdk-go-v2. I remember we discussed this before. It will be good to leave some comments so we do not miss the reason.

Yes, that is because the eks kubeconfig uses aws cli. We discussed about it and @deads2k suggested an option to copy aws cli from init container to klusterlet agent container. With this we do not have to maintain 2 copies of registration operator docker image, one for csr and other for aws auth.

https://github.com/open-cluster-management-io/enhancements/pull/125#discussion_r1678003208

With

build/Dockerfile.aws-cli
```diff
@@ -0,0 +1,25 @@
+FROM golang:1.22-bullseye AS builder
+ARG OS=linux
+ARG ARCH=x86_64
```
jaswalkiranavtar, line 3: # For ARM
ARG ARCH=aarch64

build/Dockerfile.aws-cli
jaswalkiranavtar: Pushed the change, please advise if it looks ok. Will fix the commit sign-off tomorrow moning.

qiujian16: looks good

qiujian16:
/approve

@mikeshng I think we also need to create a repo in quay for this image.

mikeshng:
> /approve
> 
> @mikeshng I think we also need to create a repo in quay for this image.

Ok. I have submitted the repo creation request. Thank you all for your patience.

mikeshng (approved):
/lgtm

repo has been created `https://quay.io/repository/open-cluster-management/aws-cli`

PR #732, opened 2024-12-03
:seedling: Fix the GitHub stale workflow to correctly state remove the stale label

mikeshng:
## Summary
- Fix the GitHub stale workflow to correctly state remove the `stale` label instead of the `stable` label.
- Renamed the file from stable to stale as well

An example of the `stable` usage: https://github.com/open-cluster-management-io/ocm/issues/514#issuecomment-2463603076


qiujian16 (commented):
/approve
/lgtm

PR #735, opened 2024-12-03
:sparkles: Adding second group for aws auth flow

jaswalkiranavtar:
## Summary
Adding a second group, that will be used when the spoke starts registration with aws auth type. The aws solution uses [EKS access entries](https://docs.aws.amazon.com/eks/latest/userguide/creating-access-entries.html) and an access entry cannot contain a group name with "system:" prefix. More details can be found on [this](https://kubernetes.slack.com/archives/C01GE7YSUUF/p1733178937248749) slack thread.

> It can’t start with system:, eks:, aws:, amazon:, or iam:.

## Related issue(s)

### DIFF
--- a/pkg/registration/hub/manifests/rbac/managedcluster-clusterrolebinding.yaml
+++ b/pkg/registration/hub/manifests/rbac/managedcluster-clusterrolebinding.yaml
@@ -12,3 +12,6 @@ subjects:
 - kind: Group
   apiGroup: rbac.authorization.k8s.io
   name: system:open-cluster-management:{{ .ManagedClusterName }}
+- kind: Group
+  apiGroup: rbac.authorization.k8s.io
+  name: open-cluster-management:{{ .ManagedClusterName }}
--- a/pkg/registration/hub/manifests/rbac/managedcluster-registration-rolebinding.yaml
+++ b/pkg/registration/hub/manifests/rbac/managedcluster-registration-rolebinding.yaml
@@ -15,3 +15,6 @@ subjects:
   - kind: Group
     apiGroup: rbac.authorization.k8s.io
     name: system:open-cluster-management:{{ .ManagedClusterName }}
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: open-cluster-management:{{ .ManagedClusterName }}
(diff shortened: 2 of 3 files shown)


mikeshng (approved):
/lgtm

Expecting test(s) to come later when we are closer to the AWS/EKS integration?

jaswalkiranavtar:
> /lgtm
> 
> Expecting test(s) to come later when we are closer to the AWS/EKS integration?

Yes, we will add them later when we implement hub side changes. Hopefully, this addition should not break anything existing.

jaswalkiranavtar:
Looks like the e2e test failed because of intermittent error. Could you please re-run this build?

mikeshng:
> Looks like the e2e test failed because of intermittent error. Could you please re-run this build?

Re-running as requested.

pkg/registration/hub/manifests/rbac/managedcluster-clusterrolebinding.yaml
```diff
@@ -12,3 +12,6 @@ subjects:
 - kind: Group
   apiGroup: rbac.authorization.k8s.io
   name: system:open-cluster-management:{{ .ManagedClusterName }}
+- kind: Group
```
qiujian16, line 15: add some comments on the reasoning and plan for migration.

qiujian16:
cc @elgnay

qiujian16:
something to followup:
1. add this group as org in csr.
2. update subject in csr to remove system prefix
3. update addon to remove system prefix https://github.com/open-cluster-management-io/ocm/pull/735

pkg/registration/hub/manifests/rbac/managedcluster-clusterrolebinding.yaml
jaswalkiranavtar: We are adding a second group, as for aws flow we are trying to use [eks access entries](https://docs.aws.amazon.com/eks/latest/userguide/creating-access-entries.html) which doesn't allow us to allow access to group with system prefix "system".

Also based on kubernetes [docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-subjects), "system" prefix is a reserved prefix for kubernetes internal usage.

The reason to add second group is to not update existing group abruptly and cause failure while upgrade existing installation to newer version of OCM. So, the plan is adding second group now to support aws flow and later in the next version, delete old group.

jaswalkiranavtar: Also request you to drive the migration of existing group for csr flow on your end.

elgnay:
LGTM

PR #736, opened 2024-12-03
:bug: Do not create second SecretController used for credential rotation

jaswalkiranavtar:
## Summary
The aws sts token is short lived, generally with 15 mins TTL, hence we don't need the second secret controller for credentials rotation for aws flow.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/spoke/spokeagent.go
+++ b/pkg/registration/spoke/spokeagent.go
@@ -45,6 +45,7 @@ import (
 var AddOnLeaseControllerSyncInterval = 30 * time.Second
 
 const AwsIrsaAuthType = "awsirsa"
+const CsrAuthType = "csr"
 
 type SpokeAgentConfig struct {
 	agentOptions       *commonoptions.AgentOptions
@@ -430,14 +431,6 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 		return fmt.Errorf("failed to create CSR option: %w", err)
 	}
 
-	// create another RegisterController for registration credential rotation
-	controllerName := fmt.Sprintf("RegisterController@cluster:%s", o.agentOptions.SpokeClusterName)
-	secretController := register.NewSecretController(
-		secretOption, csrOption, o.driver, registration.GenerateStatusUpdater(
-			hubClusterClient,
-			hubClusterInformerFactory.Cluster().V1().ManagedClusters().Lister(),
-			o.agentOptions.SpokeClusterName), recorder, controllerName)
-
 	// create ManagedClusterLeaseController to keep the spoke cluster heartbeat
 	managedClusterLeaseController := lease.NewManagedClusterLeaseController(
 		o.agentOptions.SpokeClusterName,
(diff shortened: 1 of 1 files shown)


pkg/registration/spoke/spokeagent.go
```diff
@@ -517,6 +510,18 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
 		)
 	}
 
+	if o.registrationOption.RegistrationAuth == CsrAuthType {
```
mikeshng, line 513:
```suggestion
	if o.registrationOption.RegistrationAuth != AwsIrsaAuthType {
```
This might have less chance of regression.

qiujian16:
would you explain more on why this should not be started in aws case? Will the secret not change in aws case?

jaswalkiranavtar:
> would you explain more on why this should not be started in aws case? Will the secret not change in aws case?

Right the eks kubeconfig doesn't contain any password or other secret info. It uses exec plugin to generate sts token based on the role credentials set in container env variable and the [sts token](https://aws.github.io/aws-eks-best-practices/security/docs/iam/#controlling-access-to-eks-clusters) is by default 15 mins long. The kubeconfig automatically generates new token if previous one is expired without failing the request and letting the client know.

qiujian16:
> > would you explain more on why this should not be started in aws case? Will the secret not change in aws case?
> 
> Right the eks kubeconfig doesn't contain any password or other secret info. It uses exec plugin to generate sts token based on the role credentials set in container env variable and the [sts token](https://aws.github.io/aws-eks-best-practices/security/docs/iam/#controlling-access-to-eks-clusters) is by default 15 mins long. The kubeconfig automatically generates new token if previous one is expired without failing the request and letting the client know.

Do we need to disable the controller in this case? it is noop in aws scenario, right? If so I'd rather keep the controller running.

jaswalkiranavtar:
> > > would you explain more on why this should not be started in aws case? Will the secret not change in aws case?
> > 
> > 
> > Right the eks kubeconfig doesn't contain any password or other secret info. It uses exec plugin to generate sts token based on the role credentials set in container env variable and the [sts token](https://aws.github.io/aws-eks-best-practices/security/docs/iam/#controlling-access-to-eks-clusters) is by default 15 mins long. The kubeconfig automatically generates new token if previous one is expired without failing the request and letting the client know.
> 
> Do we need to disable the controller in this case? it is noop in aws scenario, right? If so I'd rather keep the controller running.

Wouldn't it be better to not have a controller that is not going to do anything but consuming CPU memory resources?

qiujian16:
I think this would be trivial in terms of resource consumption, But it brings better code maintainability.

pkg/registration/spoke/spokeagent.go
mikeshng: Resolving. See Qiu Jian's comment: https://github.com/open-cluster-management-io/ocm/pull/736#issuecomment-2517827947 Thanks.

jaswalkiranavtar:
This has been fixed as part of this PR: https://github.com/open-cluster-management-io/ocm/pull/744
hence closing this PR

PR #750, opened 2024-12-06
:sparkles: (WIP) Extract the logic to add cluster annotations to the driver interface and add unit and integration tests

dtclxy64:
## Summary
Extract the logic to add cluster annotations to the driver interface.
Add unit and integration tests.

## Related issue(s)

### DIFF
--- a/pkg/registration/register/interface.go
+++ b/pkg/registration/register/interface.go
@@ -72,8 +72,11 @@ type RegisterDriver interface {
 	// return nil, nil.
 	InformerHandler(option any) (cache.SharedIndexInformer, factory.EventFilterFunc)
 
-	// AddClusterAnnotations adds cluster annotations for non-CSR drivers
-	AddClusterAnnotations(clusterAnnotations map[string]string, managedClusterArn string, managedClusterRoleSuffix string)
+	// ManagedClusterDecorator is to change managed cluster metadata or spec during registration process.
+	ManagedClusterDecorator(cluster *clusterv1.ManagedCluster) *clusterv1.ManagedCluster
+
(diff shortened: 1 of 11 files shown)


mikeshng (commented):
Thanks for the PR!

It looks like it's currently failing `make verify`. Could you please check?

pkg/registration/register/interface.go
```diff
@@ -71,6 +71,9 @@ type RegisterDriver interface {
 	// InformerHandler returns informer of the related object. If no object needs to be watched, the func could
 	// return nil, nil.
 	InformerHandler(option any) (cache.SharedIndexInformer, factory.EventFilterFunc)
+
+	// AddClusterAnnotations adds cluster annotations for non-CSR drivers
+	AddClusterAnnotations(clusterAnnotations map[string]string, managedClusterArn string, managedClusterRoleSuffix string)
```
qiujian16, line 76: I could understand the intension. But I wonder it is suitable for an interface since it seems quite specific to a certain implementation. I am thinking whether it makes more sense to have a method like

```
ClusterDecorator(cluster *v1.ManagedCluster) *v1.ManagedCluster
```

and call it in the creatingCluster controller. 

I can try to create an example PR to show how it looks like.

qiujian16: https://github.com/open-cluster-management-io/ocm/pull/752

dtclxy64: @qiujian16 We are incorporating the changes from #752 into our changes. We'll let you know once we are ready for another around of reviews. Thanks for creating the example PR to demonstrate the changes.

PR #753, opened 2024-12-09
✨ Add importer into registration

qiujian16:
## Related issue(s)

Fixes #

/hold

### DIFF
--- a/pkg/registration/hub/importer/importer.go
+++ b/pkg/registration/hub/importer/importer.go
@@ -8,10 +8,10 @@ import (
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcehelper"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	appsv1 "k8s.io/api/apps/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -20,7 +20,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 
@@ -37,6 +36,12 @@ import (
 	cloudproviders "open-cluster-management.io/ocm/pkg/registration/hub/importer/providers"
 )
 
+const (
+	operatorNamesapce     = "open-cluster-management"
+	bootstrapSA           = "cluster-bootstrap"
+	conditionTypeImported = "Imported"
+)
+
 var (
 	genericScheme = runtime.NewScheme()
 	genericCodecs = serializer.NewCodecFactory(genericScheme)
@@ -48,8 +53,6 @@ func init() {
 	utilruntime.Must(apiextensionsv1.AddToScheme(genericScheme))
 	utilruntime.Must(operatorv1.Install(genericScheme))
 }
-
-const operatorNamesapce = "open-cluster-management"
 
 // KlusterletConfigRenderer renders the config for klusterlet chart.
 type KlusterletConfigRenderer func(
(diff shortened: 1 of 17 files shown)


pkg/registration/hub/importer/importer.go
```diff
@@ -0,0 +1,284 @@
...
+
+	// get provider from the provider list
+	var provider cloudproviders.Interface
+	for _, p := range i.providers {
+		if p.IsManagedClusterOwner(cluster) {
+			provider = p
+		}
+	}
+	if provider == nil {
+		logger.V(4).Info("provider not found for cluster", "cluster", cluster.Name)
```
zhujian7, line 123:
```suggestion
		logger.Info("provider not found for cluster", "cluster", cluster.Name)
```

pkg/registration/hub/importer/importer.go
```diff
@@ -0,0 +1,284 @@
...
+
+	"open-cluster-management.io/ocm/pkg/common/queue"
+	"open-cluster-management.io/ocm/pkg/operator/helpers/chart"
+	cloudproviders "open-cluster-management.io/ocm/pkg/registration/hub/importer/providers"
+)
+
+const (
+	operatorNamesapce     = "open-cluster-management"
+	bootstrapSA           = "cluster-bootstrap"
+	ConditionTypeImported = "Imported"
```
zhujian7, line 42: We have the condition `ManagedClusterJoined` and `ManagedClusterConditionAvailable`, consider changing this to `ManagedClusterImported` or `ManagedClusterConditionImported`

pkg/registration/hub/importer/importer.go
```diff
@@ -0,0 +1,284 @@
...
+			logger.Error(err, "failed to decode manifest", "manifest", manifest)
+			return cluster, err
+		}
+		result := resourceapply.ApplyResult{}
+		switch t := requiredObj.(type) {
+		case *appsv1.Deployment:
+			result.Result, result.Changed, result.Error = resourceapply.ApplyDeployment(
+				ctx, clients.KubeClient.AppsV1(), recorder, t, 0)
+			results = append(results, result)
+		case *operatorv1.Klusterlet:
```
zhujian7, line 217: Here I don't quite understand why we handle the Deployment and Klusterlet separately, instead of using the resourceapply.ApplyDirectly?

pkg/registration/hub/importer/providers/interface.go
```diff
@@ -0,0 +1,63 @@
...
+	Clients(ctx context.Context, cluster *clusterv1.ManagedCluster) (*Clients, error)
+
+	// IsManagedClusterOwner check if the provider is used to manage this cluster
+	IsManagedClusterOwner(cluster *clusterv1.ManagedCluster) bool
+
+	// Register registers the provider to the importer. The provider should enqueue the resource
+	// into the queue with the name of the managed cluster
+	Register(syncCtx factory.SyncContext)
+
+	// Run starts the provider
```
zhujian7, line 28: It is better to add more details for the Run comment, I don't know what I should do from the current comment to implement the `Run`.

pkg/registration/hub/importer/providers/interface.go
```diff
@@ -0,0 +1,63 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	operatorclient "open-cluster-management.io/api/client/operator/clientset/versioned"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+// Interface is the interface that a cluster provider should implement
+type Interface interface {
+	// Clients is to return the client to connect to the target cluster.
```
zhujian7, line 18: Should we add some comments about what permission the returned client should have?

zhujian7:
👍🏻 LGTM, left some nit comments.

pkg/registration/hub/importer/importer.go
qiujian16: it only support ApplyDirectly only can apply certain static resource, deployment/klusterlet can do be applied with this func.

pkg/registration/hub/importer/importer.go
qiujian16: maybe V(2), otherwise when we enable the controller, this logs will always be printed since it watches managedCluster

pkg/registration/hub/importer/renderers.go
```diff
@@ -0,0 +1,93 @@
...
+	"k8s.io/client-go/kubernetes"
+	clientcmdapiv1 "k8s.io/client-go/tools/clientcmd/api/v1"
+	"k8s.io/utils/ptr"
+
+	sdkhelpers "open-cluster-management.io/sdk-go/pkg/helpers"
+
+	"open-cluster-management.io/ocm/pkg/operator/helpers/chart"
+)
+
+func RenderBootstrapHubKubeConfig(
```
zhiweiyin318, line 19: could we consider to move the rednerers.go to the open-cluster-management.io/sdk-go? 
the other customized component like import-controller could consumer the same render func too.

pkg/registration/hub/importer/importer.go
```diff
@@ -0,0 +1,284 @@
...
+	// If the cluster is imported, skip the reconcile
+	if meta.IsStatusConditionTrue(cluster.Status.Conditions, ConditionTypeImported) {
+		return nil
+	}
+
+	// get provider from the provider list
+	var provider cloudproviders.Interface
+	for _, p := range i.providers {
+		if p.IsManagedClusterOwner(cluster) {
+			provider = p
```
zhiweiyin318, line 119: should break here ?
open discuss:
 Is there a case a cluster have multi providers ?  
  do we need to define a default provider?

qiujian16: yes we should

pkg/registration/hub/importer/renderers.go
qiujian16: I am still thinking whether in import-controller, we should import ocm so we could reuse not only render but importer as well.

Issue #755, opened 2024-12-09
E2E tests are using the wrong image tags


zhujian7:
**Describe the bug**
The github action e2e is using the wrong images, we expected the `e2e` tag images will be used, but it always uses the `latest` images.

```
╰─$ IMAGE_TAG=e2e make test-e2e
Using existing helm from "_output/tools/bin/helm"
_output/tools/bin/helm install cluster-manager deploy/cluster-manager/chart/cluster-manager --namespace=open-cluster-management --create-namespace --set images.registry=quay.io/open-cluster-management,image.tag=e2e
NAME: cluster-manager
LAST DEPLOYED: Mon Dec  9 21:32:22 2024
NAMESPACE: open-cluster-management
STATUS: deployed
REVISION: 1
TEST SUITE: None
kubectl config view --minify --flatten > ./.hub-kubeconfig
kubectl config set clusters.kind-kind.server https://10.96.0.1 --kubeconfig ./.hub-kubeconfig
Property "clusters.kind-kind.server" set.
_output/tools/bin/helm install klusterlet deploy/klusterlet/chart/klusterlet --namespace=open-cluster-management --set klusterlet.create=false,images.registry=quay.io/open-cluster-management,image.tag=e2e --set-file bootstrapHubKubeConfig=./.hub-kubeconfig
NAME: klusterlet
LAST DEPLOYED: Mon Dec  9 21:32:24 2024
NAMESPACE: open-cluster-management
STATUS: deployed
REVISION: 1
TEST SUITE: None
go test -c ./test/e2e
./e2e.test -test.v -ginkgo.v -nil-executor-validating=true -registration-image=quay.io/open-cluster-management/registration:e2e -work-image=quay.io/open-cluster-management/work:e2e -singleton-image=quay.io/open-cluster-management/registration-operator:e2e -klusterlet-deploy-mode=Default
=== RUN   TestE2E
Running Suite: ocm E2E Suite - /Users/jiazhu/go/src/open-cluster-management.io/ocm
==================================================================================
Random Seed: 1733751165

Will run 67 of 67 specs
------------------------------
[BeforeSuite]
/Users/jiazhu/go/src/open-cluster-management.io/ocm/test/e2e/e2e_suite_test.go:85
  STEP: Setup hub @ 12/09/24 21:32:45.333
  STEP: Setup spokeTestHelper @ 12/09/24 21:32:45.335
[...]

PR #756, opened 2024-12-10
🐛 Use e2e images instead of latest

zhujian7:
## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/755


qiujian16:
good catch...

PR #759, opened 2024-12-10
:sparkles: Cluster decorator interface

jaswalkiranavtar:
## Summary
Adding ManagedCluster annotations using a Decorator

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/register/aws_irsa/aws_irsa_test.go
+++ b/pkg/registration/register/aws_irsa/aws_irsa_test.go
@@ -230,7 +230,7 @@ func TestIsHubKubeConfigValidFunc(t *testing.T) {
 	}
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
-			driver := NewAWSIRSADriver()
+			driver := NewAWSIRSADriver("", "")
 			secretOption := register.SecretOption{
 				ClusterName:       c.clusterName,
 				AgentName:         c.agentName,
--- a/pkg/registration/register/interface.go
+++ b/pkg/registration/register/interface.go
@@ -73,7 +73,7 @@ type RegisterDriver interface {
 	InformerHandler(option any) (cache.SharedIndexInformer, factory.EventFilterFunc)
 
 	// ManagedClusterDecorator is to change managed cluster metadata or spec during registration process.
-	ManagedClusterDecorator(managedClusterArn string, managedClusterRoleSuffix string) ManagedClusterDecorator
+	ManagedClusterDecorator(cluster *clusterv1.ManagedCluster) *clusterv1.ManagedCluster
 }
 
 // Approvers is the inteface that each driver should implement on hub side. The hub controller will use this driver
@@ -87,5 +87,3 @@ type Approver interface {
 	// deletes rolebindings for the agent, and then this is the additional operation a driver should process.
 	Cleanup(ctx context.Context, cluster *clusterv1.ManagedCluster) error
 }
-
-type ManagedClusterDecorator func(cluster *clusterv1.ManagedCluster) *clusterv1.ManagedCluster
--- a/pkg/registration/register/interface.go
+++ b/pkg/registration/register/interface.go
@@ -73,7 +73,7 @@ type RegisterDriver interface {
 	InformerHandler(option any) (cache.SharedIndexInformer, factory.EventFilterFunc)
 
 	// ManagedClusterDecorator is to change managed cluster metadata or spec during registration process.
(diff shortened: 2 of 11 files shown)


pkg/registration/spoke/registration/creating_controller.go
```diff
@@ -94,37 +83,17 @@ func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx fac
...
-
-		if !isIncludeByExisting {
-			managedClusterClientConfigs = append(managedClusterClientConfigs, clusterv1.ClientConfig{
-				URL:      serverURL,
-				CABundle: c.spokeCABundle,
-			})
-		}
-	}
-	if len(existingCluster.Spec.ManagedClusterClientConfigs) == len(managedClusterClientConfigs) {
+	if len(existingCluster.Spec.ManagedClusterClientConfigs) == len(managedCluster.Spec.ManagedClusterClientConfigs) {
```
mikeshng, line 91: This if might be missing the annotations check?

pkg/registration/spoke/registration/creating_controller.go
```diff
@@ -94,37 +83,17 @@ func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx fac
...
 		return nil
 	}
 
 	// update ManagedClusterClientConfigs in ManagedCluster
```
mikeshng, line 95:
```suggestion
	// update ManagedClusterClientConfigs and annotations in ManagedCluster
```

mikeshng (commented):
Left some questions. Leaving it to @qiujian16 and other registration maintainers.

Part of https://github.com/open-cluster-management-io/ocm/issues/514

pkg/registration/register/interface.go
```diff
@@ -71,6 +71,9 @@ type RegisterDriver interface {
 	// InformerHandler returns informer of the related object. If no object needs to be watched, the func could
 	// return nil, nil.
 	InformerHandler(option any) (cache.SharedIndexInformer, factory.EventFilterFunc)
+
+	// ManagedClusterDecorator is to change managed cluster metadata or spec during registration process.
+	ManagedClusterDecorator(managedClusterArn string, managedClusterRoleSuffix string) ManagedClusterDecorator
```
qiujian16, line 76: in the interface level, the input should not be 
```
managedClusterArn string, managedClusterRoleSuffix string
```
since other driver does not know about this, and if we have another driver later that needs other input, this would need to change the signature of the interface.

The interface should be the same as ManagedClusterDecorator, which is only to mutate the managedcluster.
The managedClusterArn and managedClusterRoleSuffix could be put into the driver at the NewDriver part.

pkg/registration/spoke/registration/creating_controller.go
qiujian16: we only add annotation when create. The user could update the annotation, and the agent should not update it,

pkg/registration/spoke/spokeagent.go
```diff
@@ -191,20 +190,9 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
...
-			o.registrationOption.ClusterAnnotations[operatorv1.ClusterAnnotationsKeyPrefix+"/managed-cluster-arn"] = o.registrationOption.ManagedClusterArn
-			o.registrationOption.ClusterAnnotations[operatorv1.ClusterAnnotationsKeyPrefix+"/managed-cluster-iam-role-suffix"] =
-				o.registrationOption.ManagedClusterRoleSuffix
-
-		} else {
-			panic("A valid EKS Hub Cluster ARN is required with awsirsa based authentication")
-		}
+	var registrationOption = o.registrationOption
+	if registrationOption.RegistrationAuth == AwsIrsaAuthType {
+		registerDriver = awsIrsa.NewAWSIRSADriver()
```
qiujian16, line 195: let's change this to 
```
awsIrsa.NewAWSIRSADriver(o.registrationOption.ManagedClusterArn, o.registrationOption.ManagedClusterRoleSuffix)
```
so we do not need call decorator with these parameter.

pkg/registration/register/aws_irsa/aws_irsa_test.go
```diff
@@ -230,7 +230,7 @@ func TestIsHubKubeConfigValidFunc(t *testing.T) {
 	}
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
-			driver := NewAWSIRSADriver()
+			driver := NewAWSIRSADriver("", "")
```
jaswalkiranavtar, line 233: Please ignore this one, we are planning to fix this in future, when we implement the completion of joining, Did this now to resolve a compilation error.

PR #760, opened 2024-12-11
✨ Support configuring proxy for addons

zhujian7:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/addon/templateagent/decorator.go
+++ b/pkg/addon/templateagent/decorator.go
@@ -99,6 +99,7 @@ func newDeploymentDecorator(
 			newVolumeDecorator(addonName, template),
 			newNodePlacementDecorator(privateValues),
 			newImageDecorator(privateValues),
+			newProxyDecorator(privateValues),
 		},
 	}
 }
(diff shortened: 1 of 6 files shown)


pkg/addon/templateagent/decorator.go
```diff
@@ -329,6 +331,55 @@ func (d *imageDecorator) decorate(pod *corev1.PodTemplateSpec) error {
...
+			name:  "HTTPS_PROXY",
+			value: pc.HTTPSProxy,
+		})
+	}
+	if len(pc.NoProxy) > 0 {
+		keyValues = append(keyValues, keyValuePair{
+			name:  "NO_PROXY",
+			value: pc.NoProxy,
+		})
+	}
```
elgnay, line 373: I'm not entirely sure whether the environment variables for proxy settings should be uppercase, lowercase, or if both are supported.

zhujian7: It seems there is no standard and both uppercase and lowercase versions are used depending on the application

zhujian7: both uppercase and lowercase are added.

PR #770, opened 2024-12-17
:bug: Agents stop update managedcluster status when clock is out of sync.

xuezhaojun:
## Summary

This patch aim to stop agents from updating managed cluster status when clock is already out of sync.

The issue occurs when the hub [sets the managed cluster status to unknown due to an outdated lease](https://github.com/stolostron/ocm/blob/16190e3a34d4b7dc6c1f2a039da0230b6aec850d/pkg/registration/hub/lease/controller.go#L148-L153), while the agent simultaneously [sets the managed cluster status to available because the API server is accessible](https://github.com/stolostron/ocm/blob/16190e3a34d4b7dc6c1f2a039da0230b6aec850d/pkg/registration/spoke/managedcluster/resource_reconcile.go#L52-L53). This conflict causes the managed cluster status to flap back and forth frequently, triggering numerous reconciliations and resulting in performance issues.

### DIFF
--- a/pkg/registration/hub/lease/clocksynccontroller.go
+++ b/pkg/registration/hub/lease/clocksynccontroller.go
@@ -103,10 +103,15 @@ func (c *clockSyncController) sync(ctx context.Context, syncCtx factory.SyncCont
 		return err
 	}
 	// When the agent's lease get renewed, the "now" on hub should close to the RenewTime on agent.
-	// If the two time are not close(over 1 lease duration), we assume the clock is out of sync.
-	oneLeaseDuration := time.Duration(LeaseDurationSeconds) * time.Second
+	// If the two time are not close(the same duration in the lease controller), we assume the clock is out of sync.
+	// Then, if the Clock is out of sync, the agent will not be able to update the status of managed cluster.
+	leaseDuration := time.Duration(leaseDurationTimes*cluster.Spec.LeaseDurationSeconds) * time.Second
+	if leaseDuration == 0 {
+		leaseDuration = time.Duration(LeaseDurationSeconds*leaseDurationTimes) * time.Second
+	}
+
 	if err := c.updateClusterStatusClockSynced(ctx, cluster,
-		now.Sub(observedLease.Spec.RenewTime.Time) < oneLeaseDuration && observedLease.Spec.RenewTime.Time.Sub(now) < oneLeaseDuration); err != nil {
+		now.Sub(observedLease.Spec.RenewTime.Time) < leaseDuration && observedLease.Spec.RenewTime.Time.Sub(now) < leaseDuration); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


xuezhaojun:
> what are the consequences of not been able to update the ManagedCluster status? Do we lost any kind of feature? It would happen that we would need to have different clocks. Like a managedcluster, testing different functionalities related with time, ptp, ntp, and therefore, it would be "playing" with the clock.

The OCM is not designed to work when clock is not sync, if clocks are not synced, the first priority is to fix it.

pkg/registration/hub/lease/clocksynccontroller.go
```diff
@@ -103,8 +103,9 @@ func (c *clockSyncController) sync(ctx context.Context, syncCtx factory.SyncCont
 		return err
 	}
 	// When the agent's lease get renewed, the "now" on hub should close to the RenewTime on agent.
-	// If the two time are not close(over 1 lease duration), we assume the clock is out of sync.
```
xuezhaojun, line 106: @qiujian16  Before, we treat this condition as a warning, so the threshold is low, over 1 minute the the condition is show.

But now as we are using the condition to control agent's update actions, I think it's better to keep it the same as the duration we used in lease controller.

qiujian16: do we use the leaseDurationSeconds in the spec?

qiujian16:
We need also document this change in the website.

pkg/registration/hub/lease/clocksynccontroller.go
xuezhaojun: Goot point, updated.

xuezhaojun:
> We need also document this change in the website.

Agree, will add tasks for it.

test/integration/registration/managedcluster_lease_test.go
```diff
@@ -186,23 +186,31 @@ var _ = ginkgo.Describe("Cluster Lease Update", func() {
...
 
-		assertAvailableCondition(managedClusterName, metav1.ConditionTrue, 0)
+		assertAvailableCondition(managedClusterName, metav1.ConditionUnknown, 0)
 		assertCloclSyncedCondition(managedClusterName, metav1.ConditionFalse, 0)
+
+		// run agent again, check if conditions are updated to True
+		stop = runAgent(managedClusterName, agentOptions, commOptions, spokeCfg)
+		defer stop()
+
+		assertAvailableCondition(managedClusterName, metav1.ConditionTrue, 0)
```
xuezhaojun, line 212: @qiujian16 After run agent again, check if conditions become `True`

xuezhaojun:
@qiujian16  A task of updating doc is added: https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/issues/443

PR #777, opened 2024-12-23
🐛 Only enable auth driver when type is irsa

qiujian16:
This is to ensure during upgrade scenario, the new operator can still start the old version of the agent.

## Related issue(s)

Fixes #


qiujian16:
it currently causes the upgrade e2e test failure in clusteradm https://github.com/open-cluster-management-io/clusteradm/pull/461

PR #780, opened 2024-12-25
WIP ✨ Record the generation start and applied time for manifestwork

zhujian7:
## Related issue(s)

Fixes #


pkg/work/spoke/controllers/timestampcontroller/timestamp_controller.go
```diff
@@ -0,0 +1,188 @@
+package timestampcontroller
+
+import (
```
qiujian16, line 3: could we add this in the metrics-collector instead of here?

PR #783, opened 2024-12-31
✨ Mount the awscli volume when using AWS IRSA-based authentication.

dtclxy64:
## Summary
Add an initContainer to the Klusterlet deployment manifests and mount the awscli volume when using AWS IRSA-based authentication.
Update the test util function to check for both .aws and /awscli volume mounts.

## Related issue(s)

### DIFF
--- a/manifests/klusterlet/management/klusterlet-agent-deployment.yaml
+++ b/manifests/klusterlet/management/klusterlet-agent-deployment.yaml
@@ -121,6 +121,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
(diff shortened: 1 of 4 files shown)


manifests/klusterlet/management/klusterlet-agent-deployment.yaml
```diff
@@ -121,6 +121,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
+          {{if and .RegistrationDriver .RegistrationDriver.AuthType (eq .RegistrationDriver.AuthType "awsirsa")}}
```
qiujian16, line 124: seems to me you only need the eq part

dtclxy64: @qiujian16 We've addressed the comment by only checking the auth type value as suggested.

PR #785, opened 2025-01-02
🐛 fix work agent performance issue

elgnay:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
+++ b/pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
@@ -32,8 +32,10 @@ import (
 )
 
 var (
-	ResyncInterval     = 5 * time.Minute
-	MaxRequeueDuration = 24 * time.Hour
+	// ResyncInterval defines the maximum interval for resyncing a ManifestWork. It is used to:
+	//   1) Set the `ResyncEvery` for the `ManifestWorkAgent` controller;
+	//   2) Requeue a ManifestWork after it has been successfully reconciled.
+	ResyncInterval = 5 * time.Minute
 )
 
 type workReconcile interface {
@@ -138,7 +140,7 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
 	}
 	newAppliedManifestWork := appliedManifestWork.DeepCopy()
 
-	var requeueTime = MaxRequeueDuration
+	var requeueTime = ResyncInterval
 	var errs []error
 	for _, reconciler := range m.reconcilers {
 		manifestWork, newAppliedManifestWork, err = reconciler.reconcile(
@@ -173,7 +175,7 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
 	// we do not need to requeue when manifestwork/appliedmanifestwork are updated, since a following
 	// reconcile will be executed with update event, and the requeue can be set in this following reconcile
 	// if needed.
-	if !mwUpdated && !amwUpdated && requeueTime < MaxRequeueDuration {
+	if !mwUpdated && !amwUpdated {
 		controllerContext.Queue().AddAfter(manifestWorkName, requeueTime)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -175,6 +175,9 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
 	// if needed.
 	if !mwUpdated && !amwUpdated && requeueTime < MaxRequeueDuration {
 		controllerContext.Queue().AddAfter(manifestWorkName, requeueTime)
+	} else {
+		// resync each manifestwork every 5 minutes
+		controllerContext.Queue().AddAfter(manifestWorkName, 5*time.Minute)
```
qiujian16, line 180: I do not think we need else here, we do not need to requeue when manifestwork or appliedmanfiestwork is udpated. I think we only need to change this line
if !mwUpdated && !amwUpdated && requeueTime < MaxRequeueDuration
to 
if !mwUpdated && !amwUpdated

pkg/work/spoke/spokeagent.go
```diff
@@ -87,7 +87,9 @@ func (o *WorkAgentConfig) RunWorkloadAgent(ctx context.Context, controllerContex
 	if err != nil {
 		return err
 	}
-	spokeWorkInformerFactory := workinformers.NewSharedInformerFactory(spokeWorkClient, 5*time.Minute)
+	// resync with a small interval could result in performance issue when the number of appliedmanifestworks
+	// is large.
+	spokeWorkInformerFactory := workinformers.NewSharedInformerFactory(spokeWorkClient, 21*time.Hour)
```
qiujian16, line 93: 24 hours?

elgnay: If both manifestwork informer and appliedmanifestwork informer resync every 24 hours, they may resync at the same time that will result in a huge peak. However, if they have different resync intervals, it will bring less pressure to the work agent.

qiujian16: could we add some commenst on this?

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
elgnay: Updated as suggested.

pkg/work/spoke/spokeagent.go
elgnay: Comment added.

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -138,7 +137,7 @@ func (m *ManifestWorkController) sync(ctx context.Context, controllerContext fac
 	}
 	newAppliedManifestWork := appliedManifestWork.DeepCopy()
 
-	var requeueTime = MaxRequeueDuration
+	var requeueTime = ResyncInterval
```
qiujian16, line 140: ResyncInterval is also set in withResync. Should these two to be set the same?

elgnay: Do you mean `ResyncEvery`? It has already used the `ResyncInterval` although I think it is useless for the `ManifestWorkAgent` controller.
```go
WithSync(controller.sync).ResyncEvery(ResyncInterval).ToController("ManifestWorkAgent", recorder)
```

qiujian16: can we add some comment for ResyncInterval, given it is used at multiple locations.

elgnay: Done.

zhujian7:
@elgnay  So after this change, after a manifestwork containing a configmap is applied, if I delete the configmap on the manged cluster manually, how long will the work agent recreate the configmap? 24 hours?

elgnay:
> @elgnay So after this change, after a manifestwork containing a configmap is applied, if I delete the configmap on the manged cluster manually, how long will the work agent recreate the configmap? 24 hours?

Still each ManifestWork will be resynced at intervals not exceeding 5 minutes. The only change is that this logic no longer leverages the resync of the informers. In another word, the ManifestWork will be requeued by the controller (https://github.com/open-cluster-management-io/ocm/pull/785/files#diff-f0293b9137d86645469d2f7303ca98e1845a7b482671a25427e774a5c56c08abL177) after it has been processed successfully.

PR #795, opened 2025-01-09
:seedling: use images overrides in operator deploy

zhiweiyin318:
## Summary

The two makefile commands are used in e2e tests, but in some case, the image name will be overrode by the temp names. 
So here use the override images directly .

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e-test.mk
+++ b/test/e2e-test.mk
@@ -21,7 +21,7 @@ hub-kubeconfig:
 deploy-hub: deploy-hub-helm hub-kubeconfig cluster-ip
 
 deploy-hub-helm: ensure-helm
-	$(HELM) install cluster-manager deploy/cluster-manager/chart/cluster-manager --namespace=open-cluster-management --create-namespace --set images.registry=$(IMAGE_REGISTRY),images.tag=$(IMAGE_TAG)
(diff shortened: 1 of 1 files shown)


test/e2e-test.mk
```diff
@@ -21,7 +21,7 @@ hub-kubeconfig:
 deploy-hub: deploy-hub-helm hub-kubeconfig cluster-ip
 
 deploy-hub-helm: ensure-helm
-	$(HELM) install cluster-manager deploy/cluster-manager/chart/cluster-manager --namespace=open-cluster-management --create-namespace --set images.registry=$(IMAGE_REGISTRY),images.tag=$(IMAGE_TAG)
+	$(HELM) install cluster-manager deploy/cluster-manager/chart/cluster-manager --namespace=open-cluster-management --create-namespace --set images.overrides.operatorImage=$(OPERATOR_IMAGE_NAME),images.overrides.registrationImage=$(REGISTRATION_IMAGE),images.overrides.workImage=$(WORK_IMAGE),images.overrides.placementImage=$(PLACEMENT_IMAGE),images.overrides.addOnManagerImage=$(ADDON_MANAGER_IMAGE)
```
zhujian7, line 24: why do we remove the images.registry and images.tag?

PR #802, opened 2025-01-15
:seedling: rename bootstrap sa and rbac

zhiweiyin318:
## Related issue(s)

Fixes #


qiujian16 (commented):
/approve
/lgtm

note we also need to update clusteradm

PR #807, opened 2025-01-16
:sparkles: Assume prerequisite role on hub if initailize with aws-irsa

jaswalkiranavtar:
## Summary
If the hub is initialized with aws-irsa, it should assume a precreated IAM role with predefined named which will allow it permission to create IAM roles and policies for spokes that are trying to join.

## Related issue(s)

Ref: https://github.com/open-cluster-management-io/ocm/issues/514


pkg/common/helpers/parser.go
```diff
@@ -0,0 +1,15 @@
+package helpers
+
+import "strings"
+
+func GetAwsAccountIdAndClusterName(clusterArn string) (string, string) {
```
qiujian16, line 5: some comments on both of this func will be helpful.

manifests/cluster-manager/hub/cluster-manager-registration-serviceaccount.yaml
```diff
@@ -3,3 +3,7 @@ kind: ServiceAccount
 metadata:
   name: registration-controller-sa
   namespace: {{ .ClusterManagerNamespace }}
+  {{ if .ManagedClusterIdentityCreatorRole }}
+  annotations:
+    eks.amazonaws.com/role-arn: {{ .ManagedClusterIdentityCreatorRole }}
+  {{end}}
```
qiujian16, line 9: eof

pkg/common/helpers/parser.go
```diff
@@ -0,0 +1,15 @@
+package helpers
+
+import "strings"
```
qiujian16, line 3: you might want to change the file name to aws.go or awsparser.go

manifests/cluster-manager/hub/cluster-manager-registration-serviceaccount.yaml
```diff
@@ -3,3 +3,7 @@ kind: ServiceAccount
 metadata:
   name: registration-controller-sa
   namespace: {{ .ClusterManagerNamespace }}
+  {{ if .ManagedClusterIdentityCreatorRole }}
+  annotations:
+    eks.amazonaws.com/role-arn: {{ .ManagedClusterIdentityCreatorRole }}
```
qiujian16, line 8: this is not quite extensible, e.g. the key is specifically for eks. How do you think the keys/values could be configured from the template?

qiujian16:
/approve

some nit comment, I think we need to think about https://github.com/open-cluster-management-io/ocm/pull/807#discussion_r1919475761 for extensibility, but does not necessarily need to be resolved in this PR.

manifests/cluster-manager/hub/cluster-manager-registration-serviceaccount.yaml
jaswalkiranavtar: I think same needs to be applied on klusterlet SA as well. Will address it in a seperate PR.

pkg/common/helpers/parser.go
mikeshng: Verified that it has been renamed to aws.go

manifests/cluster-manager/hub/cluster-manager-registration-serviceaccount.yaml
mikeshng: Checked.

pkg/common/helpers/parser.go
mikeshng: Checked that comments has been added.

mikeshng (approved):
/lgtm

Verified that the final Qiu Jian's Slack comment regarding updating the cluster manager (remove registrationDriver) in clustermanager_aws_test.go has been addressed.

PR #810, opened 2025-01-17
:sparkles: ✨ feature feat: add registration drivers to cluster manager helm chart template.

### LINKED ISSUE #514: Natively support OCM on EKS

ramekris3163:
## Summary
Changed cluster manager helm template to support setting the required registrationDrivers if present.

## Related issue(s)

Ref: #514

### DIFF
--- a/deploy/cluster-manager/chart/cluster-manager/templates/cluster_manager.yaml
+++ b/deploy/cluster-manager/chart/cluster-manager/templates/cluster_manager.yaml
@@ -29,6 +29,10 @@ spec:
     featureGates:
       {{- toYaml . | nindent 6 }}
     {{- end }}
+    {{- with .Values.clusterManager.registrationConfiguration.registrationDrivers }}
+    registrationDrivers:
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
   {{- with .Values.clusterManager.workConfiguration }}
   workConfiguration:
     {{- toYaml . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


mikeshng (commented):
Do you need to make some changes to https://github.com/open-cluster-management-io/ocm/blob/main/deploy/cluster-manager/chart/cluster-manager/values.yaml as well? 

I don't see the `registrationDrivers` key in that file.

ramekris3163:
> Do you need to make some changes to https://github.com/open-cluster-management-io/ocm/blob/main/deploy/cluster-manager/chart/cluster-manager/values.yaml as well?
> 
> I don't see the `registrationDrivers` key in that file.

Fixed. Have added value to the values.yaml as required.

PR #819, opened 2025-01-31
✨ Made changes for clusteradm accept to create role and policies on hub

### LINKED ISSUE #514: Natively support OCM on EKS

alex0chan:
## Summary

This PR has the enhancement for clusteradm accept to create roles and policies on the hub

## Related issue(s)

Ref: #514


manifests/managed-cluster-policy/AccessPolicy.tmpl
```diff
@@ -0,0 +1,13 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "eks:DescribeCluster",
+        "eks:ListClusters"
```
jaswalkiranavtar, line 8: I think these permissions are not needed as we are copying cert authority data from bootstrap kubeconfig.

pkg/common/helpers/helpers.go
jaswalkiranavtar, line 1: If we have a similar method used by kluster-agent, can we reuse it?

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -203,6 +210,21 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 		}
 	}
 
+	// Only create new IAM roles when status is not present
+	if !meta.IsStatusConditionTrue(managedCluster.Status.Conditions, v1.ManagedClusterConditionHubAccepted) {
```
jaswalkiranavtar, line 214: Would it make sense to remove this if condition and overwrite role, policy and access entry on every reconciliation? This would also ensure that the IAM resources are repaired if somebody accidentally or intentionally changes them manually.

Also suggesting to overwrite directly instead of making a get call first and them comparing, to save an api call.

mikeshng (commented):
Could you check and maybe address some of the workflow errors if they are valid? Thanks.

jaswalkiranavtar:
> Could you check and maybe address some of the workflow errors if they are valid? Thanks.

Yes I guess this is still WIP, we created PR to get early feedback.

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -203,6 +210,21 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
...
+		clusterManager, err := c.operatorClient.OperatorV1().ClusterManagers().Get(context.TODO(), "cluster-manager", metav1.GetOptions{})
```
jaswalkiranavtar, line 215: Instead of making this API server call and adding another permission to clusterrole, can we instead pass this info from clustermanager CR to registration-controller on hub through these command line options?

qiujian16:
I am wondering whether we need to import aws sdk here, or we just ask user to install aws cli as the prereq?

jaswalkiranavtar:
> I am wondering whether we need to import aws sdk here, or we just ask user to install aws cli as the prereq?

Yes we need to use aws sdk here to create IAM roles and policies when hub admin approves registration as part of registration controller on hub. Preferring sdk over cli as with cli, we would  have to run shell commands from within go and handling errors would be complicated.

manifests/managed-cluster-policy/AccessPolicy.tmpl
suvaanshkumar: Shall we not have any permissions policy attached then?
Theoretically seems fine but have not tested this yet.

pkg/common/helpers/helpers.go
suvaanshkumar: We don't have this method anywhere else for the changes in this branch. If there will be more declarations coming up as a part of other PRs, We can con reuse there.

pkg/registration/hub/managedcluster/controller.go
suvaanshkumar: This seems to be a valid comment and we can take this in an upcoming change. 
We will have to add some logic to  check role, trust policy and permissions policy and if they have been modified or not. It might be a little bit more complicated as compared to checking kubernetes Roles and Rolebindings created because we can simply be reapply the yaml on them where as over here we might have to check the resources using sdk.

manifests/managed-cluster-policy/TrustPolicy.tmpl
```diff
@@ -0,0 +1,20 @@
...
+                "StringEquals": {
+                    "aws:PrincipalTag/hub_cluster_account_id": "{{.hubAccountId}}",
+                    "aws:PrincipalTag/hub_cluster_name": "{{.hubClusterName}}",
+                    "aws:PrincipalTag/managed_cluster_account_id": "{{.managedClusterAccountId}}",
+                    "aws:PrincipalTag/managed_cluster_name": "{{.managedClusterName}}"
+                }
+            }
+        }
+    ]
+}
```
qiujian16, line 20: eof

pkg/registration/hub/managedcluster/controller.go
qiujian16: Is the intention to only create permission once? It might not work since the condition patch might fail, and the is called again. Can we ensure that the createPermission can be called multiple times when the accept=true which is better to tolerate error.

jaswalkiranavtar: It is already capable of doing that. In the next attempt, it will just say, role, policy or access entry already exists and do nothing. However, we will create a seperate PR to implement crossplane style continuous reconciliation, to being back AWS resource to original state.

qiujian16: If it is capable already, I would consider not to update upon this condition. We just need to ignore the alreadyExist error.

manifests/managed-cluster-policy/TrustPolicy.tmpl
jaswalkiranavtar: You mean add a new line at the end of file correct?

qiujian16: yes

pkg/registration/hub/managedcluster/controller.go
jaswalkiranavtar: Addressed.

pkg/registration/hub/manager.go
```diff
@@ -64,10 +68,14 @@ func NewHubManagerOptions() *HubManagerOptions {
 func (m *HubManagerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringSliceVar(&m.ClusterAutoApprovalUsers, "cluster-auto-approval-users", m.ClusterAutoApprovalUsers,
 		"A bootstrap user list whose cluster registration requests can be automatically approved.")
+	fs.StringSliceVar(&m.EnabledRegistrationDrivers, "enabled-registration-drivers", m.EnabledRegistrationDrivers,
+		"A list of registration drivers enabled on hub.")
```
qiujian16, line 72: this csr driver should be set by default, for backward compatible

pkg/registration/register/aws_irsa/aws_irsa.go
```diff
@@ -112,3 +123,242 @@ func NewAWSIRSADriver(managedClusterArn string, managedClusterRoleSuffix string,
...
+	return false
+}
+
+func (a *AWSIRSADriverForHub) CreatePermissions(ctx context.Context, cluster *clusterv1.ManagedCluster) error {
+
+	if !a.allows(cluster) {
+		return nil
+	}
+	//Creating config for aws
+	cfg, err := config.LoadDefaultConfig(context.TODO())
```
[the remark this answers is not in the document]
jaswalkiranavtar: Addressed

pkg/registration/register/aws_irsa/aws_irsa.go
```diff
@@ -112,3 +123,242 @@ func NewAWSIRSADriver(managedClusterArn string, managedClusterRoleSuffix string,
...
+		log.Printf("Failed to create Access Entries for aws irsa %v", err)
+	}
+
+	return nil
+}
+
+// This function creates:
+// 1. IAM Roles and Policies in the hub cluster IAM
+// 2. Returns the hubclustername and the roleArn to be used for Access Entry creation
+func CreateIAMRoleAndPolicy(ctx context.Context, hubClusterArn string, managedCluster *v1.ManagedCluster, iamClient *iam.Client) (string, string, error) {
```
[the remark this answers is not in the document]
jaswalkiranavtar: Addressed

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -203,6 +208,20 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
...
+	// Only create new IAM roles when status is not present
+	if !meta.IsStatusConditionTrue(managedCluster.Status.Conditions, v1.ManagedClusterConditionHubAccepted) {
+		if err != nil {
+			log.Printf("Failed to get cluster manager %v", err)
+			errs = append(errs, err)
+		}
+
+		err = c.registerDriverForHub.CreatePermissions(ctx, managedCluster)
+		if err != nil {
+			log.Printf("Failed to create permissions %v", err)
```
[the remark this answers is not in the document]
jaswalkiranavtar: addressed. Removed log.printf everywhere and used klog instead.

pkg/registration/register/interface.go
```diff
@@ -87,3 +87,9 @@ type Approver interface {
 	// deletes rolebindings for the agent, and then this is the additional operation a driver should process.
 	Cleanup(ctx context.Context, cluster *clusterv1.ManagedCluster) error
 }
+
+type RegisterDriverForHub interface {
```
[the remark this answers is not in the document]
jaswalkiranavtar: Addressed

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -203,6 +208,20 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 		}
 	}
 
+	// Only create new IAM roles when status is not present
+	if !meta.IsStatusConditionTrue(managedCluster.Status.Conditions, v1.ManagedClusterConditionHubAccepted) {
+		if err != nil {
+			log.Printf("Failed to get cluster manager %v", err)
```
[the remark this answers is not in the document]
jaswalkiranavtar: Addressed

pkg/registration/hub/manager.go
```diff
@@ -147,12 +155,32 @@ func (m *HubManagerOptions) RunControllerManagerWithInformers(
 	workInformers workv1informers.SharedInformerFactory,
 	addOnInformers addoninformers.SharedInformerFactory,
 ) error {
+
 	csrApprover, err := csr.NewCSRApprover(kubeClient, kubeInformers, m.ClusterAutoApprovalUsers, controllerContext.EventRecorder)
 	if err != nil {
 		return err
 	}
+	awsIRSAHubDriver, err := awsirsa.NewAWSIRSAHubDriver(ctx, m.HubClusterArn)
```
qiujian16, line 163: this is a bit completed here. I was thinking an better approach would be

```go
var approvers []register.Approver
var drivers []register.HubDriver

for _, driverName := range m.EnabledRegistrationDrivers:
   switch driverName {
   case "csr":
       approvers = append(approvers, csr.NewCSRApprover(kubeClient, kubeInformers, m.ClusterAutoApprovalUsers, controllerContext.EventRecorder))
       drivers = append(drivers, csr.NewCSRHubDriver())     
   case "isra":
       ....  
   }
approver = register.NewAggregatedApprover(approvers...)
hubDriver = register.NewAggregatedHubDriver(drivers...)
```

pkg/registration/register/interface.go
```diff
@@ -87,3 +87,12 @@ type Approver interface {
 	// deletes rolebindings for the agent, and then this is the additional operation a driver should process.
 	Cleanup(ctx context.Context, cluster *clusterv1.ManagedCluster) error
 }
+
+// HubDriver interface is used to implement operations required to complete aws-irsa registration and csr registration.
+// The Approver interface above is used to implement operations related to approving the CSR request, and permission
+// creation is not related to CSR approval. Hence, created CreatePermissions under a separate interface.
+type HubDriver interface {
```
qiujian16, line 94: hrm that is still not clear to me. Approver interface is not meant to used only for CSR, but to be implemented by each driver. It is weird if we have two interface for two different implementations. 

How about isra implements the approver interface by watching managedcluster and call aws, which also decompose the code from managedcluster controller which does not seem to be a bad idea to me.

jaswalkiranavtar: AWS registration doesn't have anything to approve. Only thing it depends on is acceptance of ManagedCluster and then creation of IAM role and policies. So we have following cases:
1. Approver: Applicable to CSR auth only
2. HubDriver: Applicable to both CSR and AWS auth
     a) For CSR auth: It will contain logic for whitelisted user list for auto approval
     b) For AWS auth: It will contain logic for whitelist for auto approval as well as creation of IAM resources for permissions.

Please note that we are planning to anyway create a separate interface for auto-approval as discussed in [this](https://cloud-native.slack.com/archives/C089E2CMT2A/p1739323490901029?thread_ts=1739228212.786339&cid=C089E2CMT2A) thread. We are planning to merge HubDriver interface and the new interface for auto-approval as a single interface. A separate controller may be required as we have everything in managedcluster controller, also it may be more maintainable to keep everything related to acceptance in a single controller.

pkg/registration/hub/manager.go
jaswalkiranavtar: Addressed

pkg/registration/hub/manager.go
jaswalkiranavtar: Addressed

pkg/registration/register/interface.go
qiujian16: > Approver: Applicable to CSR auth only

if it is applicable to CSR only, we should not have an approver interface. But I thought you are still need to implement approver interface for isra auto approver starting a new controller? 

I think for driver interface on hub side, it should contain
1. action needs to be done when cluster is accepted
2. cleanup action when cluster is deleting or hubAcceptClient=false
If so, at least the Cleanup method interface for approver should be moved into the driver interface. But I still feel the two interface should be merge to 1 and the name could be hubDriver if it is better.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
```diff
@@ -73,6 +73,21 @@ func (c *runtimeReconcile) reconcile(ctx context.Context, cm *operatorapiv1.Clus
...
+	var enabledRegistrationDrivers []string
+	if cm.Spec.RegistrationConfiguration != nil {
+		for _, registrationDriver := range cm.Spec.RegistrationConfiguration.RegistrationDrivers {
+			enabledRegistrationDrivers = append(enabledRegistrationDrivers, registrationDriver.AuthType)
+			if registrationDriver.AuthType == "awsirsa" {
+				config.HubClusterArn = registrationDriver.HubClusterArn
+			}
+		}
+		// If no registration drivers are enabled, default to csr for backward compatibility
+		if len(enabledRegistrationDrivers) == 0 {
```
qiujian16, line 85: since the default flag as csr as the default driver in registration controller already, we  do not to set default here. And it is better not to set new flag here for backward compatible.

pkg/registration/hub/manager.go
qiujian16: let's set default value in registration controller instead setting in the operator, so if the flag is not specified. we will still use the csr driver.

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -203,6 +207,12 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 		}
 	}
 
+	// If the resources already exist, errors are ignored
+	err = c.hubDriver.CreatePermissions(ctx, managedCluster)
```
zhiweiyin318, line 211: can these permissions be cleaned up after the cluster is deleted?

jaswalkiranavtar: We are planning to clean them up on setting hubAcceptsClient to false similar to how other rbac resources are cleaned up in managedcluster controller on hub. We will implement this in a seperate PR.

Shall we also clean them up on ManagedCluster cleanup?

jaswalkiranavtar:
@qiujian16 could you please run the tests so that we can get test execution feedback on latest commit.

pkg/registration/hub/managedcluster/controller.go
suvaanshkumar: I think we should clean up these resources on managed cluster cleanup only if Rbacs and other resources also gets cleaned up which I think stays.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
jaswalkiranavtar: Addressed

pkg/registration/hub/manager.go
jaswalkiranavtar: Addressed

pkg/registration/hub/managedcluster/controller.go
qiujian16: RBAC is cleaned when cluster is deleted. This is done in GC controller, but this PR https://github.com/open-cluster-management-io/ocm/pull/831 is to refactor.

jaswalkiranavtar:
@qiujian16 The e2e-singleton test failure seems to be unrelated or intermittent error. Is it possible to rerun the test?

jaswalkiranavtar:
Also, if everything except approver refactoring looks good, can we get an approval today? Will address approver changes in next PR.

qiujian16:
yes, I will tag this so we can also merge another refactor PR @zhiweiyin318 

followup:

1. interface change
2. cleanup implementation.

PR #829, opened 2025-02-12
:bug: watch ns in gc controller

zhiweiyin318:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/helpers/helpers.go
+++ b/pkg/registration/helpers/helpers.go
@@ -176,3 +176,14 @@ func IsCSRSupported(nativeClient kubernetes.Interface) (bool, bool, error) {
 	}
 	return v1CSRSupported, v1beta1CSRSupported, nil
 }
+
+func HasFinalizer(finalizers []string, finalizer string) bool {
+	found := false
+	for i := range finalizers {
+		if finalizers[i] == finalizer {
+			found = true
+			break
+		}
+	}
+	return found
+}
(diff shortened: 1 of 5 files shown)


pkg/registration/helpers/helpers.go
```diff
@@ -176,3 +176,14 @@ func IsCSRSupported(nativeClient kubernetes.Interface) (bool, bool, error) {
 	}
 	return v1CSRSupported, v1beta1CSRSupported, nil
 }
+
+func HasFinalizer(finalizers []string, finalizer string) bool {
```
qiujian16, line 180: don't we already have a common finalizer utility?

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -194,12 +194,19 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
...
-		append(manifests.ClusterSpecificRBACFiles, manifests.ClusterSpecificRoleBindings...)...)
-	for _, result := range resourceResults {
-		if result.Error != nil {
-			errs = append(errs, result.Error)
+	// the ManagedClusterFinalizer is added and removed by gc controller.
+	// the rbac resources are deleted by gc controller too.
+	// should apply the rbac resources after the finalizer is added to the cluster, because in the case the cluster is
+	// created and deleted so quickly, the rbac resources are applied but the cluster is gone, there is no event to
+	// gc controller to trigger gc .
+	if helpers.HasFinalizer(managedCluster.Finalizers, v1.ManagedClusterFinalizer) {
```
qiujian16, line 202: a better way is directly return when there is no finalizer. This should be done at the beginning of the sync.

qiujian16 (commented):
need some test on this. A quick create and delete might be good enough.

PR #831, opened 2025-02-13
:seedling: refator gc and move rbac deletion to cluster controller

zhiweiyin318:
## Related issue(s)

Fixes #


pkg/registration/hub/managedcluster/controller.go
```diff
@@ -105,8 +130,21 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 	newManagedCluster := managedCluster.DeepCopy()
 
 	if !managedCluster.DeletionTimestamp.IsZero() {
-		// the cleanup job is moved to gc controller
-		return nil
+		requeue, err := c.removeClusterRbac(ctx, managedClusterName, true)
```
qiujian16, line 133: I think we should run cleanup in approver at first, then removeClusterRBAC. Also use requeu error.

zhiweiyin318: done

PR #850, opened 2025-02-26
:sparkles: Add managed cluster auto-approval for awsirsa

jaswalkiranavtar:
## Summary
Implemented auto approval based in list of ARN patterns

## Related issue(s)


pkg/registration/register/csr/hub_driver.go
```diff
@@ -190,30 +197,40 @@ func NewCSRHubDriver(
...
 	// noop
 	return nil
 }
+
+func (c *CSRHubDriver) Accept(cluster *clusterv1.ManagedCluster) bool {
+	return c.allows(cluster)
+}
+
+func (c *CSRHubDriver) allows(cluster *clusterv1.ManagedCluster) bool {
+	_, isManagedClusterArnPresent := cluster.Annotations["agent.open-cluster-management.io/managed-cluster-arn"]
```
qiujian16, line 233: csr hub driver should not consider these annotations. I think what this needs to do is to check the related csr on the hub, and if there is existing one which is created by the approved user, the cluster can be accepted. 

cc @skeeey

jaswalkiranavtar: Existing acceptance of csr ManagedCluster doesn't check anything related to csr. It would be a new feature.

jaswalkiranavtar: I think the problem here is we get managedcluster object in managedcluster controller but not csr, but this controller is triggered based on changes to o ly managedcluster cr.

We will.have to make another api call to fetch csr. Would that be acceptable?

Also there can be multiple csrs for a single spoke but only one Managedcluster cr, how will we decide that which csr to look at?

Seems like there is additional complication involved here, request other registration experts to help implement here.

jaswalkiranavtar: We added this annotation condition here to check that anything that doesn't fir awsirsa is a csr request.

skeeey: I think for `CSRHubDriver. Accept` we should return `true` directly  and don‘t do other checks to keep the behavior: https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/managedcluster/controller.go#L140

pkg/registration/hub/manager.go
```diff
@@ -76,6 +78,8 @@ func (m *HubManagerOptions) AddFlags(fs *pflag.FlagSet) {
 			"The flag works only when ResourceCleanup feature gate is enable.")
 	fs.StringVar(&m.HubClusterArn, "hub-cluster-arn", m.HubClusterArn,
 		"Hub Cluster Arn required to connect to Hub and create IAM Roles and Policies")
+	fs.StringSliceVar(&m.AutoApprovedCSRUsers, "auto-approved-csr-users", m.AutoApprovedCSRUsers, "")
+	fs.StringSliceVar(&m.AutoApprovedARNPatterns, "auto-approved-arn-patterns", m.AutoApprovedARNPatterns, "")
```
skeeey, line 82: please add description for the flag `auto-approved-csr-users` and `auto-approved-arn-pattern`

And for flag `cluster-auto-approval-users`, we should mark it deprecated, because the `auto-approved-csr-users` will be instead of `cluster-auto-approval-users`

pkg/registration/hub/manager.go
```diff
@@ -159,13 +163,13 @@ func (m *HubManagerOptions) RunControllerManagerWithInformers(
 	for _, enabledRegistrationDriver := range m.EnabledRegistrationDrivers {
 		switch enabledRegistrationDriver {
 		case "csr":
-			csrDriver, err := csr.NewCSRHubDriver(kubeClient, kubeInformers, m.ClusterAutoApprovalUsers, controllerContext.EventRecorder)
+			csrDriver, err := csr.NewCSRHubDriver(kubeClient, kubeInformers, m.ClusterAutoApprovalUsers, m.AutoApprovedCSRUsers, controllerContext.EventRecorder)
```
skeeey, line 166: I think we should not introduce a new parameter `autoApprovedCSRUsers`, we may rename the  `clusterAutoApprovalUsers ` to `autoApprovedCSRUsers ` and  handle the compatibility here, as long as `autoApprovedCSRUsers` is set, we will use it, else, use the `clusterAutoApprovalUsers`

pkg/registration/register/csr/hub_driver.go
qiujian16: > I think for CSRHubDriver. Accept we should return true

I think the same. I think the aggreatedAccept func should check all driver accepts it, it should make an AND rather than OR check. And in isra driver, it should return true if the cluster does not have certain annotations

jaswalkiranavtar: @qiujian16 do you mean, it should return **false**(not true) if the cluster does not have certain annotations? I believe that is the only way ANDing them would work instead of ORing.

pkg/registration/hub/manager.go
jaswalkiranavtar: Like this?

```fs.MarkDeprecated("cluster-auto-approval-users", "use auto-approved-csr-users flag instead of this")```

Comment (jaswalkiranavtar) [reply pkg/registration/hub/manager.go]: addressed

File: pkg/registration/hub/manager.go (hunk 3 above)
Comment (jaswalkiranavtar) [reply pkg/registration/hub/manager.go]: addressed

File: pkg/registration/register/csr/hub_driver.go (hunk 1 above)
Comment (jaswalkiranavtar) [reply pkg/registration/register/csr/hub_driver.go]: Your suggestion works, please disregard above message.

File: pkg/registration/register/common.go [hunk 4]
```diff
@@ -196,6 +200,15 @@ func (a *AggregatedHubDriver) CreatePermissions(ctx context.Context, cluster *cl
 
 }
 
+func (a *AggregatedHubDriver) Accept(cluster *clusterv1.ManagedCluster) bool {
```
qiujian16, line 203: nit, add a comment here on it will need each driver to accept.

pkg/registration/register/interface.go
```diff
@@ -91,4 +91,8 @@ type HubDriver interface {
 	// Cleanup is executed when hubAcceptClient in ManagedCluster is set false or cluster is deleting. The hub controller
 	// deletes rolebindings for the agent, and then this is the additional operation a driver should process.
 	Cleanup(ctx context.Context, cluster *clusterv1.ManagedCluster) error
+
+	// Accept is executed when autoapproval is enabled to verify if the managedcluster uses csr or awsirsa and
+	// setting hubAcceptClient based on autoApprovedIdentities
+	Accept(cluster *clusterv1.ManagedCluster) bool
```
qiujian16, line 97: nit, add some comment on if the cluster is not managed by the driver, the driver should return accept=true.

qiujian16 (commented):
/approve

some nit comments

leave to @skeeey for final review.

jaswalkiranavtar:
Addressed both above coments, @mikeshng please review

pkg/registration/register/common.go
mikeshng: Validated that the Qiu Jian's comment has been addressed.

pkg/registration/register/interface.go
mikeshng: Validated that the Qiu Jian's comment has been addressed.

jaswalkiranavtar:
@mikeshng i am hoping both @skeeey and @qiujian16 looked at it  yesterday and they are ok with it. 

We are waiting for this merge as we want to rebase another PR on it. Currently we are blocked because of this PR.

PR #855, opened 2025-02-27
:seedling: chore: e2e hosted and singleton should not be run on DOC-ONLY changes

mikeshng:
## Summary
e2e hosted and singleton GitHub workflow should not be run on DOC-ONLY changes

## Related issue(s)

Fixes NA


mikeshng:
@qiujian16 could you PTAL and this PR so doc fixes won't run the two unnecessary e2e (hosted, singleton) github actions?  Thanks.

Issue #510, opened 2024-06-12
addon is default managed by addon-manager


haoqing0110:
**Describe the enhancement**
A clear and concise description of what the enhancement request is.

Next step of https://github.com/open-cluster-management-io/ocm/issues/355 , when most of the addons migrate to be managed by addon-manager.

**OCM future**
- The Addon-manager will not add annotation `addon.open-cluster-management.io/lifecycle: addon-manager` to cma any more and the addon is default (no annotation) managed by addon-manager.
- Stop supporting the addon-framework version v0.8.1 (no bug fix on v0.8.z).
- Minimal supported version is v0.9.3. (TO BE STOP in the future).

- [x] addon-framework: when annotation is empty default managed by addon-manager https://github.com/open-cluster-management-io/addon-framework/pull/302
- [x] ocm: do not add lifecycle annotation to cma and default managed by addon-manager https://github.com/open-cluster-management-io/ocm/pull/856
- [x] doc: update addon part in website https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/pull/480

haoqing0110:
@qiujian16 I will update the website supported version, then can mark this as done.

PR #856, opened 2025-02-27
:warning: breaking change: do not add lifecycle annotation to cma and default managed by addon-manager

haoqing0110:
## Related issue(s)

Ref: https://github.com/open-cluster-management-io/ocm/issues/510


haoqing0110:
/assign @qiujian16
cc @zhujian7

PR #862, opened 2025-03-03
:sparkles: Make ocm changes to tag aws resources by OCM hub

amrcoder:
## Summary
Given the hub is initialized using awsirsa registration-auth and a list of tags are passed.
While manually or auto approving a ManagedCluster registration request, then the created aws resources like role, policy and access entry are tagged with list of passed tags.

## Related issue(s)

Fixes # [https://github.com/open-cluster-management-io/ocm/issues/514](https://github.com/open-cluster-management-io/ocm/issues/514)


qiujian16 (commented):
/approve
/assign @zhiweiyin318 would you review on the operator part?

PR #865, opened 2025-03-05
:bug: ARN pattern should be comma separated string instead of a string enclosed in square brackets

jaswalkiranavtar:
## Summary
We are using StringSliceVar type to store AutoApprovedARNPatterns [here](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/manager.go#L87).

Hence, cluster-manager-operator should convert the [array](https://github.com/open-cluster-management-io/ocm/blob/main/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go#L82) into comma separated list of ARN patterns, as that is what registration-controller binary will accept. 

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/514

### DIFF
--- a/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
+++ b/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
@@ -79,7 +79,10 @@ func (c *runtimeReconcile) reconcile(ctx context.Context, cm *operatorapiv1.Clus
 			enabledRegistrationDrivers = append(enabledRegistrationDrivers, registrationDriver.AuthType)
 			if registrationDriver.AuthType == "awsirsa" {
 				config.HubClusterArn = registrationDriver.HubClusterArn
-				config.AutoApprovedARNPatterns = registrationDriver.AutoApprovedIdentities
+				var AutoApprovedIdentities = registrationDriver.AutoApprovedIdentities
+				if AutoApprovedIdentities != nil {
+					config.AutoApprovedARNPatterns = strings.Join(AutoApprovedIdentities, ",")
+				}
 			} else if registrationDriver.AuthType == "csr" {
 				config.AutoApprovedCSRUsers = registrationDriver.AutoApprovedIdentities
 			}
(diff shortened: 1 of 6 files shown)


test/integration/operator/clustermanager_aws_test.go
```diff
@@ -77,5 +80,27 @@ var _ = ginkgo.Describe("ClusterManager Default Mode with aws registration", fun
...
+				return present && strings.SplitN(autoApprovedArnPatterns, "=", 2)[1] ==
+					"arn:aws:eks:us-west-2:123456789013:cluster/.*,arn:aws:eks:us-west-2:123456789012:cluster/.*"
+			}, eventuallyTimeout, eventuallyInterval).Should(gomega.BeTrue())
+		})
 	})
 })
+
+func findMatchingArg(args []string, pattern string) (string, bool) {
+	for _, commandLineArg := range args {
+		if strings.Contains(commandLineArg, pattern) {
```
mikeshng, line 101: nit: probably should do the exact equals check here in case you have multiple arguments with similar strings. ie: `--foo-bar-foo`  `--foo-bar` and doing a `--foo-bar` will just grab the first one which is `--foo-bar-foo`.

mikeshng (approved):
/lgtm

/assign @qiujian16

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
```diff
@@ -79,7 +79,10 @@ func (c *runtimeReconcile) reconcile(ctx context.Context, cm *operatorapiv1.Clus
 			enabledRegistrationDrivers = append(enabledRegistrationDrivers, registrationDriver.AuthType)
 			if registrationDriver.AuthType == "awsirsa" {
 				config.HubClusterArn = registrationDriver.HubClusterArn
-				config.AutoApprovedARNPatterns = registrationDriver.AutoApprovedIdentities
+				var AutoApprovedIdentities = registrationDriver.AutoApprovedIdentities
```
qiujian16, line 82: I do not think this line is necessary.

can we just strings.join(registrationDriver.AutoApprovedIdentities, ",")

and why CSR part does not need this change?

jaswalkiranavtar: Addressed. I think we overlooked it. Fixed for CSR as well, as well as added an integration test for csr.

Issue #864, opened 2025-03-05
[OCM Incubation] Update the arch diagram to add the addon manager


zhujian7:
The arch should be updated to add the global addon manager:
- https://open-cluster-management.io/docs/concepts/architecture/#hub-spoke-architecture
- https://github.com/open-cluster-management-io/ocm/blob/main/assets/ocm-arch.drawio

gitatractivo:
Hi, We need to add global cluster-manager-addon-manager right? Inside cluster manager?

gitatractivo:
Hi I wanted to know does this diagram close the issue? or any more changes required?

zhujian7:
> Hi I wanted to know does this diagram close the issue? or any more changes required?

Hi @gitatractivo, no, the PR just make the diagram more pull model like, but did not add the addon manager.

@qiujian16 what do you think about adding the addon-manager into the cluster manager box [like above](https://github.com/open-cluster-management-io/ocm/issues/864#issuecomment-2706677788)? Also, we have 2 operators that did not appear in the diagram, should we add them as well?

gitatractivo:
Hi does this commit look good? If you want I can add the operators as well

qiujian16:
> what do you think about adding the addon-manager into the cluster manager box https://github.com/open-cluster-management-io/ocm/issues/864#issuecomment-2706677788? Also, we have 2 operators that did not appear in the diagram, should we add them as well?

yeah I think we should

gitatractivo:
Hi, could you specify which two operators needed to be added? I'll add them and make a pr.

zhujian7:
@gitatractivo 
- cluster-manager operator runs on the hub cluster, watches the ClusterManager CR to install OCM hub components
- klusterlet operator runs on the managed cluster, watches the Klusterlet CR to install OCM agent components

you can find the code [here](https://github.com/open-cluster-management-io/ocm/tree/main/pkg/operator/operators)

PR #875, opened 2025-03-07
:book: docs: Update the OCM arch diagram to be more hub-spoke/pull-model like

mikeshng:
## Summary

docs: Update the OCM arch diagram to be more hub-spoke/pull-model like

## Related issue(s)

Related https://github.com/open-cluster-management-io/ocm/issues/864


mikeshng:
FYI @gitatractivo

PR #894, opened 2025-03-17
:book: docs: Remove the username and password requirement from Argo CD Agent solution

mikeshng:
## Summary
docs: Remove the username and password requirement from Argo CD Agent solution

I have a working POC mTLS Argo CD Agent authentication working. Removing the username and password requirement from the Argo CD Agent solution will make it more secure and easier to use.

https://github.com/open-cluster-management-io/addon-contrib/pull/26 will need to be merged first.

/hold


mikeshng:
/unhold 

https://github.com/open-cluster-management-io/addon-contrib/pull/26 is now merged

PR #909, opened 2025-03-25
🌱 Code refactor on registration driver

qiujian16:
1. Move driver's options to each drivers implemenation
2. Add BuildClients interface so it is possible to support other driver later.
3. create a general option to simplify driver init
4. remove option with any type

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/register/common.go
+++ b/pkg/registration/register/common.go
@@ -5,17 +5,26 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/klog/v2"
 
+	addonclient "open-cluster-management.io/api/client/addon/clientset/versioned"
+	addoninformers "open-cluster-management.io/api/client/addon/informers/externalversions"
+	clusterv1client "open-cluster-management.io/api/client/cluster/clientset/versioned"
 	hubclusterclientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	clusterv1informers "open-cluster-management.io/api/client/cluster/informers/externalversions"
 	clusterv1listers "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	"open-cluster-management.io/sdk-go/pkg/patcher"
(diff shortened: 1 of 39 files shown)


jaswalkiranavtar:
Builds are failing, other than that refactoring looks good to me

pkg/registration/spoke/spokeagent.go
```diff
@@ -226,33 +206,31 @@ func (o *SpokeAgentConfig) RunSpokeAgentWithSpokeInformers(ctx context.Context,
...
+	secretOption := register.SecretOption{
+		SecretNamespace:          o.agentOptions.ComponentNamespace,
+		SecretName:               o.registrationOption.HubKubeconfigSecret,
+		ClusterName:              o.agentOptions.SpokeClusterName,
+		AgentName:                o.agentOptions.AgentID,
+		ManagementSecretInformer: namespacedManagementKubeInformerFactory.Core().V1().Secrets().Informer(),
+		ManagementCoreClient:     managementKubeClient.CoreV1(),
+		HubKubeconfigFile:        o.agentOptions.HubKubeconfigFile,
+		HubKubeconfigDir:         o.agentOptions.HubKubeconfigDir,
+		BootStrapKubeConfig:      kubeconfig,
```
skeeey, line 224: could we use `o.currentBootstrapKubeConfig` instead of loading the kubeconfig here?

pkg/registration/register/factory/options.go
```diff
@@ -0,0 +1,51 @@
...
+	fs.StringVar(&s.RegistrationAuth, "registration-auth", s.RegistrationAuth,
+		"The type of authentication to use to authenticate with hub.")
+	s.CSROption.AddFlags(fs)
+	s.AWSISRAOption.AddFlags(fs)
+}
+
+func (s *Options) Validate() error {
+	switch s.RegistrationAuth {
+	case helpers.AwsIrsaAuthType:
+		fmt.Println("go to aws driver")
```
skeeey, line 37: remove debug info

pkg/registration/register/common.go
```diff
@@ -227,3 +236,64 @@ func (a *AggregatedHubDriver) Cleanup(ctx context.Context, cluster *clusterv1.Ma
...
+	AddonClient           addonclient.Interface
+	ClusterInfomerFactory clusterv1informers.SharedInformerFactory
+	KubeInformerFactory   informers.SharedInformerFactory
+	AddonInformerFactory  addoninformers.SharedInformerFactory
+}
+
+func BuildClientsFromSecretOption(s SecretOption, bootstrapped bool) (*Clients, error) {
+	var kubeConfig *rest.Config
+	var err error
+	if bootstrapped {
```
skeeey, line 253: may rename the `bootstrapped` to `bootstrap`, the `bootstrapped` may have confusions here

pkg/registration/spoke/spokeagent.go
qiujian16: this is loaded from currentBootstrapKubeConfig

skeeey: I mean we put the BootstrapKubeConfig file path in the SecretOption, not a kubeconfig object, we load the kubeconfig object in the BuildClientsFromSecretOption
in feature, if we don't use kube bootstrap, we can handle this in one place, does it make sense?

qiujian16: I see, yeah we could do that.

skeeey: OK, got it

skeeey:
LGTM

Issue #842, opened 2025-02-20
Stop work agent to apply resources under certain condition defined in the spec.


qiujian16:
**Describe the enhancement**
For certain resource, the user wants the manfiestwork to not apply resources anymore upon certain conditions.
e.g. user creates a job with the manifestwork, and the manifestwork should not create or update job anymore after the job is completed.

mikeshng:
FYI @bhperry

qiujian16:
I believe this feature has been completed.

PR #910, opened 2025-03-26
✨ Workload conditions

bhperry:
## Summary

Support for evaluating manifestwork condition rules

Related PR: https://github.com/open-cluster-management-io/api/pull/362

## Related issue(s)

https://github.com/open-cluster-management-io/enhancements/pull/139
https://github.com/open-cluster-management-io/ocm/issues/842


pkg/work/spoke/conditions/reader.go
```diff
@@ -0,0 +1,211 @@
...
+		ext.Strings(),
+		library.Lists(),
+		library.Regex(),
+		library.URLs(),
+		library.Quantity(),
+		library.IP(),
+		library.CIDR(),
+		library.Format(),
+		// TODO
+		// library.SemverLib(),
```
bhperry, line 180: @haoqing0110 Copied these over from your [CEL selector PR](https://github.com/open-cluster-management-io/ocm/pull/693). Might be a good idea to have a shared helper that establishes a common set of cel envs used by all OCM components, and then add extra envs specific to the use case on top of that

haoqing0110: This is a draft PR for some cel library and common things https://github.com/open-cluster-management-io/sdk-go/pull/105 .

bhperry: perfect!

pkg/work/spoke/conditions/reader.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	return "Manifest is not " + rule.Condition, nil
+}
+
+func (s *ConditionReader) celEnv(opts ...cel.EnvOption) (*cel.Env, error) {
+	opts = slices.Concat(
+		opts,
+		common.BaseEnvOpts,
+		[]cel.EnvOption{
+			cel.Function("hasConditions", cel.Overload(
```
haoqing0110, line 184: Should this be put into sdk-go library? Some description of the usage and an example may help reader better unstand how this function works.

haoqing0110:
FYI PRs we add cost budget, runtime cost estimator for placement.
https://github.com/open-cluster-management-io/sdk-go/pull/114
https://github.com/open-cluster-management-io/ocm/pull/964

pkg/work/spoke/conditions/reader.go
bhperry: Do you mean the hasConditions function? Yes I'd be happy to add it to sdk-go.  I found checking for the existence of status conditions to be rather verbose in raw CEL, so it could certainly be useful elsewhere.

pkg/work/helper/helpers.go
```diff
@@ -387,7 +390,7 @@ func FindManifestConfiguration(resourceMeta workapiv1.ManifestResourceMeta,
 		}
 	}
 
-	if len(rstOption.FeedbackRules) == 0 && rstOption.UpdateStrategy == nil {
+	if len(rstOption.FeedbackRules) == 0 && len(rstOption.ConditionRules) == 0 && rstOption.UpdateStrategy == nil {
```
qiujian16, line 393: you might want to add some ut.

pkg/work/spoke/conditions/reader.go
```diff
@@ -0,0 +1,281 @@
...
+		)
+		if err != nil {
+			return metav1.ConditionUnknown, workapiv1.ConditionRuleInternalError, remainingBudget, err
+		}
+
+		out, newBudget, err := evaluate(ctx, prg, remainingBudget, expression, map[string]any{
+			"object": obj.Object,
+		})
+		if err != nil {
+			return metav1.ConditionFalse, workapiv1.ConditionRuleExpressionError, remainingBudget, err
```
qiujian16, line 157: could you add some comment on why here is false and line156 return unknown?

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
```diff
@@ -77,6 +80,9 @@ func (m *manifestworkReconciler) reconcile(
 		// Add applied status condition
 		manifestCondition.Conditions = append(manifestCondition.Conditions, buildAppliedStatusCondition(result))
 
+		// Add result conditions
+		manifestCondition.Conditions = append(manifestCondition.Conditions, result.Conditions...)
```
qiujian16: I think this reader should be used in status controller, the thing is status controller will return faster and periodically, and you can use status update interval to control how frequent the resource will be read.

qiujian16 (commented):
Could you add some integration test also?

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
bhperry: Hmm ok I'll take a look at that

pkg/work/spoke/conditions/reader.go
bhperry: It's a bit arbitrary. I returned unknown on line 156 because creating the program is not in any way controlled by the user that wrote the CEL expression. Lines 146 and 163 return errors with condition false because they relate to errors in the CEL expression itself, so the user is in control. In other words, I chose to represent a bad rule as false rather than unknown and reserve unknown for cases where the user has no control.

bhperry: Realizing it may make more sense to just always set status Unknown when there is an error though. What do you think?

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
bhperry: That was definitely the right call. Integration test wouldn't even pass until I moved it to status controller because the job status update doesn't trigger reconcile on manifestwork.

pkg/work/spoke/conditions/reader.go
bhperry: The reason I originally decided to treat user errors at status False was because the conditions are always evaluated on unstructured.Unstructured objects. So fields defined with omitempty will be missing when the user might expect them to always exist with their default value.

e.g.
CRD with status defined as 
```
type CustomObjectStatus struct {
    ...
    Something int64 `json:"something,omitempty"`
}
```

Condition Expression: `object.status.something > 0`

Results in:
```
- Type: "MyCondition"
  Status: "False"
  Reason: "ConditionRuleExpressionError"
  Message: "no such key: something"
```

qiujian16:
I love this! thanks!
/approve
leave to @haoqing0110 for another look.

PR #932, opened 2025-04-01
✨ Support configuring resource requirements for addon agents

zhujian7:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/addon/templateagent/decorator.go
+++ b/pkg/addon/templateagent/decorator.go
@@ -183,7 +183,8 @@ func (d *daemonSetDecorator) decorate(obj *unstructured.Unstructured) (*unstruct
 
 type podTemplateSpecDecorator interface {
 	// decorate modifies the pod template in place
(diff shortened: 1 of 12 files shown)


pkg/addon/templateagent/decorator.go
```diff
@@ -508,6 +512,60 @@ func (d *caBundleDecorator) decorate(pod *corev1.PodTemplateSpec) error {
...
+	if !ok {
+		return nil
+	}
+
+	regexRequirements, ok := requirements.([]addonfactory.RegexResourceRequirements)
+	if !ok {
+		return fmt.Errorf("resource requirements value is invalid")
+	}
+
+	for i := range pod.Spec.Containers {
```
elgnay, line 547: Should we consider `pod.Spec.InitContainers` as well?

zhujian7: I am not quite sure, all other decorators do not consider the init containers.

zhujian7:
@qiujian16 PTAL

PR #939, opened 2025-04-07
:bug: fix incorrect agent namespace in klusterlet helm chart

zhiweiyin318:
## Summary
1. correct the definition of  agentNamespace and differentiate between agentNamespace and klusterletNamespace:
agentNamespace is the klusterlet name in hosted mode.
agentNamespace is the klusterlet.namespace in default mode.
agentNamespace is open-cluster-management-agent if klusterlet.namespace is empty in default mode.

2. separate hosted mode from `noOperator = true` because noOperator = true is a special case,  could be hosted mode or default mode.

## Related issue(s)

Fixes #

### DIFF
--- a/deploy/klusterlet/chart/klusterlet/templates/cluster_role.yaml
+++ b/deploy/klusterlet/chart/klusterlet/templates/cluster_role.yaml
@@ -1,3 +1,4 @@
+{{- if or ( eq .Values.klusterlet.mode "Singleton") (eq .Values.klusterlet.mode "Default") }}
 {{- if not .Values.noOperator }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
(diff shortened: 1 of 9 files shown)


deploy/klusterlet/chart/klusterlet/templates/cluster_role.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if or ( eq .Values.klusterlet.mode "Singleton") (eq .Values.klusterlet.mode "Default") ( eq .Values.klusterlet.mode "") }}
```
qiujian16, line 1: can mode be an empty string

zhiweiyin318: checked that there is a default value for mode in values.yaml. 
removed the empty check here.

PR #942, opened 2025-04-08
 ⚠️ Remove resource relating to k8s v1.11

qiujian16:
1. remove v1beta1 crds
2. remove static rbac files

## Related issue(s)

Fixes #


zhujian7:
@qiujian16 I am not sure if we should set ⚠️ as the title prefix.

qiujian16:
updated

Issue #931, opened 2025-04-01
observability in OCM


clyang82:
**Describe the enhancement**
This issue aims to talk about the observability in OCM. We did the experimental about how to leverage addon manager and opentelemetry collector for tracing purpose. refer to https://github.com/open-cluster-management-io/addon-contrib/tree/main/open-telemetry-addon. Since it is 3 years ago, it is out of date.

Today, we receive some feedback from the community about can we collect some basic node metrics of the managed clusters and can we collect the AI workload metrics from the managed clusters. the opentelemetry collector is still suitable for this case. You know the opentelemetry collector supports:
- Metrics (like Prometheus)
- Traces (for distributed tracing)
- Logs (for logging pipelines)
- Correlation between metrics, traces, and logs

In order to address the requirement, the proposal is leverage the addon template and opentelementry collector to collect the metrics.
- [Addon template](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/82-addon-template) - simplify the development of an add-on. it makes developers no longer need to compile and deploy an addon manager for each add-on
- [Opentelemetry collector](https://github.com/open-telemetry/opentelemetry-collector-contrib) - use contrib so that we can have as the Jaeger and Prometheus components in.
- [Node exporter](https://github.com/prometheus/node_exporter) - deploy it to the managed cluster so that we can have some basic node metrics

<img width="362" alt="Image" src="https://github.com/user-attachments/assets/46eb6f10-c15e-4216-87bc-210c0e08943e" />

the idea is:
1. deploy an opentelemetry collector in the hub cluster. it can be configured to export the metrics to different receives. i.e.: grafana, prometheus
2. use addon-template to deploy the opentelemetry-collector and node exporter. the otel config can point to the node exporter metrics server. for example:
```
receivers:
  # Collect own metrics
  prometheus:
    config:
      scrape_configs:
      - job_name: 'node'
        scrape_interval: 10s
        static_configs:
        - targets: ['node-exporter.open-cluster-management-addon:9100']
```
the otel config can also point to the opentelemery collector in the hub cluster. for example:
```
exporters:
  otlp:
    endpoint: otelcol2:4317
    tls:
      cert_file: file.cert
      key_file: file.key
  otlp/2:
    endpoint: otelcol2:4317
    tls:
      insecure: true
```

FYI @qiujian16

qiujian16:
@clyang82 I think this has been completed? Do we still need a document in the website for this?

[...]

PR #975, opened 2025-04-30
:sparkles: feature Add a new option to enable otel roundtrip tracing

dinhxuanvu:
## Summary

Currently, there is no option to enable otel tracing for ocm. Adding EnableOtel option to allow users to config the controller to support otel tracing by wrapping http.RoundTripper with one that starts a span, injects the span context into the outbound request headers, and enriches it with metrics.

## Related issue(s)

Fixes #


dinhxuanvu:
@frzifus FYI

qiujian16:
cc @clyang82 @elgnay

dinhxuanvu:
Related to the Observability enhancement (Issues #931)

Issue #993, opened 2025-05-15
The e2e test passed inconsistently


zhiweiyin318:
**Describe the bug**
There are flaky e2e test cases in addonmanagement_test.go .

**Additional context**
```
Enable addon management feature gate Template type addon should be configured by addon deployment config for namespace [addon-manager]
/home/runner/work/ocm/ocm/test/e2e/addonmanagement_test.go:539
  STEP: create addon custom sign secret namespace @ 05/15/25 02:52:42.714
  STEP: create addon custom sign secret @ 05/15/25 02:52:42.717
  STEP: create addon template resources for cluster e2e-universal-managedcluster @ 05/15/25 02:52:42.733
  STEP: create the addon hello-template on the managed cluster namespace e2e-universal-managedcluster @ 05/15/25 02:52:42.748
  STEP: wait the addon e2e-universal-managedcluster/hello-template available condition to be true @ 05/15/25 02:52:42.753
  [FAILED] in [BeforeEach] - /home/runner/work/ocm/ocm/vendor/github.com/onsi/gomega/internal/async_assertion.go:145 @ 05/15/25 02:54:12.754
  STEP: delete the addon hello-template on the managed cluster namespace e2e-universal-managedcluster @ 05/15/25 02:54:12.754
  STEP: delete addon template resources for cluster e2e-universal-managedcluster @ 05/15/25 02:55:22.829
  STEP: delete addon custom sign secret @ 05/15/25 02:55:22.873
  STEP: delete addon custom sign secret namespace @ 05/15/25 02:55:22.877
• [FAILED] [160.185 seconds]
Enable addon management feature gate [BeforeEach] Template type addon should be configured by addon deployment config for namespace [addon-manager]
  [BeforeEach] /home/runner/work/ocm/ocm/test/e2e/addonmanagement_test.go:95
  [It] /home/runner/work/ocm/ocm/test/e2e/addonmanagement_test.go:539

  [FAILED] Timed out after 90.000s.
  Expected success, but got an error:
      <*errors.errorString | 0xc000d1e850>: 
[...]

Issue #878, opened 2025-03-10
flaky e2e test when cleanup the cluster


qiujian16:
**Describe the bug**
The happens occasionally

  [FAILED] in [AfterSuite] - /home/runner/work/ocm/ocm/vendor/github.com/onsi/gomega/internal/async_assertion.go:145 @ 03/09/25 05:06:08.161
[AfterSuite] [FAILED] [95.018 seconds]
[AfterSuite] 
/home/runner/work/ocm/ocm/test/e2e/e2e_suite_test.go:193

  [FAILED] Timed out after 90.003s.
  Expected success, but got an error:
      <*errors.errorString | 0xc000cf9220>: 
      managed cluster e2e-universal-managedcluster still exists
      {
          s: "managed cluster e2e-universal-managedcluster still exists",
      }
  In [AfterSuite] at: /home/runner/work/ocm/ocm/vendor/github.com/onsi/gomega/internal/async_assertion.go:145

an example is here
https://github.com/open-cluster-management-io/ocm/actions/runs/13744813008/job/38438437176?pr=876

**To Reproduce**
run e2e in github action

PR #995, opened 2025-05-16
:bug: fix flaky e2e test case

zhiweiyin318:
## Summary
the manifestwork of addon is left over after addon is deleted since cluster is deleted before the manifestworks are gone. 
the left manifestwork is blocking the cluster deletion. 

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/993
https://github.com/open-cluster-management-io/ocm/issues/878

### DIFF
--- a/test/e2e/addonmanagement_test.go
+++ b/test/e2e/addonmanagement_test.go
@@ -5,6 +5,7 @@ import (
 	"embed"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	ginkgo "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -191,6 +191,51 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
+	By("clean the resources in the cluster ns before delete the cluster after the test case")
+	err := hub.AddonClient.AddonV1alpha1().ManagedClusterAddOns(universalClusterName).
```
qiujian16, line 195: shouldn't we have the gc to clean all these resources?

Issue #988, opened 2025-05-08
ManifestWorkReplicaSet deleted before its associated ManifestWorks are fully deleted


ahmad-ibra:
**Describe the bug**
When deleting a `ManifestWorkReplicaSet`, it ends up getting cleaned up from the hub cluster before all of the `ManifestWorks` associated with it are deleted.

**To Reproduce**
1. Apply a `ManifestWorkReplicaSet` and wait for a `ManifestWork` to be created for each spoke
2. Delete the `ManifestWorkReplicaSet`
3. Observe the `ManifestWorkReplicaSet` being deleted 
4. Observe the the `ManifestWorks` being deleted some time after step 3 

**Expected behavior**
`ManifestWorks` are all deleted before the `ManifestWorkReplicaSet` is deleted

**Environment ie: OCM version, Kubernetes version and provider:**
OCM Version: v0.16.0

**Additional context**
Here's a video showing the creation and deletion of a `ManifestWorkReplicaSet` along with the `ManifestWorks` that get created.

https://github.com/user-attachments/assets/90970450-16fe-4a2b-91f0-e3d06031f913

The reason this is an issue for my use case is that we have a CR that creates a `ManifestWorkReplicaSet` and retrieves status information about the spoke from the `ManifestWorks`. If we delete our CR, and re-apply it before the `ManifestWorks` are actually deleted, it will report the wrong status from the spoke cluster.

If there is appetite to get this fixed, I'd be happy to look at implementing the change. I believe the issue is caused because the WorkApplier doesn't have a `foreground` propagation policy configured:
https://github.com/open-cluster-management-io/sdk-go/blob/062f078a839599698bea3e7e93c9010bdf187389/pkg/apis/work/v1/applier/workapplier.go#L37

mikeshng:
CC  @elgnay @zhiweiyin318

zhiweiyin318:
There is a finalizer `work.open-cluster-management.io/manifest-work-cleanup` on the ManifestWorkReplicaSet.
This finalizer is removed after delete all related ManifestWorks (not wait the ManifestWorks are gone).
@ahmad-ibra  your exception is that the ManifestWorkReplicaSet should be gone after all related ManifestWorks are gone, right ?

ahmad-ibra:
@zhiweiyin318 yes exactly. i was expecting the ManifestWorkReplicaSet to only have its finalizers removed after the ManifestWorks are all gone.

zhiweiyin318:
[...]

PR #996, opened 2025-05-16
:sparkles: add deletionPolicy for manifestworkReplicaset

zhiweiyin318:
## Summary
 Ddd a filed CascadeDeletionPolicy to configure the deletion policy for the manifestWorkReplicaset.
Defined 2 deletion policies: Background and Foreground.

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/988


qiujian16:
looks good in general, but I think we need to add a integration test.

PR #964, opened 2025-04-23
:sparkles: add cost budget, runtime cost estimator and metrics

haoqing0110:
## Related issue(s)

https://github.com/open-cluster-management-io/ocm/issues/843
https://github.com/open-cluster-management-io/ocm/issues/955

Per expression cost limit exceeded
```
I0424 09:55:58.810235       1 cel.go:175] "Expression evaluation failed" rule="managedCluster.metadata.labels[\"test\"] == \"true\"" cluster="cluster1" err="operation cancelled: actual cost limit exceeded"
```
Total expression cost budget exceeded
```
I0424 09:21:39.903577       1 cel.go:206] "Cost budget exceeded" rule="managedCluster.metadata.labels[\"test\"] == \"true\"" cost=5 budget=0
```
Metrics
```
# HELP scheduling_cel_runtime_duration_seconds [ALPHA] How long in seconds CEL expressions validation runs for a placement.
# TYPE scheduling_cel_runtime_duration_seconds histogram
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="1e-06"} 0
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="9.999999999999999e-06"} 0
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="9.999999999999999e-05"} 3
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="0.001"} 6
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="0.01"} 8
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="0.1"} 8
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="1"} 8
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="10"} 8
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="100"} 8
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="1000"} 8
scheduling_cel_runtime_duration_seconds_bucket{name="placement_scheduling",le="+Inf"} 8
scheduling_cel_runtime_duration_seconds_sum{name="placement_scheduling"} 0.0037694679999999998
scheduling_cel_runtime_duration_seconds_count{name="placement_scheduling"} 8
```

PR #1003, opened 2025-05-21
🌱 Common CEL evaluation functions

bhperry:
## Summary
Extracting CEL evaluation/budgeting functions from the placement CELSelector into the common package so that they can be used for CEL eval across OCM. Need pretty much the same thing for #910, so may as well have generic functions that can be used anywhere.

Could also pull these out into the sdk-go cel package if preferred. Thought I'd start with this and see what others think.

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/pull/964
https://github.com/open-cluster-management-io/ocm/pull/910

### DIFF
--- a/pkg/common/helpers/cel.go
+++ b/pkg/common/helpers/cel.go
@@ -1,0 +1,71 @@
+package helpers
+
+import (
+	"context"
+	"math"
+
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types/ref"
+	"k8s.io/klog/v2"
+)
+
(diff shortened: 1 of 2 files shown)


haoqing0110:
/lgtm

Not sure if putting this part to sdk-go cel package will make users confused about the usage? since this common thing is only part of Validate(), and does not include Compile() as well.

bhperry:
Anyone know what's going on with the e2e failure? Doesn't seem to be related to the code changes at a glance.

haoqing0110:
@bhperry maybe re-run one more time?

PR #889, opened 2025-03-12
:rocket: Added support for about-api for cluster properties

gnana997:
## Summary
Added support for sigs.k8s.io/about-api for the cluster properties.
## Related issue(s)
#838 
Fixes #
Added sigs.k8s.io/about-api for fetching the cluster-properties and have precedence over cluster-claims for the required reserved claim names.

PR #1006, opened 2025-05-26
✨ Added support for about-api for cluster properties

qiujian16:
## Summary

original PR and add more tests: https://github.com/open-cluster-management-io/ocm/pull/889

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/registration/spoke/managedcluster/claim_reconcile.go
+++ b/pkg/registration/spoke/managedcluster/claim_reconcile.go
@@ -3,21 +3,22 @@ package managedcluster
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strings"
+
 	"github.com/openshift/library-go/pkg/operator/events"
 	"k8s.io/apimachinery/pkg/api/meta"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/sets"
+	aboutv1alpha1listers "sigs.k8s.io/about-api/pkg/generated/listers/apis/v1alpha1"
+
 	clusterv1alpha1listers "open-cluster-management.io/api/client/cluster/listers/cluster/v1alpha1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	clusterv1alpha1 "open-cluster-management.io/api/cluster/v1alpha1"
 	ocmfeature "open-cluster-management.io/api/feature"
+
 	"open-cluster-management.io/ocm/pkg/features"
-	aboutv1alpha1 "sigs.k8s.io/about-api/pkg/apis/v1alpha1"
-	aboutv1alpha1listers "sigs.k8s.io/about-api/pkg/generated/listers/apis/v1alpha1"
-	"sort"
-	"strings"
 )
 
 const labelCustomizedOnly = "open-cluster-management.io/spoke-only"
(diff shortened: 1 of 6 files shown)


pkg/registration/spoke/managedcluster/claim_reconcile.go
```diff
@@ -46,31 +47,70 @@ func (r *claimReconcile) reconcile(ctx context.Context, cluster *clusterv1.Manag
...
+		}
 	}
 
 	// check if the cluster claim is one of the reserved claims or has a reserved suffix.
 	// if so, it will be treated as a reserved claim and will always be exposed.
 	reservedClaimNames := sets.New(clusterv1alpha1.ReservedClusterClaimNames[:]...)
 	reservedClaimSuffixes := sets.New(r.reservedClusterClaimSuffixes...)
+
+	// when ClusterProperties feature is not enabled, the informer will not be started and the lister will
+	// return an empty list, so we do not need to check featuregate here.
```
zhujian7, line 74: The comment says "we do not need to check featuregate here", but the code does check it. correct the comment?

pkg/registration/spoke/managedcluster/claim_reconcile.go
```diff
@@ -46,31 +47,70 @@ func (r *claimReconcile) reconcile(ctx context.Context, cluster *clusterv1.Manag
...
+		if err != nil {
+			return fmt.Errorf("unable to list cluster properties: %w", err)
+		}
+
+		for _, property := range clusterProperties {
+			propertiesMap[property.Name] = property
+		}
+	}
+
+	// convert claim to properties
```
zhujian7, line 87: here is a little bit weird that we convert the claim to clusterProperty in line 93, and again we convert the clusterProperty to claim in line 104.

We can declare a `claimsMap := map[string]clusterv1.ManagedClusterClaim{}` at the beginning of the func, and for range the listed clusterclaims and clusterProperties, then covert both of them to clusterv1.ManagedClusterClaim and insert. finally we for range  the claimsMap directly to do the reserved claim matching.

pkg/registration/spoke/managedcluster/claim_reconcile.go
```diff
@@ -46,31 +47,70 @@ func (r *claimReconcile) reconcile(ctx context.Context, cluster *clusterv1.Manag
...
+	}
 
+	for _, property := range propertiesMap {
+		managedClusterClaim := clusterv1.ManagedClusterClaim{
+			Name:  property.Name,
+			Value: property.Spec.Value,
+		}
 		if matchReservedClaims(reservedClaimNames, reservedClaimSuffixes, managedClusterClaim) {
 			reservedClaims = append(reservedClaims, managedClusterClaim)
+			reservedClaimNames.Insert(property.Name)
```
zhujian7, line 110: why do we need to change the 4 reservedClaimNames?

pkg/registration/spoke/managedcluster/claim_reconcile.go
qiujian16: clusterv1.ManagedClusterClaim  is different from clusterv1alpha1.ClusterClaim. the former is the struct in managedcluster. The idea here is all clusterv1alpha1.ClusterClaim are converted to the properties, then we construct clusterv1.ManagedClusterClaim based on properties

pkg/registration/spoke/managedcluster/claim_reconcile.go
qiujian16: fixed

pkg/registration/spoke/managedcluster/claim_reconcile.go
qiujian16: make sense, updated.

zhujian7:
lgtm, leave @elgnay to take a final review.

PR #1015, opened 2025-05-29
:bug:  check works in the addon management e2e case

zhiweiyin318:
## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/addonmanagement_test.go
+++ b/test/e2e/addonmanagement_test.go
@@ -42,7 +42,7 @@ const (
 	originalImageValue                       = "quay.io/open-cluster-management/addon-examples:latest"
 	overrideImageValue                       = "quay.io/ocm/addon-examples:latest"
 	customSignerName                         = "example.com/signer-name"
-	//#nosec G101
+	// #nosec G101
 	customSignerSecretName = "addon-signer-secret"
 )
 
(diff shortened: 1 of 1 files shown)


test/e2e/addonmanagement_test.go
```diff
@@ -163,6 +163,22 @@ var _ = ginkgo.Describe("Enable addon management feature gate", ginkgo.Ordered,
 			return fmt.Errorf("the managedClusterAddon %s should be deleted", addOnName)
 		}).ShouldNot(gomega.HaveOccurred())
 
+		gomega.Eventually(func() error {
```
zhujian7: addon was deleted, there should not be any work here, if we want to check if any work exists, I recommend not using Eventually.

zhiweiyin318: ok. done.

test/e2e/addonmanagement_test.go
```diff
@@ -153,14 +153,24 @@ var _ = ginkgo.Describe("Enable addon management feature gate", ginkgo.Ordered,
...
 				context.TODO(), addOnName, metav1.GetOptions{})
+			if err == nil {
+				return fmt.Errorf("the managedClusterAddon %s should be deleted", addOnName)
+			}
+			if err != nil && !errors.IsNotFound(err) {
+				return err
+			}
+
+			// check works after addon is not found
+			works, err := hub.WorkClient.WorkV1().ManifestWorks(universalClusterName).List(
```
zhujian7, line 164: I am thinking moving this check out of the eventually, to line 175, so if the addon is deleted, but there are works left, we make the test fail. and the print message will show which work is left.

zhiweiyin318: done

Issue #957, opened 2025-04-17
Apply all labels that Cluster Manager has to all resources it creates and keep labels synced


jeffw17:
**Describe the enhancement**
Similar to this [issue](https://github.com/open-cluster-management-io/ocm/issues/467), the resources created by the Cluster Manager does not have any of the labels that the Cluster Manager has. As discussed with @qiujian16 and @mikeshng, we agree that we can create an enhancement to apply all labels to resources created by the Cluster Manager and they will also be synced if any labels has been changed.

qiujian16:
@jeffw17 I think we have completed this?

jeffw17:
@qiujian16 Hi Jian Qiu, we've only implemented the label sync for resources created by operator and registration controller. We still have the other controllers to implement for (e.g. work, placement, etc.). We have this as part of our backlog and will work on this in the near future. I think for now we can keep this open a bit longer. When's the next release for OCM? If its upcoming, we'll try to prioritize it for the release.

qiujian16:
Let's target it for end of September.

PR #1021, opened 2025-06-02
✨ Enable sync labels for clustermanager operator and registration controller 

jeffw17:
## Summary

Previously, there was a feature implemented where the klusterlet resources have labels passed down to resources created indirectly by the controllers. This feature was not implemented in the clustermanager side. For example, the operator goes ahead and create the various deployments on the open-cluster-management-hub namespace and those deployments will now have the same labels as the operator. As such, the changes in this PR implements this feature. The feature in this PR however only applies to the cluster-manager operator and the registration-controller. In a future date, we will implement the same feature for the other controllers (i.e. placement-controller, work-controller, etc.). To enable this feature, in the helm charts, you can set enableSyncLabels to true or, in clusteradm commands, you can add the --enable-sync-labels=true flag.

## Related issue(s)
https://github.com/open-cluster-management-io/ocm/issues/957

Fixes # [957](https://github.com/open-cluster-management-io/ocm/issues/957)


pkg/registration/helpers/helpers.go
```diff
@@ -176,3 +179,21 @@ func IsCSRSupported(nativeClient kubernetes.Interface) (bool, bool, error) {
 	}
 	return v1CSRSupported, v1beta1CSRSupported, nil
 }
+
+func ParseLabels(labels string) map[string]string {
```
qiujian16: you could use this https://github.com/kubernetes/kubectl/blob/master/pkg/generate/generate.go#L178

pkg/registration/hub/manager.go
```diff
@@ -206,6 +209,7 @@ func (m *HubManagerOptions) RunControllerManagerWithInformers(
 		workInformers.Work().V1().ManifestWorks(),
 		hubDriver,
 		controllerContext.EventRecorder,
+		m.Labels,
```
qiujian16: it is better to parse the Label in the Complete phase, so it can fail fast if it is set incorrectly.

pkg/operator/helpers/helpers.go
```diff
@@ -831,6 +834,23 @@ func GetKlusterletAgentLabels(klusterlet *operatorapiv1.Klusterlet) map[string]s
 	return labels
 }
 
+func GetClusterManagerLabels(cm *operatorapiv1.ClusterManager) map[string]string {
```
qiujian16: is this necessary? Seems to me only a check on whether label is set.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -222,6 +225,11 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 		config.WorkWebhook = convertWebhookConfiguration(clusterManager.Spec.DeployOption.Hosted.WorkWebhookConfiguration)
 	}
 
+	if n.enableSyncLabels {
+		config.LabelsString = helpers.ConvertLabelsMapToString(helpers.GetClusterManagerLabels(clusterManager))
+		config.Labels = helpers.GetClusterManagerLabels(clusterManager)
```
qiujian16: could you just do helpers.GetClusterManagerLabels(clusterManager.Labels)?

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
```diff
@@ -136,6 +136,11 @@ func (c *runtimeReconcile) reconcile(ctx context.Context, cm *operatorapiv1.Clus
 		}
 	}
 
+	if cm.Labels != nil {
```
qiujian16: why do we need this again, I think config has been set on the controller already.

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -195,10 +196,17 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 	namespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: managedClusterName,
-			Labels: map[string]string{
-				v1.ClusterNameLabelKey: managedClusterName,
-			},
-		},
+			Labels: func() map[string]string {
+				labels := map[string]string{
```
qiujian16:
```go
labels := ParseLabels(c.labels)
labels[v1.ClusterNameLabelKey] = managedClusterName
```
This is cleaner.

pkg/registration/hub/manager.go
```diff
@@ -89,6 +90,8 @@ func (m *HubManagerOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringSliceVar(&m.AutoApprovedARNPatterns, "auto-approved-arn-patterns", m.AutoApprovedARNPatterns,
 		"A list of AWS EKS ARN patterns such that an EKS cluster will be auto approved if its ARN matches with any of the patterns")
 	fs.StringSliceVar(&m.AwsResourceTags, "aws-resource-tags", m.AwsResourceTags, "A list of tags to apply to AWS resources created through the OCM controllers")
+	fs.StringVar(&m.Labels, "labels", m.Labels,
```
qiujian16, line 94: It is better to set it as fs.StringSliceVar, and convert it to map in complete.

go.mod
```diff
@@ -27,14 +29,15 @@ require (
...
-	k8s.io/api v0.32.4
+	k8s.io/api v0.33.1
 	k8s.io/apiextensions-apiserver v0.32.2
-	k8s.io/apimachinery v0.32.4
+	k8s.io/apimachinery v0.33.1
 	k8s.io/apiserver v0.32.3
-	k8s.io/client-go v0.32.4
-	k8s.io/component-base v0.32.4
+	k8s.io/client-go v0.33.1
+	k8s.io/component-base v0.33.1
```
qiujian16: please do not upgrade the library

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
jeffw17: We want to sync the labels continuously. In case the labels are changed on the clustermanager, the new labels should be reflected.

pkg/registration/hub/manager.go
jeffw17: If we use StringSliceVar which returns m.Labels as a []string then we won't be able to leverage generate.ParseLabels() as it expects a string to convert to map.

pkg/registration/helpers/helpers.go
qiujian16: do we still need this func?

test/integration/registration/integration_suite_test.go
```diff
@@ -37,8 +37,12 @@ import (
 )
 
 const (
-	eventuallyTimeout  = 30 // seconds
-	eventuallyInterval = 1  // seconds
+	eventuallyTimeout     = 30 // seconds
+	eventuallyInterval    = 1  // seconds
+	testCustomLabel       = "custom-label"
```
qiujian16: this should not be set as global variable. It will be really hard to maintain. You should consider stop the existing hub, customize the option and restart the hub in the specific test. Example like this https://github.com/open-cluster-management-io/ocm/blob/main/test/integration/registration/spokecluster_aws_joining_test.go#L32-L47

test/integration/operator/integration_suite_test.go
```diff
@@ -129,6 +129,11 @@ var _ = ginkgo.BeforeSuite(func() {
 	_, err = operatorClient.OperatorV1().ClusterManagers().Create(context.Background(), &operatorapiv1.ClusterManager{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: clusterManagerName,
+			Labels: map[string]string{
+				"app":      "clustermanager",
```
qiujian16: same here, consider a test that update the clustermanager, and check that the label is updated on the resource instead of setting it globally.

pkg/registration/helpers/helpers.go
suvaanshkumar: Removed it

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
qiujian16: hrm I do not think it will happen in reconcile, the cm is passed from the controller and it has been updated already, this will not be changed in the reconcile loop.

test/integration/registration/integration_suite_test.go
```diff
@@ -216,6 +216,10 @@ var _ = ginkgo.BeforeSuite(func() {
 		go func() {
 			defer ginkgo.GinkgoRecover()
 			m.ImportOption.APIServerURL = cfg.Host
+			m.EnabledRegistrationDrivers = []string{helpers.CSRAuthType, helpers.AwsIrsaAuthType}
```
qiujian16: I remember I actually remove this part on purpose, so we do not trigger aws driver in the other integration test to avoid error messages.

test/integration/registration/managedcluster_accept_test.go
```diff
@@ -13,10 +13,32 @@ import (
...
+	var testCustomLabel = "custom-label"
+	var testCustomLabelValue = "custom-value"
+	var testCustomLabel2 = "custom-label2"
+	var testCustomLabelValue2 = "custom-value2"
+
+	BeforeAll(func() {
+		// stop the hub and start new hub with the updated option
+		stopHub()
+
+		awsHubOption := hub.NewHubManagerOptions()
```
qiujian16: it is not awsHubOption, maybe hubOptionWithLabelSetting

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_runtime_reconcile.go
jeffw17: Removed

pkg/registration/hub/managedcluster/controller.go
```diff
@@ -75,8 +76,12 @@ func NewManagedClusterController(
 	clusterRoleBindingInformer rbacv1informers.ClusterRoleBindingInformer,
 	manifestWorkInformer workinformers.ManifestWorkInformer,
 	hubDriver register.HubDriver,
-	recorder events.Recorder) factory.Controller {
+	recorder events.Recorder, labels map[string]string) factory.Controller {
 
+	deepCopyLabels := make(map[string]string, len(labels))
```
qiujian16, line 82: nit add a comment here on why we need to copy.

mikeshng (approved):
/lgtm

re-running the e2e now

PR #1025, opened 2025-06-05
✨ Enable about-api when ClusterProperty featuregate is enabled

qiujian16:
## Related issue(s)

Fixes #


manifests/klusterlet/managed/klusterlet-registration-clusterrole.yaml
```diff
@@ -20,3 +20,7 @@ rules:
 - apiGroups: ["cluster.open-cluster-management.io"]
   resources: ["clusterclaims"]
   verbs: ["get", "list", "watch"]
+  # Allow agent to list clusterproperties
+- apiGroups: ["about.k8s.io"]
```
zhujian7, line 24: I did not see any code change needs these permission, so this is to add permission for the previous merged PR https://github.com/open-cluster-management-io/ocm/pull/1006 , right?

zhujian7:
lgtm

manifests/klusterlet/managed/klusterlet-registration-clusterrole.yaml
qiujian16: yes

PR #1027, opened 2025-06-06
⚠️ use managedclustraddon install namespace when addondeploymentconfig not set

zhujian7:
## Summary

Currently, when deploying an addonTemplate type addon, if the addondeploymentconfig is not set, we will deploy the resources into the namespace of the first deployment/daemonset defined in the addonTemplate. This PR changes to deploy the resources into the namespace of managedclusteraddon.spec.installnamespace.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/addon/templateagent/decorator.go
+++ b/pkg/addon/templateagent/decorator.go
@@ -29,7 +29,7 @@ type namespaceDecorator struct {
 	paths map[string][]string
 }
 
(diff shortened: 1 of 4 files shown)


zhujian7:
/hold
Not sure if this change is necessary as we are going to deprecate the managedclusteraddon.spec.installnamespace

qiujian16:
we should use the namespace in status

PR #1033, opened 2025-06-11
✨ Manifest completion

bhperry:
## Summary

- Ignore updates to manifests that have the Complete condition
- Aggregate the Complete condition from manifests to ManifestWork
  - Any manifest with a Complete condition rule is completable
  - Once all completable manifests within a work are complete, the work is considered complete
- Ignore updates to ManifestWork that has the Complete condition
- Delete ManifestWork that has Complete condition and TTLSecondsAfterFinished delete option if the TTL since completion has been satisfied

## Related issue(s)

https://github.com/open-cluster-management-io/api/pull/378
https://github.com/open-cluster-management-io/enhancements/pull/139
https://github.com/open-cluster-management-io/ocm/issues/842

## Summary by CodeRabbit

* **New Features**
  * Improved aggregation and management of manifest conditions with enhanced condition rule evaluation and pruning.
  * Added support for marking ManifestWork resources as complete based on manifest conditions.
* **Bug Fixes**
  * Prevented unnecessary reconciliation of completed ManifestWork resources.
  * Avoided reapplying manifests already marked as complete.
* **Refactor**
  * Modularized condition evaluation and aggregation for clearer lifecycle management.
  * Updated test setup and validation for better condition handling and extensibility.
* **Tests**
  * Expanded condition rule tests, including aggregation and condition removal scenarios.
  * Centralized condition validation in test utilities for consistent error reporting.
  * Added integration tests for ManifestWork completion based on condition rules.
  * Enhanced ManagedCluster importer tests to wait for namespace creation.
* **Chores**
  * Enabled passing additional runtime arguments to integration tests via Makefile.


bhperry:
Anyone understand why ManifestWork deletion isn't working on `test-cloudevents-integration`?

I see this note about skipped tests on the makefile
> - manifestworkreplicaset_test.go, this test needs to update the work status with the hub work client,
>   cloud events work client does not support it. (TODO) may add e2e to for mwrs.

Maybe related to that? Since deleting work relies on seeing updates to the work status conditions. Should I add an exclusion for these tests to cloudevents?

qiujian16:
> Anyone understand why ManifestWork deletion isn't working on `test-cloudevents-integration`?
> 
> I see this note about skipped tests on the makefile
> 
> > * manifestworkreplicaset_test.go, this test needs to update the work status with the hub work client,
> >   cloud events work client does not support it. (TODO) may add e2e to for mwrs.
> 
> Maybe related to that? Since deleting work relies on seeing updates to the work status conditions. Should I add an exclusion for these tests to cloudevents?

@skeeey thought?

I think the main reason is here https://github.com/open-cluster-management-io/ocm/pull/1033#pullrequestreview-2919300152. The cloudevent client does not have permission to delete the work.

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -208,3 +215,45 @@ func (m *ManifestWorkController) applyAppliedManifestWork(ctx context.Context, w
 	_, err = m.appliedManifestWorkPatcher.PatchSpec(ctx, appliedManifestWork, requiredAppliedWork.Spec, appliedManifestWork.Spec)
 	return appliedManifestWork, err
 }
+
+func (m *ManifestWorkController) manifestWorkCompletion(
+	ctx context.Context, controllerContext factory.SyncContext, manifestWork *workapiv1.ManifestWork,
+) (bool, error) {
+	complete := meta.FindStatusCondition(manifestWork.Status.Conditions, workapiv1.ManifestComplete)
```
qiujian16: why not IsStatusConditionTrue?

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -131,13 +134,31 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
...
+		meta.SetStatusCondition(&manifestWork.Status.Conditions, metav1.Condition{
+			Type:               workapiv1.WorkComplete,
+			Status:             metav1.ConditionTrue,
+			Reason:             workapiv1.WorkManifestsComplete,
+			Message:            "All completion condition rules passed",
+			LastTransitionTime: metav1.Now(),
+			ObservedGeneration: manifestWork.Generation,
+		})
+	} else {
+		meta.RemoveStatusCondition(&manifestWork.Status.Conditions, workapiv1.WorkComplete)
```
qiujian16: what if numCompletable > 0 && numCompletable < numCompleted ? should it return false instead of removal?

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -252,14 +273,29 @@ func (c *AvailableStatusController) getFeedbackValues(
...
+		meta.SetStatusCondition(manifestConditions, condition)
 	}
 
-	return c.conditionReader.EvaluateConditions(ctx, obj, option.ConditionRules)
+	// Remove conditions set by old rules that no longer exist
+	// All conditions set by rules have reason prefixed with "ConditionRule" to easily identify them vs. conditions set elsewhere
+	// Additionally they will always set ObservedGeneration in order to identify a condition set by a rule that no longer exists.
+	// Generation will always update when the ManifestWorkSpec changes, so removing a rule will result in a new generation.
+	*manifestConditions = slices.DeleteFunc(*manifestConditions, func(c metav1.Condition) bool {
+		return strings.HasPrefix(c.Reason, "ConditionRule") && c.ObservedGeneration != generation
```
qiujian16: Why to do need to check the prefix, and run delete? Everytime we reconcile, it should use the updated reason.

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
bhperry: Because it uses last transition time from the condition to determine if TTL has elapsed for deletion

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -208,3 +215,45 @@ func (m *ManifestWorkController) applyAppliedManifestWork(ctx context.Context, w
...
+	if complete != nil && complete.Status == metav1.ConditionTrue {
+		var err error
+		// check if work has TTL set and it has elapsed since completion
+		if ttl, ok := remainingTtl(complete.LastTransitionTime, manifestWork.Spec.DeleteOption); ok {
+			if ttl <= 0 {
+				// Only delete if resourceVersion matches in case complete condition has changed
+				deleteOpts := metav1.DeleteOptions{Preconditions: &metav1.Preconditions{ResourceVersion: &manifestWork.ResourceVersion}}
+				err = m.manifestWorkClient.Delete(ctx, manifestWork.Name, deleteOpts)
```
qiujian16: I do not think manifestwork should be deleted on the spoke side. The agent should not have the permission to do that. Would you remove the deletion logic for now. I think if we really need this, it will be a controller on hub that perform the action after ttl in the future.

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
```diff
@@ -181,6 +183,16 @@ func (m *manifestworkReconciler) applyOneManifest(
 		return result
 	}
 
+	// manifests with the Complete condition are not updated
+	manifestCondition := helper.FindManifestCondition(resMeta, workStatus.ResourceStatus.Manifests)
+	if manifestCondition != nil {
+		complete := meta.FindStatusCondition(manifestCondition.Conditions, workapiv1.ManifestComplete)
```
qiujian16: better to use IsStatusConditionTrue here.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
bhperry: Because conditions aren't replaced each run if the status controller, they are merged into the existing slice. Some conditions are set in manifestcontroller, so it can't just fully reconstruct conditions each run.

Without this, old conditions for rules that are deleted (or have their Condition renamed) never get cleaned up.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
bhperry: Not possible. numCompletable is strictly >= numCompleted.

In the case where numCompleted < numCompletable I chose to remove the Complete condition from top level status rather than set false because that is more like the way Job uses it (the condition doesn't get added to job status until it is True)

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: I see, this is because we are supporting arbitraty type and we do not know which one is from conditionRule?
Could we create a clear func something like IsConditionGeneratedByConditionRule which might be easier to read.
This line also needs more comments on why we need to delete.

go.mod
```diff
@@ -2,6 +2,8 @@ module open-cluster-management.io/ocm
 
 go 1.23.6
 
+replace open-cluster-management.io/api => github.com/bhperry/ocm-api v0.0.0-20250611151546-e02ef2723ba7
```
qiujian16: this needs to be updated.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: if so, why do we need to remove the condition? My understanding is we do not add complete until numCompleted reaches to numCompletable. When will we need to remove the condition? It is like a final state to me.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
bhperry: > I see, this is because we are supporting arbitraty type and we do not know which one is from conditionRule?

Correct

> Could we create a clear func something like IsConditionGeneratedByConditionRule which might be easier to read.
> This line also needs more comments on why we need to delete.

Good idea. Will do.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
bhperry: Could treat it that way. However since Complete is more loosely defined here than in Job (user defined CEL rules could set it any way they want), it is entirely possible that one or more manifest's Complete condition could flip from True to False. I did not see a reason to enforce that it is a final state, rather let users set it however they would like.

If you would prefer to treat it as a final state that cannot be removed I'm not necessarily opposed. If that were the case then would the status controller also stop doing updates once the Complete condition is set? Would it stop syncing status feedback as well?

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
bhperry: It seems like an arbitrary distinction to say that the spoke can't have delete permission. When it already has full update permissions, so it could for instance remove every manifest from the work. Or add Complete conditions and a TTL to make the hub delete it.

qiujian16: I do not think agent has update permisison either, it can only update status, but not the spec.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: It is better to always exclude Applied condition, since it should never be controlled here.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: hrm, If the condition can flip from true to false, shouldn't we set the aggregate complete condition to false instead of removal. And it will not be the same as job's complete status since complete is the end state for job.

I prefer it is not the end state, which means to me:
- if condition rule is defined, always set the aggregated condition and never remove. Its status could be either True or False
- if condition rule is removed from the work spec, the related aggregated condition should be removed.
This is clearer logic to me.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -131,13 +134,31 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
...
-		// Read conditions according to condition rules
-		conditions := c.evaluateConditions(ctx, obj, option)
-		for _, condition := range conditions {
-			meta.SetStatusCondition(manifestConditions, condition)
+		// Update manifest conditions according to condition rules
+		c.evaluateConditionRules(ctx, manifestConditions, obj, option, manifestWork.Generation)
+
+		// Check if manifest is completable/completed
+		if complete := meta.FindStatusCondition(*manifestConditions, workapiv1.ManifestComplete); complete != nil {
+			numCompletable++
```
qiujian16: after another look, and I feel we need more thought on this...we are treating complete condition in a very special way here, while I think we should have general approach for any condition type. It means for each condition type from condition rule, We should check all the manifest that has the condition and aggregate them.

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
bhperry: It has full update permission https://github.com/open-cluster-management-io/ocm/blob/27fc65e174508ae9a2ebe9f84b08f333420dd9f6/pkg/registration/hub/manifests/rbac/managedcluster-work-clusterrole.yaml#L18-L20

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
bhperry: I was trying to avoid a situation where a list of excluded conditions has to be maintained. Easy source for bugs if new conditions are added in manifestworkcontroller at some point and not added to the list of exclusions.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
bhperry: Ok. That does make them much more convenient to use too from a user perspective. Don't need to hunt through manifest status conditions to find them (unless you need granularity), they'll always be aggregated to the top.  I like that.

bhperry: Updated in https://github.com/open-cluster-management-io/ocm/pull/1033/commits/89d8fc8120d80056397071a4e2e01ab7d421a198 to do generic condition aggregation, and included handling for the Available condition as well. Is that what you were looking for?

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
qiujian16: We only use it to update finalizer of the work. Sorry for the confusion, but this is we have to add the permission to update finalizer since there is no finer grained permission rule right now. But we need to avoid update spec or delete manifestwork from the agent side. The cleanup can be fully done on the gc controller on the hub.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: ok, that is fine

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
qiujian16: yes, it looks good

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
bhperry: Removed deletion from spoke for now. Hub GC will be a followup.

go.mod
qiujian16: please remove this line.

pkg/work/spoke/controllers/manifestcontroller/manifestwork_controller.go
```diff
@@ -46,6 +46,7 @@ type workReconcile interface {
 // ManifestWorkController is to reconcile the workload resources
 // fetched from hub cluster on spoke cluster.
 type ManifestWorkController struct {
+	manifestWorkClient         workv1client.ManifestWorkInterface
```
qiujian16: you do not need this client

qiujian16:
/approve
/lgtm
/hold

I think it is ready to merge, hold in case anyone want to take another look.

mikeshng:
CC @elgnay @zhujian7 @zhiweiyin318

Issue #1032, opened 2025-06-11
configurable healthcheck process port in klusterlet


dongbeiqing91:
**Describe the enhancement**

We are trying to use `hostNetwork` in ocm agent registration and work pods, but fail because health check process is using **hardcoded** 8443. If the node has another process using 8443 then the pod starts crashing.

Please make the the health check process port configurable in klusterlet.

o-farag:
@dongbeiqing91 @bhperry may I work on this?

bhperry:
> [@dongbeiqing91](https://github.com/dongbeiqing91) [@bhperry](https://github.com/bhperry) may I work on this?

Fine by me

mikeshng:
Thank you @o-farag !

CC @zhujian7 @zhiweiyin318 @xuezhaojun

dongbeiqing91:
Thanks @o-farag !

shivansh-source:
@mikeshng can i work on this ??

mikeshng:
@shivansh-source yes!

You have been an amazing contributor. Please go ahead!

PR #1035, opened 2025-06-12
✨ Support default mode webhook networking configuration

bhperry:
## Summary

Add support for customization of network config on webhooks running in Default installation mode (i.e. running in hub cluster)

The primary reason to add this is because [EKS with Calico CNI](https://docs.tigera.io/calico/latest/getting-started/kubernetes/managed-public-cloud/eks#install-eks-with-calico-networking) requires that webhooks run with `hostNetwork: true`, otherwise the webhook pods are unreachable from the k8s apiserver.

In order to support the webhooks running in hostNetwork mode, it is important to be able to configure the ports that they bind to to avoid conflicts on nodes. To run in hostNetwork mode, a user might set clustermanager something like this:
```
kind: ClusterManager
...
spec:
  ...
  deployOption:
    default:
      registrationWebhookConfiguration:
        healthProbeBindAddress: :18000
        metricsBindAddress: :18080
        hostNetwork: true
        port: 19443
      workWebhookConfiguration:
        healthProbeBindAddress: 28000
        metricsBindAddress: :28080
        hostNetwork: true
        port: 29443
    mode: Default
```

This ensures the two webhooks will not conflict with each other, or other open ports on the nodes (depending on the particular setup of the cluster of course)

## Related issue(s)

Related to (but does not fix) https://github.com/open-cluster-management-io/ocm/issues/1032


zhiweiyin318:
@bhperry looks like the addon webhook is missing 
https://github.com/open-cluster-management-io/ocm/blob/main/manifests/cluster-manager/hub/addon-manager/webhook-service.yaml 
https://github.com/open-cluster-management-io/ocm/blob/main/manifests/cluster-manager/management/addon-manager/webhook-deployment.yaml

bhperry:
@zhiweiyin318 Yes addon webhook was added after I originally wrote this. The API changes for the first two webhooks are already merged, but there's no configuration for the addon webhook at all yet so we'll need to update that. Was thinking I'd follow up after this PR is merged, but happy to do it all in one if preferred.

qiujian16:
/approve

let's merge this at first, and put addon webhook is the followup. Also I'd like to investigate whether the current webhook can be replaced by validatingadmissionpolicy.

PR #1041, opened 2025-06-16
:bug:  fix no requeue when return requeueError

zhiweiyin318:
## Summary
should requeue when get a requeueError.

## Related issue(s)

Fixes #


qiujian16 (commented):
/approve
/lgtm

PR #1073, opened 2025-07-16
:bug:revert the app label

zhiweiyin318:
## Summary
cannot change the app label in the deployments, since it is used in the spec.selector.matchLabels, will affect the upgrade.
## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **Chores**
  * Standardized deployment labels in cluster manager manifests by replacing dynamic app label values with static values for improved consistency. No changes to deployment functionality or configuration beyond label updates.

### DIFF
--- a/manifests/cluster-manager/management/cluster-manager-registration-deployment.yaml
+++ b/manifests/cluster-manager/management/cluster-manager-registration-deployment.yaml
@@ -14,11 +14,11 @@ spec:
   replicas: {{ .Replica }}
   selector:
     matchLabels:
-      app: {{ .ClusterManagerName }}-registration-controller
+      app: clustermanager-registration-controller
   template:
     metadata:
       labels:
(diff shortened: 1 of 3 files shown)


manifests/cluster-manager/management/cluster-manager-registration-deployment.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: {{ .ClusterManagerName }}-registration-controller
   namespace: {{ .ClusterManagerNamespace }}
   labels:
-    app: {{ .ClusterManagerName }}-registration-controller
+    app: clustermanager-controller
```
qiujian16: I think it should be clustermanager-registration-controller

zhiweiyin318: fixed.

PR #1078, opened 2025-07-21
🌱 Build a single hub manager command

qiujian16:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Added new "hub-manager" and "webhook-server" commands for enhanced hub management and webhook server operations.
  * Introduced centralized options and setup for running the hub manager and webhook server.
  * Improved integration with Kubernetes by supporting advanced controller and webhook configurations.

* **Tests**
  * Added integration tests for the new hub manager functionality to ensure reliability and stability.

### DIFF
--- a/cmd/registration-operator/main.go
+++ b/cmd/registration-operator/main.go
@@ -48,6 +48,7 @@ func newNucleusCommand() *cobra.Command {
 
 	cmd.AddCommand(hub.NewHubOperatorCmd())
 	cmd.AddCommand(hub.NewHubManagerCmd())
+	cmd.AddCommand(hub.NewWebhookCmd())
 	cmd.AddCommand(spoke.NewKlusterletOperatorCmd())
 	cmd.AddCommand(spoke.NewKlusterletAgentCmd())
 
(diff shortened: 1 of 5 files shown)


cmd/registration-operator/main.go
```diff
@@ -47,6 +47,8 @@ func newNucleusCommand() *cobra.Command {
 	}
 
 	cmd.AddCommand(hub.NewHubOperatorCmd())
+	cmd.AddCommand(hub.NewHubManagerCmd())
+	cmd.AddCommand(hub.NewWebhookCmd())
```
bhperry, line 51: What commands are these replacing?

qiujian16: I was thinking to introduce a singleton mode in the clustermanager also, so we do not deploy registration/placement /work controller and webhooks separately, but only two deployment: a single controller and a webhook.

bhperry: Oh gotcha. It would be helpful to have some comments or command descriptions that explain that. As a newcomer to the project, it is often difficult to tell which command/manager is which for spoke singleton vs. separate.

qiujian16: yeah, that is a good suggestion.

PR #1112, opened 2025-08-06
:seedling: [refactor] using dir to reorg cluster-manager manifests

skeeey:
## Related issue(s)

Fixes #


qiujian16 (commented):
should we also put deployment in the corresponding directory?

skeeey:
> should we also put deployment in the corresponding directory?

yes, updated

Issue #1118, opened 2025-08-07
Helm Chart support


softcane:
I'm just referring to this thread. https://github.com/open-cluster-management-io/ocm/issues/1067#issuecomment-3058737606

```custom-ocm-hub-helm(custom helm chart)
->templates
-->cluster-manager-helm-release.yaml(This yaml has the HelmRelease which installs cluster-manager helmchart)
-->rotate-token-cron-job.yaml(This yaml creates a cronjob which runs every 40 minutes generates the required token by running the command:"kubectl create token agent-registration-bootstrap --namespace open-cluster-management" and then generate a kubeconfig and store it somewhere)
-->cluster-manager-post-install.yaml(This yaml has a post install job which creates a initial job from the cronjob since the cronjob only runs every 40 minutes)
->Chart.yaml

custom-ocm-managed-helm(custom helm chart)
->templates
-->klusterlet-helm-release.yaml(This yaml has the HelmRelease which installs klusterlet helmchart)
-->klusterlet-post-install.yaml(This yaml has a post install job which creates the required secret(bootstrap-hub-kubeconfig) after retreiving the kubeconfig generated by hub from storage)
->chart.yaml

When installing the custom-ocm-hub-helm chart, it will generate the kubeconfig and store it in the storage.
While installing the custom-ocm-managed-helm, it will retrieve the secret from storage and create the required secret(bootstrap-hub-kubeconfig).

```

Where can I find these custom charts? If it's not possible to share the link, then at least can you paste the spec.

softcane:
@dosu

I'm getting a fowling error. It would be nice to get the spec on how these should be created and updated.

```Operation for resource &ResourceAttributes{Namespace:,Verb:create,Group:cluster.open-cluster-management.io,Version:,Resource:managedclusters,Subresource:,Name:,FieldSelector:nil,LabelSelector:nil,} is not allowed with bootstrap secret "open-cluster-management-agent"/"bootstrap-hub-kubeconfig" to apiserver https://xxxx.gr7.us-east-1.eks.amazonaws.com```

qiujian16:
It seems the chart is not maintained by the community. Would like to see the update of the community chart accorindly @jaswalkiranavtar

softcane:
@qiujian16
I'm looking for some guidance on how to install OCM cluster manager and Klusterlet using helm charts without using the clusteradm command. on eks.

softcane:
@dosu

[...]

PR #1123, opened 2025-08-12
:sparkles: ✨ feature feat: add labels to managedcluster resource if present 

ramekris3163:
## Summary

Label support for the OCM (Open Cluster Management) ManagedCluster resource, following the same patterns as existing annotation decorator and have the ability to pass the labels as part of the helm chart.

## Related issue(s)
#1118 

## Summary by CodeRabbit

* **New Features**
  * Support for cluster labels during registration and agent bootstrap; labels are applied to ManagedClusters and propagated alongside annotations.
  * New --cluster-labels option to supply key=value labels; startup will include the argument only when provided.
  * CRD/schema extended to accept clusterLabels in registration configuration.

* **Documentation**
  * Added a commented sample showing how to define cluster labels in configuration.

* **Tests**
  * Expanded tests for label/annotation propagation, updates, and equality helpers.


pkg/registration/spoke/registration/creating_controller.go
```diff
@@ -148,3 +170,46 @@ func ClientConfigDecorator(externalServerURLs []string, caBundle []byte) Managed
...
+		}
+		if !bytes.Equal(config1.CABundle, config2.CABundle) {
+			return false
+		}
+	}
+	return true
+}
+
+// equalLabels compares two label maps for equality
+func equalLabels(labels1, labels2 map[string]string) bool {
```
qiujian16, line 192: we cannot do such equal since the label of cluster can also be set by user on hub. So such label setting can only happen when creating the cluster.

If the label exists already on the cluster, the controller cannot update the label or annotation since it might be updated by user on purpose.

pkg/registration/spoke/registration/creating_controller.go
```diff
@@ -148,3 +170,46 @@ func ClientConfigDecorator(externalServerURLs []string, caBundle []byte) Managed
 		return cluster
 	}
 }
+
+// equalClientConfigs compares two ClientConfig slices for equality
+func equalClientConfigs(configs1, configs2 []clusterv1.ClientConfig) bool {
```
qiujian16, line 175: same here, clientconfigs can also be updated by user, so we cannot update clientConfigs based on equality.

pkg/registration/spoke/registration/creating_controller.go
```diff
@@ -112,15 +116,33 @@ func skipUnauthorizedError(err error) error {
...
+}
+
+// LabelDecorator set labels from label map to ManagedCluster
+func LabelDecorator(labels map[string]string) ManagedClusterDecorator {
+	return func(cluster *clusterv1.ManagedCluster) *clusterv1.ManagedCluster {
+		if cluster.Labels == nil {
+			cluster.Labels = make(map[string]string)
+		}
+
+		for k, v := range labels {
```
qiujian16, line 142: note we should not allowing set ANY label from the spoke. For instance, open-cluster-management.io/clusterset label is to define the label on the managedcluster, it should not be set by agent.

qiujian16 (commented):
The main outstanding part is we cannot let agent to update label/annotation/clientConfigs arbitrarily since the these can also be set by user or other actors on the hub. We need to define a more strict rule for agent to set these values on hub.
1. It can only be set when creating.
2. If it is set after creating, there must be a clearer rule to avoid conflict.

pkg/registration/spoke/registration/creating_controller.go
```diff
@@ -89,11 +90,14 @@ func (c *managedClusterCreatingController) sync(ctx context.Context, syncCtx fac
 		managedCluster = decorator(managedCluster)
 	}
 
-	if len(existingCluster.Spec.ManagedClusterClientConfigs) == len(managedCluster.Spec.ManagedClusterClientConfigs) {
```
qiujian16, line 92: the reason we only compare size is because the configs can be set on the hub with different cabundle or endpoint. And the agent should trust setting of hub since hub would have more accurate info on how to access spoke apiserver. 

@zhujian7 I wonder whether to update the cluster should be returned by the decorator rather than compare in the controller.

ramekris3163:
> The main outstanding part is we cannot let agent to update label/annotation/clientConfigs arbitrarily since the these can also be set by user or other actors on the hub. We need to define a more strict rule for agent to set these values on hub.
> 
> 1. It can only be set when creating.
> 2. If it is set after creating, there must be a clearer rule to avoid conflict.

@qiujian16 can we then only enable setting of the label/annotation/clientConfigs during creation and never be able to update it later?

deploy/.DS_Store
zhujian7, line 1: Seems not necessary?

pkg/registration/spoke/registration/creating_controller.go
zhujian7: Yes, make sense, I think we can let the decorator interface return an additional `modified`(bool) value like the openshift library-go does.

qiujian16:
> @qiujian16 can we then only enable setting of the label/annotation/clientConfigs during creation and never be able to update it later?

I think there is case that user create managedcluster at first on the hub without clientConfigs, and then agent should update that when registered which will requires an update, but we do not want it to be updated if the same endpoint already exists on in the clientConfigs. This has not been well documented. 

On the other hand, I think it is fine to only set label/annotation when creating.

arturshadnik:
Hi @qiujian16 @ramekris3163, just wondering if there are any updates on this work? Would be great to see this go in 🙏

mikeshng:
@arturshadnik perhaps you can take it over? WDYT?

PR #1134, opened 2025-08-20
:seedling: support cert auto approve for grpc

skeeey:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

- **New Features**
  - CLI flags to auto-approve specified gRPC users and set gRPC certificate signing duration (default 720h).
  - gRPC path runs CSR approving and signing controllers in parallel.
  - New exported constant to record CSR creating user as an annotation.

- **Improvements**
  - CSR handling is signer-aware with a standardized exported CSR info shape and multi-version CSR support.

- **Tests**
  - New integration/unit tests for mTLS, CSR auto-approval, event filtering, and certificate rotation.

- **Chores**
  - Dependency cleanup in go.mod.

### DIFF
--- a/pkg/common/helpers/constants.go
+++ b/pkg/common/helpers/constants.go
@@ -7,3 +7,6 @@ const (
 )
 
 const GRPCCAuthSigner = "open-cluster-management.io/grpc"
+
+// CSRUserAnnotation will be added to a CSR and used to identify the user who request the CSR
+const CSRUserAnnotation = "open-cluster-management.io/csruser"
--- a/pkg/registration/hub/manager.go
+++ b/pkg/registration/hub/manager.go
@@ -60,8 +60,11 @@ type HubManagerOptions struct {
 	AutoApprovedARNPatterns    []string
 	AwsResourceTags            []string
 	Labels                     string
-	GRPCCAFile                 string
-	GRPCCAKeyFile              string
+	// TODO (skeeey) introduce hub options for different drives to group these options
+	AutoApprovedGRPCUsers []string
+	GRPCCAFile            string
+	GRPCCAKeyFile         string
+	GRPCSigningDuration   time.Duration
 }
 
 // NewHubManagerOptions returns a HubManagerOptions
@@ -71,6 +74,7 @@ func NewHubManagerOptions() *HubManagerOptions {
 			"work.open-cluster-management.io/v1/manifestworks"},
 		ImportOption:               importeroptions.New(),
 		EnabledRegistrationDrivers: []string{commonhelpers.CSRAuthType},
+		GRPCSigningDuration:        720 * time.Hour,
 	}
 }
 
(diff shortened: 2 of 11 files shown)


pkg/registration/register/csr/approve_reconciler.go
```diff
@@ -234,3 +240,27 @@ func newCSRInfo(logger klog.Logger, csr any) csrInfo {
...
+	switch *csr.Spec.SignerName {
+	case certificatesv1.KubeAPIServerClientSignerName:
+		return csr.Spec.Username
+	case helpers.GRPCCAuthSigner:
+		return csr.Annotations[helpers.CSRUserAnnotation]
+	default:
+		klog.Errorf("unsupported CSR signer %s", *csr.Spec.SignerName)
+		return ""
+	}
+}
```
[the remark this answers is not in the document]
skeeey: the signer name should not be a nil here

pkg/server/services/csr/csr.go
```diff
@@ -84,6 +86,7 @@ func (c *CSRService) HandleStatusUpdate(ctx context.Context, evt *cloudevents.Ev
 
 	switch eventType.Action {
 	case types.CreateRequestAction:
+		csr.Annotations = map[string]string{helpers.CSRUserAnnotation: fmt.Sprintf("%v", ctx.Value(authn.ContextUserKey))}
 		_, err := c.csrClient.CertificatesV1().CertificateSigningRequests().Create(ctx, csr, metav1.CreateOptions{})
 		return err
```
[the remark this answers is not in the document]
skeeey: we need create new annotation here

pkg/common/helpers/constants.go
```diff
@@ -7,3 +7,5 @@ const (
 )
 
 const GRPCCAuthSigner = "open-cluster-management.io/grpc"
+
+const CSRUserAnnotation = "open-cluster-management.io/csruser"
```
qiujian16, line 12: comments on this annoation is added on which resources, and the what the value it should be

pkg/registration/hub/manager.go
```diff
@@ -58,10 +58,12 @@ type HubManagerOptions struct {
 	HubClusterArn              string
 	AutoApprovedCSRUsers       []string
 	AutoApprovedARNPatterns    []string
+	AutoApprovedGRPCUsers      []string
 	AwsResourceTags            []string
 	Labels                     string
 	GRPCCAFile                 string
 	GRPCCAKeyFile              string
+	GRPCSigningDuration        time.Duration
```
qiujian16: that seems like we should move a driver specific option/flag to driver itself.

pkg/registration/register/csr/approve_reconciler.go
```diff
@@ -209,7 +215,7 @@ func newCSRInfo(logger klog.Logger, csr any) csrInfo {
 			name:       v.Name,
 			labels:     v.Labels,
 			signerName: v.Spec.SignerName,
-			username:   v.Spec.Username,
+			username:   csrUsername(v),
```
qiujian16: this should be default to userName, and can be customized in grpc driver.

pkg/server/services/csr/csr.go
```diff
@@ -84,6 +86,7 @@ func (c *CSRService) HandleStatusUpdate(ctx context.Context, evt *cloudevents.Ev
 
 	switch eventType.Action {
 	case types.CreateRequestAction:
+		csr.Annotations = map[string]string{helpers.CSRUserAnnotation: fmt.Sprintf("%v", ctx.Value(authn.ContextUserKey))}
```
qiujian16, line 93: comments here, on why we need to add this annotation.

pkg/registration/register/grpc/hub_driver.go
```diff
@@ -53,7 +69,14 @@ func NewGRPCHubDriver(
 		return nil, err
 	}
 	return &GRPCHubDriver{
-		controller: newCSRSignController(
+		csrApprovingController: csr.NewCSRApprovingController(
+			kubeInformers.Certificates().V1().CertificateSigningRequests().Informer(),
```
qiujian16, line 74: do we have a event filter in approving controller? We need it to filter based on signer and label.

pkg/registration/register/csr/approve_reconciler.go
skeeey: > can be customized in grpc driver

Define a different way to get username in grpc driver api? e.g. from username/annotation?

for grpc registration, and it seems we can only get the username from the annotation？

skeeey: or give a `usernameGetter` function for each driver implementation?

csr driver will get username from spec
grcp dirver will get username from annotation

pkg/common/helpers/constants.go
skeeey: added

pkg/registration/hub/manager.go
skeeey: yeah, we have the agent option for each driver for now, I think we also need the hub option for each driver
Add a TODO for this as the following task

pkg/server/services/csr/csr.go
skeeey: added

pkg/registration/register/csr/approve_reconciler.go
skeeey: refactor this part, I add a crsinfo getter for approving controller, then each driver is able to have its own getter

pkg/registration/register/grpc/hub_driver.go
skeeey: we have filtered the label in the kubeInformers, I add an event filter in approving controller to filter the signer

pkg/server/services/csr/csr.go
qiujian16: could you add something like 
the username in the csr is the service account of gRPC rather than the user of agent. The approver controller in the registration will not be able to know where this csr originates from. Hence the annotation with agent user name is added here.

skeeey: added

PR #1138, opened 2025-08-25
🐛 Fix flaky issue in e2e tests

qiujian16:
## Summary

Only start one clustermater replica to reduce resource usate
Adjust sequence of remove klusterlet.

## Related issue(s)

Fixes #


zhujian7:
LGTM

Issue #138, opened 2023-05-18
Request for new Release cut for the Add-ons Cluster-proxy, Managed-ServiceAccount and Cluster-Gateway


suresh-j2:
**Describe the bug**
I would like to request a new release cut for the add-ons. cluster-proxy, managed-serviceaccount and cluster-gateway

**To Reproduce**

**Expected behavior**

**Environment ie: OCM version, Kubernetes version and provider:**

**Additional context**
Currently the Add-on helm charts are not updated for few months and hence its not working as per the document.

xuezhaojun:
@suresh-j2  Hello, "its not working as per the document", do you mean this one: https://open-cluster-management.io/scenarios/pushing-kube-api-requests/ ?

suresh-j2:
@xuezhaojun yes thats right. Also, was facing issue initially doc as well. https://open-cluster-management.io/developer-guides/addon/.  The release version is not having certain packages used in the code example.

xuezhaojun:
Hi, @suresh-j2, we're going to release 0.3.0 for cluster-proxy and managed-serviceaccount:
* https://github.com/open-cluster-management-io/cluster-proxy/pull/161
* https://github.com/open-cluster-management-io/managed-serviceaccount/pull/64

After the Pull Request is merged, we are going to update the document as well.

I notice you said you're facing problems in https://open-cluster-management.io/developer-guides/addon/, do you mind sharing what specific issue you're facing?

suresh-j2:
Thanks @xuezhaojun .  Sorry the link was wrong. I guess, i was facing issue while trying out this examples. https://github.com/open-cluster-management-io/cluster-proxy/blob/main/examples/access-exported-services/main.go. These code refers to some of the methods which is not released. For ex, GetProxyHost. I believe with the new release this should also be fixed.

xuezhaojun:
Got it, I will notify in this thread once new release is done.

Issue #135, opened 2023-05-11
 Enhance the placement schedule time when add a new cluster


ldpliu:
The placement schedule time is too long when add a new cluster.
steps:
1. create 100 managedclusters
2. create 100 placements which select these managedclusters
3. create a new managedcluster which should also be selected by all the placements
4. watch the placement schedule time(placement.Status.NumberOfSelectedClusters  PlacementDecisions.Status.Decisions)

![image](https://github.com/open-cluster-management-io/OCM/assets/56991288/45eb316e-d75e-46dc-9f6c-da00fd2067a9)

![image](https://github.com/open-cluster-management-io/OCM/assets/56991288/ca74bff6-2244-4f02-8737-cca063e43159)

PR #1157, opened 2025-09-02
🌱 upgrade grpc server.

morvencao:
Signed-off-by: morvencao <lcao@redhat.com>

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

## Summary

Upgrade the gRPC server to use the new structure ([#135](https://github.com/open-cluster-management-io/sdk-go/pull/135)) and metrics ([#138](https://github.com/open-cluster-management-io/sdk-go/pull/138)) from `sdk-go`.

## Related issue(s)

Fixes #


pkg/server/grpc/options.go
```diff
@@ -47,29 +51,34 @@ func (o *GRPCServerOptions) Run(ctx context.Context, controllerContext *controll
...
+	grpcEventServer.RegisterService(eventce.EventEventDataType,
+		event.NewEventService(clients.KubeClient))
+	grpcEventServer.RegisterService(leasece.LeaseEventDataType,
+		lease.NewLeaseService(clients.KubeClient, clients.KubeInformers.Coordination().V1().Leases()))
+	grpcEventServer.RegisterService(payload.ManifestBundleEventDataType,
+		work.NewWorkService(clients.WorkClient, clients.WorkInformers.Work().V1().ManifestWorks()))
+
+	// initlize and run grpc server
+	authorizer := grpcauthz.NewSARAuthorizer(clients.KubeClient)
+	return sdkgrpc.NewGRPCServer(serverOptions).
```
skeeey, line 74: can we use the `WithPreStartHooks` to start clients here?

morvencao: preStartHooks were removed in https://github.com/open-cluster-management-io/sdk-go/pull/135

morvencao: I can see from the given example, clients needs to be started in a go rountine: https://github.com/open-cluster-management-io/sdk-go/pull/135/#issue-3291389695

skeeey: hrm... the `StartHooks` may help us to provide a clear code struct, @xuezhaojun do we have some concerns for them?

xuezhaojun: I'm curious about why `StartHooks` provides a clearer code structure.

I would prefer to initialize everything before starting. If something needs to be hooked just before or after the start process, and timing is critical, I agree we should add hooks back. 

However, for this particular case, we can initialize everything before starting, so I don't see the necessity of using a hook.


Just like client-go/controller-runtime's manager doesn't have a hook to start informer.

xuezhaojun: An example: https://github.com/stolostron/ocm/blob/8b2ba5b7af8b827c2251c3ebad6a1bd4af722403/pkg/registration/hub/manager.go#L348-L365

```go
	go clusterInformers.Start(ctx.Done())
	go workInformers.Start(ctx.Done())
	go kubeInformers.Start(ctx.Done())
	go addOnInformers.Start(ctx.Done())
	if features.HubMutableFeatureGate.Enabled(ocmfeature.ClusterProfile) {
		go clusterProfileInformers.Start(ctx.Done())
	}

	go managedClusterController.Run(ctx, 1)
	go taintController.Run(ctx, 1)
	go hubDriver.Run(ctx, 1)
	go leaseController.Run(ctx, 1)
	go clockSyncController.Run(ctx, 1)
	go managedClusterSetController.Run(ctx, 1)
	go managedClusterSetBindingController.Run(ctx, 1)
	go clusterroleController.Run(ctx, 1)
	go addOnHealthCheckController.Run(ctx, 1)
	go addOnFeatureDiscoveryController.Run(ctx, 1)
```

Controllers don't set a `PreRunHook` and put informer's `Start` into it.

xuezhaojun: cc @qiujian16 thoughts?

skeeey: I think PreStartHooks help us clear to know what should be start previously

qiujian16: If we need to control some thing start exactly before server starts, having PreStartHooks is able to control the sequence.

xuezhaojun: I don't think this is a `start exactly before` case, and maybe except for grpc server, we have other objects also need clients as inputs. I don't think let grpc server to control client's start is a good idea.

qiujian16: apiserver has postStartHook https://github.com/kubernetes/apiserver/blob/dac6b3e30053051c3cd5569497b054a33b3d1fa7/pkg/server/config.go#L583.

xuezhaojun: I also don't think `PreStartHook` can remind sdk-go's user to start clients. 

```
// PreStartHook is an interface to start hook before grpc server is started.
type PreStartHook interface {
	// Start should be a non-blocking call
	Run(ctx context.Context)
}
```

This is the original implementation of `PreStartHook`, if I'm a sdk-go user, I foget that kubeclient need to `Start` before use.  Reading this interface can't remind me that.

xuezhaojun: @qiujian16 I think apiserver has a good practice because it need hooks to excute **exact** after `Start`. 

**But it doesn't set a function to wrap everything it needs for starting**, otherwise nanturally, we will see a `PreStartHook` function.

It's the same with our case, we can put `clients.Start` anywhere before `server.Start`, exact timing doesn't matter.

The reason why apiserver doesn't need a PreHook is also the reason why we don't need a PreHook here.

xuezhaojun: The analysis from claude:

Prompt:(No bias)
```
In this PR https://github.com/open-cluster-management-io/sdk-go/pull/135, we remove PreStartHook function, and in this discussion and following: 
      https://github.com/open-cluster-management-io/ocm/pull/1157#discussion_r2315046833 We discussion whether should we add PreStartHook back. What do you think?
```

---

Result:
```
⏺ Based on my analysis of PR #135 and the discussion, here's my perspective on whether PreStartHook should be added back:

  I recommend NOT adding PreStartHook back for the following reasons:

  1. Current Architecture is Cleaner and More Explicit

  The new pattern in PR #135 shows explicit initialization:
  go clients.Run(ctx)  // Clear, explicit client startup

  return grpcbuilder.NewGRPCServerBuilder(serverOptions)
      .WithAuthenticator(...)
      .WithRegisterFunc(grpcEventServer.Register)
      .Run(ctx)

  This is much clearer than the previous hook pattern where initialization was buried in unclear hooks.

  2. Original PreStartHook Interface Was Problematic

  From the PR diff, the original interface was:
  type PreStartHook interface {
      // Start should be a non-blocking call
      Run(ctx context.Context)
  }

  As xuezhaojun pointed out in the discussion: "Reading this interface can't remind me that [kubeclient need to Start]" - the interface didn't clearly communicate what
  needed to be started or when.

  3. Builder Pattern Provides Better Control

  The current GRPCServer from pkg/server/grpc/server.go:24 uses a clean builder pattern with:
  - WithRegisterFunc() for service registration
  - WithAuthenticator() for auth
  - WithUnaryAuthorizer() / WithStreamAuthorizer() for authorization

  This provides the same flexibility as hooks but with clear, typed interfaces.

  4. Explicit Ordering is Superior

  The hook pattern made startup ordering implicit and potentially confusing. The current approach makes dependencies explicit:
  1. Start clients first: go clients.Run(ctx)
  2. Then build and start server with all dependencies ready

  5. Kubernetes Community Trend

  While the apiserver uses post-start hooks, the trend in modern Kubernetes components is toward explicit initialization patterns rather than hook-based systems for better
   maintainability.

  Conclusion: The current explicit initialization pattern is clearer, more maintainable, and eliminates the ambiguity that existed with PreStartHook. The builder pattern
  already provides sufficient flexibility for different initialization needs.
```

morvencao: any conclusion on this? should we keep the current behavior or start a new thread to continue the discussion?

qiujian16: we do not need to resolve that in this PR I think

PR #1158, opened 2025-09-02
✨ Deleted manifestwork when it is completed for ttl seconds.

qiujian16:
## Related issue(s)

Fixes #


qiujian16:
/assign @bhperry 
/hold
still need to update the operator part, we might need a new feature gate CleanUpCompletedWork in API at first.

test/integration/work/completedmanifestwork_test.go
```diff
@@ -0,0 +1,258 @@
...
+			work, err = hubWorkClient.WorkV1().ManifestWorks(commOptions.SpokeClusterName).Create(context.Background(), work, metav1.CreateOptions{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			ginkgo.By("Verifying the ManifestWork is deleted immediately with zero TTL")
+			// Should be deleted quickly since TTL is 0
+			gomega.Eventually(func() bool {
+				_, err = hubWorkClient.WorkV1().ManifestWorks(commOptions.SpokeClusterName).Get(context.Background(), workName, metav1.GetOptions{})
+				return errors.IsNotFound(err)
+			}, 15*time.Second, eventuallyInterval).Should(gomega.BeTrue())
+		})
```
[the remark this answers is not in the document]
qiujian16: it would be flaky since manifestwork may be deleted already when checking complete status

pkg/work/hub/controllers/completedmanifestwork/controller.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// CompletedManifestWorkController is to delete the manifestworks when it has the completed condition.
+type CompletedManifestWorkController struct {
+	workClient workclientset.Interface
+	workLister worklisters.ManifestWorkLister
+}
+
+// NewCompletedManifestWorkController creates a new CompletedManifestWorkController
+func NewCompletedManifestWorkController(
```
bhperry: Could there be other cases where manifestworks would be garbage collected in future? This could have a more generic name like `ManifestWorkGarbageCollectionController` so that it can more easily be extended. I think that also makes it more clear what its role is.

qiujian16: updated

qiujian16:
operator part and e2e are added.
/unhold

pkg/work/hub/controllers/completedmanifestwork/controller.go
bhperry: Probably should change the feature gate from https://github.com/open-cluster-management-io/api/pull/394 to match

qiujian16: Hrm I think feature gate can be more specific. It is possible we enable this feature gate but disable other relating to gc in the future.

deploy/klusterlet/chart/klusterlet/crds/0000_00_operator.open-cluster-management.io_klusterlets.crd.yaml
```diff
@@ -45,7 +45,7 @@ spec:
             properties:
               clusterName:
                 description: |-
-                  ClusterName is the name of the managed cluster to be created on hub.
+                  clusterName is the name of the managed cluster to be created on hub.
```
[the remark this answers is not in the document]
qiujian16: this is copied from vendor via make update and guarded by make verify

pkg/work/hub/controllers/completedmanifestwork/controller.go
bhperry: The purpose of the flag is to prevent starting an extra controller by default right? So would it have to search for any releated GC flags in order to determine if it should start the controller? Seems simpler to have one flag for the controller, and add additional flags in future if they are needed.

qiujian16: so there are two features will start this extra controller, ManifestWorkReplicaSet and CleanUp feature. Any of these two feature enabled will start the controller, and the controller will has the feature flags that operator will set to start separated goroutine for each feature. I expect later this controller will always start, and whether a feature is enabled will be controlled by the feature flag on this controller.

Issue #1156, opened 2025-09-02
ManagedClusterAddons are not removed when ClusterManagementAddon is deleted


haoqing0110:
**Describe the bug**

The addon is implemented by addontemplate, when ClusterManagementAddon is deleted, the ManagedClusterAddons are not removed.

**To Reproduce**

**Expected behavior**

**Environment ie: OCM version, Kubernetes version and provider:**

**Additional context**

PR #1160, opened 2025-09-03
🐛 Fix ManagedClusterAddons not removed when ClusterManagementAddon is deleted

zhujian7:
The addon template controller was stopping addon managers immediately when ClusterManagementAddon was deleted, without waiting for pre-delete jobs to complete or ManagedClusterAddons to be cleaned up via owner reference cascading deletion.

This change implements the TODO at line 105 by checking if all ManagedClusterAddons are deleted before stopping the manager. The controller now uses field selectors to efficiently query for remaining ManagedClusterAddons and requeues after 10 seconds if any still exist, allowing time for proper cleanup.

🤖 Generated with [Claude Code](https://claude.ai/code)

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/1156

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevents add-on managers from stopping while dependent managed add-ons still exist; fixes a ConfigMap test variable typo.

- **New Features**
  - Reduces unnecessary reconciliations by filtering to templates referenced by managed add-ons and uses indexed discovery for manager gating.

- **Tests**
  - Adds unit tests for gated shutdown scenarios and e2e checks verifying deletion waits for propagated-resource cleanup (two identical e2e cases present).

- **Chores**
  - Improved structured, context-aware informational logging.

### DIFF
--- a/pkg/addon/controllers/addontemplate/controller.go
+++ b/pkg/addon/controllers/addontemplate/controller.go
@@ -2,7 +2,6 @@ package addontemplate
 
 import (
 	"context"
-	"reflect"
 	"time"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
@@ -15,7 +14,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 
 	"open-cluster-management.io/addon-framework/pkg/addonfactory"
@@ -54,7 +52,6 @@ type addonTemplateController struct {
 	workInformers              workv1informers.SharedInformerFactory
 	runControllerFunc          runController
 	eventRecorder              events.Recorder
-	queue                      workqueue.TypedRateLimitingInterface[any]
 }
 
 type runController func(ctx context.Context, addonName string) error
@@ -72,9 +69,6 @@ func NewAddonTemplateController(
 	recorder events.Recorder,
 	runController ...runController,
 ) factory.Controller {
-	controllerName := "addon-template-controller"
-	syncCtx := factory.NewSyncContext(controllerName, recorder)
-
 	c := &addonTemplateController{
 		kubeConfig:                 hubKubeconfig,
 		kubeClient:                 hubKubeClient,
@@ -88,7 +82,6 @@ func NewAddonTemplateController(
 		dynamicInformers:           dynamicInformers,
 		workInformers:              workInformers,
 		eventRecorder:              recorder,
-		queue:                      syncCtx.Queue(),
 	}
 
 	if len(runController) > 0 {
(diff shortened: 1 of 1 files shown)


pkg/addon/controllers/addontemplate/controller.go
```diff
@@ -101,14 +101,35 @@ func NewAddonTemplateController(
...
-	// TODO: check if all managed cluster addon instances are deleted
+	ctx context.Context, syncCtx factory.SyncContext, addOnName string) error {
+	logger := klog.FromContext(ctx)
+
+	// Check if all managed cluster addon instances are deleted before stopping the manager
+	// This allows pre-delete jobs to complete
+	listOptions := metav1.ListOptions{
+		FieldSelector: "metadata.name=" + addOnName,
+	}
+	mcaList, err := c.addonClient.AddonV1alpha1().ManagedClusterAddOns("").List(ctx, listOptions)
```
qiujian16: can we list from lister instead calling api directly?

qiujian16 (commented):
another thing we need to confirm is when cma is deleting, does install controller still create mca?

pkg/addon/controllers/addontemplate/controller.go
```diff
@@ -101,14 +108,40 @@ func NewAddonTemplateController(
...
 
 func (c *addonTemplateController) stopUnusedManagers(
-	ctx context.Context, syncCtx factory.SyncContext, addOnName string) {
-	// TODO: check if all managed cluster addon instances are deleted
+	ctx context.Context, syncCtx factory.SyncContext, addOnName string) error {
+	logger := klog.FromContext(ctx)
+
+	// Check if all managed cluster addon instances are deleted before stopping the manager
+	// This allows pre-delete jobs to complete
+	mcaList, err := c.mcaLister.List(labels.Everything())
```
qiujian16: I think we have an index addonindex.ManagedClusterAddonByName for this.

pkg/addon/controllers/addontemplate/controller.go
```diff
@@ -101,14 +108,40 @@ func NewAddonTemplateController(
...
+			matchingMCAs = append(matchingMCAs, mca)
+		}
+	}
+
+	// Check if there are still ManagedClusterAddOns for this addon
+	if len(matchingMCAs) > 0 {
+		logger.Info("ManagedClusterAddOn still exists, waiting for deletion",
+			"addonName", addOnName, "count", len(matchingMCAs))
+		// Requeue to check again later
+		syncCtx.Queue().AddAfter(addOnName, requeueDelay)
```
qiujian16: do we need to requeue? deletion of MCA will also trigger the reconcile, right?

zhujian7: no, the controller only watches CMA

zhujian7:
@qiujian16 @haoqing0110 PTAL

pkg/addon/controllers/addontemplate/controller.go
qiujian16: should we watch MCA? we could filter by configuration in status of mca

zhujian7:
@qiujian16 PTAL

pkg/addon/controllers/addontemplate/controller.go
```diff
@@ -88,27 +97,98 @@ func NewAddonTemplateController(
...
+			if !ok {
+				return
+			}
+			newMCA, ok := newObj.(*addonv1alpha1.ManagedClusterAddOn)
+			if !ok {
+				return
+			}
+
+			// Extract AddOnTemplate configReferences from both old and new
+			oldTemplateRefs := extractAddOnTemplateConfigRefs(oldMCA)
```
qiujian16: can we just filter the mca based on whether it is an addonTemplate based addon?

pkg/addon/controllers/addontemplate/controller.go
```diff
@@ -88,27 +97,98 @@ func NewAddonTemplateController(
...
+			newTemplateRefs := extractAddOnTemplateConfigRefs(newMCA)
+
+			// Only process if AddOnTemplate configReferences changed
+			if !reflect.DeepEqual(oldTemplateRefs, newTemplateRefs) {
+				// Queue the addon name to trigger reconciliation
+				c.queue.Add(newMCA.Name)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			mca, ok := obj.(*addonv1alpha1.ManagedClusterAddOn)
```
qiujian16: same here, we only need to care template based addon.

pkg/addon/controllers/addontemplate/controller.go
zhujian7: This will not only filter the addonTemplate-based addon, but also filter only the template that has changes.
Do you mean we do not need to compare whether the template has changed, we enqueue all update events as long as it is template-based?

pkg/addon/controllers/addontemplate/controller.go
zhujian7: Fixed in the below lines 145-148

pkg/addon/controllers/addontemplate/controller.go
qiujian16: I think so, why do we only compare template change? If this is only for deleting, should we just filter the one that is template addon and is in deleting?

zhujian7: Fixed. PTAL

PR #1170, opened 2025-09-08
:sparkles: [wip] support cluster import config secret

zhiweiyin318:
## Summary
support cluster-import-config secret to render the klusterlet chart during import CPAI cluster

## Related issue(s)

Fixes #


pkg/registration/hub/importer/options/options.go
```diff
@@ -1,16 +1,30 @@
...
+	RenderFromConfigSecret string = "render-from-config-secret"
+	RenderAuto             string = "render-auto"
+)
+
 func New() *Options {
 	return &Options{
-		BootstrapSA: "open-cluster-management/agent-registration-bootstrap",
+		BootstrapSA:          "open-cluster-management/agent-registration-bootstrap",
+		ImporterRendererList: []string{RenderFromConfigSecret},
 	}
```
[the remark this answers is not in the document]
zhiweiyin318: the default renderer is changed to RenderFromConfigSecret

pkg/registration/hub/importer/options/options.go
```diff
@@ -1,16 +1,30 @@
...
-	AgentImage   string
-	BootstrapSA  string
+	APIServerURL      string
+	AgentImage        string
+	BootstrapSA       string
+	ImporterRenderers []string
 }
 
+const (
+	RenderFromConfigSecret string = "render-from-config-secret"
```
qiujian16, line 21: we need to add some description for each for this const

manifests/cluster-manager/hub/registration/clusterrole.yaml
```diff
@@ -115,6 +115,12 @@ rules:
 - apiGroups: ["cluster.x-k8s.io"]
   resources: ["clusters"]
   verbs: ["get", "list", "watch"]
+- apiGroups: [""]
```
qiujian16, line 120: add comments here on this is for registration to import clusters.

pkg/registration/helpers/constants.go
```diff
@@ -0,0 +1,6 @@
+package helpers
+
+const (
+	ClusterImportConfigSecret = "cluster-import-config"
```
qiujian16: why do we define a new file here? can we just put in the importer directory if it is only for importer.

pkg/registration/hub/importer/options/options.go
zhiweiyin318: added

manifests/cluster-manager/hub/registration/clusterrole.yaml
zhiweiyin318: added

pkg/registration/helpers/constants.go
zhiweiyin318: removed the new file. moved the const to importer.go

PR #1191, opened 2025-09-22
:bug: Add startTime initialization and wait in hubTimeoutController

xuezhaojun:


pkg/registration/spoke/registration/hub_timeout_controller.go
```diff
@@ -57,6 +60,13 @@ func (c *hubTimeoutController) sync(ctx context.Context, syncCtx factory.SyncCon
 		c.lastLeaseRenewTime = lease.Spec.RenewTime.Time
 	}
```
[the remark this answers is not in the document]
xuezhaojun: No, lease.Spec.RenewTime.Time is not a pointer, but a time.Time. It won't be nil.

pkg/registration/spoke/registration/hub_timeout_controller.go
```diff
@@ -57,6 +60,13 @@ func (c *hubTimeoutController) sync(ctx context.Context, syncCtx factory.SyncCon
...
 
+	// If `startTime` within 10s, skip the timeout check.
+	// This is to wait for the first lease update. Otherwise, if old lease is already
+	// expired, it will be treated as timeout immediately.
+	if time.Since(c.startTime) < time.Second*10 {
```
qiujian16, line 78: Why old lease is still there? should it be deleted already? Is this because we forget to delete lease when delete the managedcluster?

xuezhaojun: In this global hub's migration case: [first, migrate from hub1 to hub2, and then migrate back from hub2 to hub1],  in the hub1, gh doesn't delete resources in the managed cluster namespace, so resources like lease of the managed cluster exist in the process.

xuezhaojun: In similar cases with other seniors, for instance, if Hub1 unexpectedly fails (such as a power outage), we don't get the opportunity to clean up everything in Hub1, leaving the old lease still active there. The managed cluster detects the timeout and switches to Hub2.

When Hub1 comes back online and we attempt to migrate the managed cluster back, we end up migrating it to a hub (Hub1) where the old lease still exists, which subsequently triggers a timeout if timeout controller run before lease controller.

qiujian16: could you add this case as comment here without mentioning global hub? So we know clearly why this is added and avoid accidentally removal.

xuezhaojun: Good point! The complete case is added as the comment above the code.

PR #1193, opened 2025-09-23
✨ support managed namespaces

elgnay:
Related KEP: https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/146-fleet-namespaces

## Related issue(s)

Fixes #

## Summary by CodeRabbit

- **New Features**
  - Hub aggregates managed namespaces across cluster sets and updates cluster status; a new controller is started during startup.
  - Spoke agents compute a hub-specific hash, use hub-scoped namespace informers, and reconcile hub-managed namespaces on spokes with hub-scoped labels, per-namespace status, and cleanup of stale namespaces.
  - Rollback behavior for default/global cluster sets now only restores the cluster selector (other fields remain editable).

- **Chores**
  - Agent RBAC expanded to allow namespace create/get/list/update/watch/patch.

- **Tests**
  - Added extensive unit and e2e tests covering propagation, labeling, status conditions, updates, cleanup, multi-cluster-set scenarios, and hub-hash initialization.

### DIFF
--- a/pkg/registration/hub/managedcluster/managed_namespace_controller.go
+++ b/pkg/registration/hub/managedcluster/managed_namespace_controller.go
@@ -1,0 +1,243 @@
+package managedcluster
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"sort"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	clientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	clusterinformerv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
+	clusterinformerv1beta2 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1beta2"
+	clusterlisterv1 "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	clusterlisterv1beta2 "open-cluster-management.io/api/client/cluster/listers/cluster/v1beta2"
+	v1 "open-cluster-management.io/api/cluster/v1"
+	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta2"
+	clustersdkv1beta2 "open-cluster-management.io/sdk-go/pkg/apis/cluster/v1beta2"
+	"open-cluster-management.io/sdk-go/pkg/patcher"
+)
+
+// managedNamespaceController reconciles managed namespaces for ManagedClusters
+// by watching ManagedCluster changes and updating their managed namespace status based on
+// all ManagedClusterSets they belong to
+type managedNamespaceController struct {
+	clusterPatcher   patcher.Patcher[*v1.ManagedCluster, v1.ManagedClusterSpec, v1.ManagedClusterStatus]
+	clusterLister    clusterlisterv1.ManagedClusterLister
+	clusterSetLister clusterlisterv1beta2.ManagedClusterSetLister
+	eventRecorder    events.Recorder
+}
+
+// NewManagedNamespaceController creates a new managed namespace controller
+func NewManagedNamespaceController(
+	clusterClient clientset.Interface,
+	clusterInformer clusterinformerv1.ManagedClusterInformer,
+	clusterSetInformer clusterinformerv1beta2.ManagedClusterSetInformer,
+	recorder events.Recorder) factory.Controller {
+
+	controllerName := "managed-namespace-controller"
+	syncCtx := factory.NewSyncContext(controllerName, recorder)
+
+	c := &managedNamespaceController{
+		clusterPatcher: patcher.NewPatcher[
+			*v1.ManagedCluster, v1.ManagedClusterSpec, v1.ManagedClusterStatus](
+			clusterClient.ClusterV1().ManagedClusters()),
+		clusterLister:    clusterInformer.Lister(),
+		clusterSetLister: clusterSetInformer.Lister(),
+		eventRecorder:    recorder.WithComponentSuffix("managed-namespace-controller"),
+	}
+
+	// Add explicit event handlers for ManagedCluster
+	_, err := clusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			cluster, ok := obj.(*v1.ManagedCluster)
+			if !ok {
+				utilruntime.HandleError(fmt.Errorf("error to get ManagedCluster object: %v", obj))
+				return
+			}
+			syncCtx.Queue().Add(cluster.Name)
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			oldCluster, ok := oldObj.(*v1.ManagedCluster)
+			if !ok {
+				utilruntime.HandleError(fmt.Errorf("error to get old ManagedCluster object: %v", oldObj))
+				return
+			}
+			newCluster, ok := newObj.(*v1.ManagedCluster)
+			if !ok {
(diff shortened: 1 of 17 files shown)


pkg/registration/hub/managedclusterset/managed_namespace_controller.go
```diff
@@ -0,0 +1,352 @@
...
+		WithSyncContext(syncCtx).
+		WithInformersQueueKeysFunc(queue.QueueKeyByMetaName, clusterSetInformer.Informer()).
+		WithBareInformers(clusterInformer.Informer()).
+		WithSync(c.sync).
+		ToController("ManagedNamespaceController", recorder)
+}
+
+func (c *managedNamespaceController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	logger := klog.FromContext(ctx)
+	clusterSetName := syncCtx.QueueKey()
```
qiujian16: Actually I am thinking that we enqueue cluster upon clusterset event, enqueue all clusterst in the clusterset and filtered cluster event (only enqueue with label change). So we only update one cluster in each sync loop.

pkg/registration/hub/managedclusterset/managed_namespace_controller.go
```diff
@@ -0,0 +1,352 @@
...
+
+	// Get all clusters that previously had this cluster set's namespaces
+	previousClusters, err := c.getClustersPreviouslyInSet(clusterSet.Name)
+	if err != nil {
+		return fmt.Errorf("failed to get clusters previously in cluster set %q: %w", clusterSet.Name, err)
+	}
+
+	// Update clusters that should have these namespaces
+	for _, cluster := range currentClusters {
+		if err := c.updateClusterManagedNamespaces(ctx, cluster, clusterSet); err != nil {
```
qiujian16: I think for each cluster, we can find all related clustersets and then build a namespace list.

pkg/registration/spoke/managedcluster/namespace_controller.go
```diff
@@ -0,0 +1,262 @@
...
+
+	if err := c.reconcileManagedNamespaces(ctx, cluster); err != nil {
+		return fmt.Errorf("failed to reconcile managed namespaces for cluster %q: %w", c.clusterName, err)
+	}
+
+	return nil
+}
+
+// reconcileManagedNamespaces reconciles managed namespaces for the cluster
+func (c *managedNamespaceController) reconcileManagedNamespaces(ctx context.Context, cluster *clusterv1.ManagedCluster) error {
```
qiujian16: actually could we put this as a reconciler in managedcluster controller. This will help do reduce the conflict risk. The managedcluster controller will update status once when all reconciler is done.

pkg/registration/hub/managedclusterset/managed_namespace_controller.go
elgnay: Updated as suggested.

pkg/registration/hub/managedclusterset/managed_namespace_controller.go
elgnay: Updated as suggested.

pkg/registration/spoke/managedcluster/namespace_controller.go
elgnay: Updated as suggested.

pkg/registration/hub/managedcluster/managed_namespace_controller.go
```diff
@@ -0,0 +1,292 @@
...
+			if !ok {
+				utilruntime.HandleError(fmt.Errorf("error to get new ManagedCluster object: %v", newObj))
+				return
+			}
+			// Only care about label changes that affect cluster set membership
+			if !reflect.DeepEqual(oldCluster.Labels, newCluster.Labels) {
+				syncCtx.Queue().Add(newCluster.Name)
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
```
qiujian16: Is delete necessary? After deleted event, most likely you will not get it in the cache.

pkg/registration/hub/managedcluster/managed_namespace_controller.go
```diff
@@ -0,0 +1,292 @@
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	// If cluster is being deleted, clear managed namespaces
+	if !cluster.DeletionTimestamp.IsZero() {
+		logger.V(4).Info("ManagedCluster is being deleted, clearing managed namespaces", "clusterName", clusterName)
+		return c.clearManagedNamespaces(ctx, cluster)
```
qiujian16: Why do we need this step?

pkg/registration/hub/managedcluster/managed_namespace_controller.go
```diff
@@ -0,0 +1,292 @@
...
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("expected ManagedClusterSet, got %T", obj))
+		return nil
+	}
+
+	if clusterSet == nil {
+		return nil
+	}
+
+	clusterNames := make(map[string]bool)
```
qiujian16: I think you can just use set[string]

pkg/registration/spoke/managedcluster/namespace_controller.go
qiujian16: is this part already updated? I think we will move this to as a reconciler and add here https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/spoke/managedcluster/status_controller.go#L112?

pkg/registration/hub/managedcluster/managed_namespace_controller.go
elgnay: Clearing the managed cluster when it enters the terminating state triggers an update to the `clusterset.open-cluster-management.io/<hub_hash>` label on the managed namespaces on the managed cluster. After the cluster is detached, the label on the managed namespace should be set to `false`. Otherwise, some labels may remain set to `true` incorrectly.

pkg/registration/spoke/managedcluster/namespace_controller.go
elgnay: I missed this part. If the managed namespace controller becomes a reconciler for the managed cluster controller, it won’t act immediately when a managed namespace is deleted from the managed cluster, since it doesn’t watch namespaces. Should it be watching the managed namespace as well?

pkg/registration/hub/managedcluster/managed_namespace_controller.go
qiujian16: can we do it when the agent find the cluster is in deleting state? It cannot be guaranteed anyway, since agent may miss the event when deleting.

pkg/registration/spoke/managedcluster/namespace_controller.go
qiujian16: I do not feel we need at the starting point. When we have labels on the namespace that we are able to filter the namespace by agent, we can do that in managedcluster controller.

pkg/registration/hub/managedcluster/managed_namespace_controller.go
elgnay: Removed.

pkg/registration/hub/managedcluster/managed_namespace_controller.go
elgnay: Yes. Moved the logic to spoke side.

pkg/registration/hub/managedcluster/managed_namespace_controller.go
elgnay: Updated.

pkg/registration/spoke/managedcluster/status_controller.go
```diff
@@ -116,6 +126,12 @@ func newManagedClusterStatusController(
 				reservedClusterClaimSuffixes: reservedClusterClaimSuffixes,
 				aboutLister:                  propertyInformer.Lister(),
 			},
+			&managedNamespaceReconcile{
+				hubClusterSetLabel:   GetHubClusterSetLabel(hubHash),
```
qiujian16, line 130: will hubHash be an issue when we switch to a new hub?

elgnay: After switching to the new hub, the label `clusterset.open-cluster-management.io/<old_hub_hash>` remains set to `true`, which could cause cleanup issues if managed namespace deletion is supported in the future.

qiujian16 (commented):
/approve
/lgtm

PR #1269, opened 2025-11-26
🌱 use SDK basecontroller for better logging.

qiujian16:
## Summary

basecontroller in library-go does not support contextual logging. The basecontroller in sdk-go has implemented contextual logging.

## Related issue(s)

Fixes #


pkg/addon/templateagent/registration.go
```diff
@@ -471,8 +474,14 @@ func (a *CRDTemplateAgentAddon) createPermissionBinding(clusterName, addonName,
 		binding.OwnerReferences = []metav1.OwnerReference{*owner}
 	}
 
+	// TODO(qiujian16) this should have ctx passed to build the wrapper
+	recorderWrapper := commonrecorder.NewEventsRecorderWrapper(
```
zhujian7, line 478: So with this `events.NewContextualLoggingEventRecorder`, there will be only log printed, but no k8s event created, right?

pkg/registration/spoke/spokeagent.go
```diff
@@ -148,7 +147,7 @@ func (o *SpokeAgentConfig) RunSpokeAgent(ctx context.Context, controllerContext
 		informers.NewSharedInformerFactory(spokeKubeClient, 10*time.Minute),
 		clusterv1informers.NewSharedInformerFactory(spokeClusterClient, 10*time.Minute),
 		aboutinformers.NewSharedInformerFactory(aboutClusterClient, 10*time.Minute),
-		controllerContext.EventRecorder,
+		events.NewContextualLoggingEventRecorder("registration-agent"),
```
zhujian7, line 150: Same question here, with this recorder, only log, no k8s event created?

zhujian7:
@qiujian16  I am a little confused about the title. From my understanding, most content in this PR is about `recorder`, but why the main benefit is better logging? Could you add some description to the PR to give some details?

pkg/addon/templateagent/registration.go
qiujian16: yes, we actually have not used these events for debugging today. And we could implement k8s event if needed in the future.

qiujian16:
> @qiujian16 I am a little confused about the title. From my understanding, most content in this PR is about `recorder`, but why the main benefit is better logging? Could you add some description to the PR to give some details?

added some explanation. The main reason is basecontroller in sdk-go has contextual logging implemented.

pkg/common/recorder/event_recorder_wrapper.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (e *EventsRecorderWrapper) WithComponentSuffix(suffix string) librarygoevents.Recorder {
+	e.recorder.WithComponentSuffix(suffix)
+	return e
+}
+
+func (e *EventsRecorderWrapper) WithContext(ctx context.Context) librarygoevents.Recorder {
+	eCopy := *e
+	eCopy.ctx = ctx
+	return e
```
xuezhaojun: Should return eCopy?
```
func (e *EventsRecorderWrapper) WithContext(ctx context.Context) librarygoevents.Recorder {
	eCopy := *e
	eCopy.ctx = ctx
	return &eCopy    
}
```

qiujian16: good catch, fixed

pkg/common/recorder/event_recorder_wrapper.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (e *EventsRecorderWrapper) Warning(reason, message string) {
+	e.recorder.Warning(e.ctx, reason, message)
+}
+
+func (e *EventsRecorderWrapper) Warningf(reason, messageFmt string, args ...interface{}) {
+	e.recorder.Warningf(e.ctx, reason, messageFmt, args...)
+}
+
+func (e *EventsRecorderWrapper) ForComponent(componentName string) librarygoevents.Recorder {
```
xuezhaojun, line 43: Seems here we have similar issues for these 2 functions, expected to change to the following way:

```
func (e *EventsRecorderWrapper) ForComponent(componentName string) librarygoevents.Recorder {
    return &EventsRecorderWrapper{
        recorder: e.recorder.ForComponent(componentName),
        ctx:      e.ctx,
    }
}

func (e *EventsRecorderWrapper) WithComponentSuffix(suffix string) librarygoevents.Recorder {
    return &EventsRecorderWrapper{
        recorder: e.recorder.WithComponentSuffix(suffix),
        ctx:      e.ctx,
    }
}
```

Based on the interface defination, `e.recorder.ForComponent(componentName)` and `e.recorder.WithComponentSuffix(suffix)` returns new recorder instead of modifying the old recorder. So the old way will make new recorder not used.

xuezhaojun: interface defination:
```
	// ForComponent allows to fiddle the component name before sending the event to sink.
	// Making more unique components will prevent the spam filter in upstream event sink from dropping
	// events.
	ForComponent(componentName string) Recorder

	// WithComponentSuffix is similar to ForComponent except it just suffix the current component name instead of overriding.
	WithComponentSuffix(componentNameSuffix string) Recorder
```

qiujian16: this is intentional, we need this to return a new one to avoid race if multiple components call at the same time.

See implementation here https://github.com/openshift/library-go/blob/master/pkg/operator/events/recorder_logging.go#L35-L38

xuezhaojun: But the current implementation doesn't use the new recorder, because:

```
func (e *EventsRecorderWrapper) ForComponent(componentName string) librarygoevents.Recorder {
	e.recorder.ForComponent(componentName)
	return e
}
```

equals to 

```
func (e *EventsRecorderWrapper) ForComponent(componentName string) librarygoevents.Recorder {
	newRecorder := e.recorder.ForComponent(componentName) 
	return e // if we return `e` here, the above newRecorder is throwed.
}
```

haoqing0110:
LGTM

pkg/common/recorder/event_recorder_wrapper.go
qiujian16: ah right, updated.

xuezhaojun:
Also look good to me!

PR #1282, opened 2025-12-05
✨   Update transition time for Applied + StatusFeedbackSynced

annelaucg:
## Summary
Updated based on 1292. 

1. Add observedGeneration for Applied and StatusFeedback Synced type condition
2. Update lastTransitionTime for Applied when observedGeneration updated 
3. Update lastTransitionTime for StatusFeedback when values in the status feedback response are updated.

## Related issue(s)
#1292  

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Conditions now remove stale entries when a resource's generation changes, preventing outdated duplicates.
  * Applied and feedback conditions record the manifest's observed generation so condition visibility matches manifest generation.
  * Status feedback now refreshes LastTransitionTime when feedback values change, ensuring time-based transitions reflect real updates.

* **Tests**
  * New and extended tests validate ObservedGeneration and LastTransitionTime updates for applied, available, and feedback conditions.

### DIFF
--- a/pkg/work/helper/helpers.go
+++ b/pkg/work/helper/helpers.go
@@ -129,6 +129,14 @@ func MergeStatusConditions(conditions []metav1.Condition, newConditions []metav1
 
 	merged = append(merged, conditions...)
 	for _, condition := range newConditions {
+		for i := range merged {
+			if merged[i].Type == condition.Type {
+				if merged[i].ObservedGeneration != condition.ObservedGeneration {
+					meta.RemoveStatusCondition(&merged, condition.Type)
+				}
+				break
+			}
+		}
 		// merge two conditions if necessary
 		meta.SetStatusCondition(&merged, condition)
 	}
--- a/pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
+++ b/pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
@@ -136,6 +136,10 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
 
 		// Read status of the resource according to feedback rules.
 		values, statusFeedbackCondition := c.getFeedbackValues(obj, option)
+		valuesChanged := !equality.Semantic.DeepEqual(manifest.StatusFeedbacks.Values, values)
+		if valuesChanged {
+			meta.RemoveStatusCondition(manifestConditions, statusFeedbackCondition.Type)
+		}
 		meta.SetStatusCondition(manifestConditions, statusFeedbackCondition)
 		manifestWork.Status.ResourceStatus.Manifests[index].StatusFeedbacks.Values = values
 
--- a/pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
+++ b/pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
@@ -77,7 +77,7 @@ func (m *manifestworkReconciler) reconcile(
 		}
 
 		// Add applied status condition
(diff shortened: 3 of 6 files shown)


pkg/work/helper/helpers.go
```diff
@@ -129,8 +129,35 @@ func MergeStatusConditions(conditions []metav1.Condition, newConditions []metav1
...
+			prevObservedGeneration int64
+		)
+		for i := range merged {
+			if merged[i].Type == condition.Type {
+				hadPrevious = true
+				prevStatus = merged[i].Status
+				prevObservedGeneration = merged[i].ObservedGeneration
+				break
+			}
+		}
```
qiujian16, line 139: could we make it simpler? e.g

1. getCondition from merged.
2. if a condition is found, check the generation, if generation is changed, remove the old one from the merged
3. set newCondition.

pkg/work/spoke/controllers/statuscontroller/availablestatus_controller.go
```diff
@@ -136,6 +136,14 @@ func (c *AvailableStatusController) syncManifestWork(ctx context.Context, origin
 
 		// Read status of the resource according to feedback rules.
 		values, statusFeedbackCondition := c.getFeedbackValues(obj, option)
+		valuesChanged := !equality.Semantic.DeepEqual(manifest.StatusFeedbacks.Values, values)
+		if valuesChanged {
+			meta.RemoveStatusCondition(manifestConditions, statusFeedbackCondition.Type)
```
qiujian16, line 141: same here, I think you just need to remove condition when value is changed.

qiujian16 (commented):
We will need some integration tests:

1. update work, check observedGeneration and lastTransitionTime is updated
2. create a work, update fields of the applied resource. Check lastTransitionTime is updated.

qiujian16:
cc @elgnay @zhujian7

pkg/work/helper/helpers.go
```diff
@@ -129,8 +129,35 @@ func MergeStatusConditions(conditions []metav1.Condition, newConditions []metav1
...
 		meta.SetStatusCondition(&merged, condition)
+
+		// If status is unchanged but ObservedGeneration advanced, bump LastTransitionTime to reflect a new apply
+		if hadPrevious &&
+			prevStatus == condition.Status &&
+			condition.ObservedGeneration != 0 &&
+			prevObservedGeneration != condition.ObservedGeneration {
+			for i := range merged {
+				if merged[i].Type == condition.Type {
+					merged[i].LastTransitionTime = metav1.NewTime(time.Now())
```
elgnay: The `LastTransitionTime` of the apply condition records the last successful or failed apply time, which is usually the initial apply time. With this change, we are able to trace the apply time for the latest generation, but we lose the information about the first apply time. I'm not quite sure which timestamp is more valuable.

annelaucg: I see, what is the use case to track the first apply time? If we have iterated multiple times, wouldn't we want to track the applied time of the current iteration? What do you think?

annelaucg: @elgnay Any thoughts on the above comment? I'll update on the tests this week.

elgnay: There was a use case where users wanted to understand the delay of the ManifestWork API, calculated as the first applied time minus the creation time. With this change, it can no longer be used to measure the rough API delay. Nevertheless, I think the change is fine.

qiujian16:
signoff needs to be handled. Others looks good

/approve

/assign @elgnay

zhujian7:
Hi @annelaucg please signoff your commits, thx!

PR #1310, opened 2025-12-30
:warning:  Add duplicate manifest detection in ManifestWork webhook validation

xuezhaojun:
## Summary

This PR adds validation to detect and reject duplicate manifests in ManifestWork resources. A manifest is considered duplicate when it has the same apiVersion, kind, namespace, and name as another manifest in the same ManifestWork.

This prevents issues where duplicate manifests with different specs can cause state inconsistency, as the Work Agent applies manifests sequentially and later entries would overwrite earlier ones.

### Implementation Improvements (based on review feedback)

1. **Use `sets.Set[string]` instead of `map[string]int`** - Cleaner code for tracking seen manifests
2. **Use lightweight `manifestMeta` struct instead of `unstructured.Unstructured`** - Reduces computation resource by only decoding the metadata fields needed for duplicate detection (TypeMeta + ObjectMeta)
3. **Simplified error message** - Now shows `duplicate manifest at index %d: %s` without the first occurrence index

### Test Coverage

- Unit tests for duplicate detection scenarios
- E2E test added: `Should respond bad request when creating a manifestwork with duplicate manifests`

## Files Changed

- `pkg/work/webhook/common/validator.go` - Core validation logic
- `pkg/work/webhook/common/validator_test.go` - Unit tests
- `test/e2e/work_webhook_test.go` - E2E test for duplicate manifest validation

## Summary by CodeRabbit

* **New Features**
  * Manifest validation now detects duplicate manifests and reports duplicate index/key; validator instance is configurable with a manifest limit.

* **Bug Fixes**
  * Validation errors more clearly surface parsing, missing-field, and duplicate issues; duplicate manifest submissions are rejected.

* **Tests**
  * Added unit and end-to-end tests covering duplicate manifests, same-name different-namespace/kind, missing fields, limits, and unique-manifest scenarios.

### DIFF
--- a/pkg/work/webhook/common/validator.go
+++ b/pkg/work/webhook/common/validator.go
@@ -1,10 +1,13 @@
 package common
 
 import (
+	"encoding/json"
 	"fmt"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	workv1 "open-cluster-management.io/api/work/v1"
 )
@@ -33,11 +36,25 @@ func (m *Validator) ValidateManifests(manifests []workv1.Manifest) error {
 		return fmt.Errorf("the size of manifests is %v bytes which exceeds the %v limit", totalSize, m.limit)
 	}
 
-	for _, manifest := range manifests {
+	// Track seen manifests to detect duplicates
+	seen := sets.New[string]()
+
+	for i, manifest := range manifests {
 		err := validateManifest(manifest.Raw)
 		if err != nil {
 			return err
 		}
+
+		// Check for duplicate manifests
+		info, err := extractManifestInfo(manifest.Raw)
+		if err != nil {
+			return fmt.Errorf("failed to extract metadata from manifest at index %d: %w", i, err)
+		}
+
+		if seen.Has(info.key) {
+			return fmt.Errorf("duplicate manifest for resource %s/%s with resource type %s", info.namespace, info.name, info.gvk)
+		}
+		seen.Insert(info.key)
 	}
 
 	return nil
@@ -61,4 +78,27 @@ func validateManifest(manifest []byte) error {
 	}
 
 	return nil
+}
+
+// manifestInfo contains the metadata needed for duplicate detection and error messages.
+type manifestInfo struct {
+	key       string // unique key for duplicate detection: apiVersion/kind/namespace/name
+	name      string
+	namespace string
+	gvk       string // apiVersion.kind format for error messages
+}
+
+// extractManifestInfo extracts metadata from a manifest for duplicate detection.
+func extractManifestInfo(manifest []byte) (*manifestInfo, error) {
+	meta := &metav1.PartialObjectMetadata{}
+	if err := json.Unmarshal(manifest, meta); err != nil {
+		return nil, err
+	}
+
+	return &manifestInfo{
+		key:       fmt.Sprintf("%s/%s/%s/%s", meta.APIVersion, meta.Kind, meta.Namespace, meta.Name),
+		name:      meta.Name,
+		namespace: meta.Namespace,
+		gvk:       fmt.Sprintf("%s.%s", meta.APIVersion, meta.Kind),
+	}, nil
 }
--- a/pkg/work/webhook/common/validator_test.go
+++ b/pkg/work/webhook/common/validator_test.go
@@ -32,6 +32,40 @@ func newManifest(size int) workv1.Manifest {
 	manifest.Raw = objectStr
 	return manifest
 }
+
+func newManifestWithNameAndNamespace(name, namespace string) workv1.Manifest {
+	obj := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "v1",
+			"kind":       "ConfigMap",
+			"metadata": map[string]interface{}{
+				"namespace": namespace,
+				"name":      name,
+			},
+		},
+	}
+	objectStr, _ := obj.MarshalJSON()
+	manifest := workv1.Manifest{}
+	manifest.Raw = objectStr
+	return manifest
+}
+
+func newManifestWithKind(name, namespace, kind string) workv1.Manifest {
+	obj := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "v1",
+			"kind":       kind,
+			"metadata": map[string]interface{}{
+				"namespace": namespace,
+				"name":      name,
+			},
+		},
+	}
+	objectStr, _ := obj.MarshalJSON()
+	manifest := workv1.Manifest{}
+	manifest.Raw = objectStr
+	return manifest
+}
 func Test_Validator(t *testing.T) {
 	cases := []struct {
 		name          string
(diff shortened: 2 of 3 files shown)


pkg/work/webhook/common/validator.go
```diff
@@ -33,11 +33,27 @@ func (m *Validator) ValidateManifests(manifests []workv1.Manifest) error {
 		return fmt.Errorf("the size of manifests is %v bytes which exceeds the %v limit", totalSize, m.limit)
 	}
 
-	for _, manifest := range manifests {
+	// Track seen manifests to detect duplicates
+	seen := make(map[string]int) // key -> first occurrence index
```
qiujian16: use Sets[string] would make code cleaner

pkg/work/webhook/common/validator.go
```diff
@@ -62,3 +78,19 @@ func validateManifest(manifest []byte) error {
 
 	return nil
 }
+
+// extractManifestKey extracts a unique identifier from a manifest.
+// The key format is: apiVersion/kind/namespace/name
+// For cluster-scoped resources, namespace will be empty.
+func extractManifestKey(manifest []byte) (string, error) {
+	unstructuredObj := &unstructured.Unstructured{}
```
qiujian16: instead of deserialize the whole object, would decoding to metaObject reduce some computation resource?

pkg/work/webhook/common/validator.go
xuezhaojun: Done. Changed to use `sets.New[string]()` for cleaner code.

pkg/work/webhook/common/validator.go
xuezhaojun: Done. Defined a lightweight `manifestMeta` struct with only `TypeMeta` and `ObjectMeta`, using `json.Unmarshal` instead of `unstructured.Unstructured`.

pkg/work/webhook/common/validator.go
```diff
@@ -33,11 +43,27 @@ func (m *Validator) ValidateManifests(manifests []workv1.Manifest) error {
...
+			// If we cannot extract the key, skip duplicate check
+			// (validateManifest would have caught invalid manifests)
+			continue
+		}
+
+		if seen.Has(key) {
+			return fmt.Errorf("duplicate manifest at index %d: %s", i, key)
+		}
+		seen.Insert(key)
 	}
```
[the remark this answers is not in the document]
xuezhaojun: Good catch! Changed to return the error instead of silently skipping.

pkg/work/webhook/common/validator.go
```diff
@@ -33,11 +43,27 @@ func (m *Validator) ValidateManifests(manifests []workv1.Manifest) error {
...
+		if err != nil {
+			// If we cannot extract the key, skip duplicate check
+			// (validateManifest would have caught invalid manifests)
+			continue
+		}
+
+		if seen.Has(key) {
+			return fmt.Errorf("duplicate manifest at index %d: %s", i, key)
+		}
+		seen.Insert(key)
```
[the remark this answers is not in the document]
xuezhaojun: This was intentional based on maintainer feedback. @qiujian16 suggested using `sets.Set[string]` for cleaner code, which means we trade the first occurrence index for simpler implementation. The PR description has been updated to reflect this change.

pkg/work/webhook/common/validator.go
qiujian16: can you just use https://pkg.go.dev/k8s.io/apimachinery@v0.35.0/pkg/apis/meta/v1#PartialObjectMetadata

qiujian16:
the PR title should start with ⚠️

xuezhaojun:
> the PR title should start with ⚠️

Fixed.

xuezhaojun:
@qiujian16  please take another look! Thanks!

qiujian16:
https://github.com/open-cluster-management-io/ocm/pull/1310#discussion_r2684694723 this is still outstanding.

pkg/work/webhook/common/validator.go
xuezhaojun: @qiujian16  Done. Changed to use `metav1.PartialObjectMetadata` directly. That's a new knowledge to me! Thanks for the information!

pkg/work/webhook/common/validator.go
```diff
@@ -33,11 +36,25 @@ func (m *Validator) ValidateManifests(manifests []workv1.Manifest) error {
...
 		}
+
+		// Check for duplicate manifests
+		key, err := extractManifestKey(manifest.Raw)
+		if err != nil {
+			return fmt.Errorf("failed to extract key from manifest at index %d: %w", i, err)
+		}
+
+		if seen.Has(key) {
+			return fmt.Errorf("duplicate manifest at index %d: %s", i, key)
```
qiujian16: Last comment

could we have a more readable error message since this will be returned to user. Something like
duplicate manifest for resource namespace/name with resource type v1.pod?

xuezhaojun: Done. Changed error message to: `duplicate manifest for resource namespace/name with resource type apiVersion.kind`

qiujian16:
/approve
/lgtm
/hold

hold until we make sure that nothing in prod breaks.

pkg/work/webhook/common/validator.go
```diff
@@ -62,3 +79,26 @@ func validateManifest(manifest []byte) error {
...
+
+// extractManifestInfo extracts metadata from a manifest for duplicate detection.
+func extractManifestInfo(manifest []byte) (*manifestInfo, error) {
+	meta := &metav1.PartialObjectMetadata{}
+	if err := json.Unmarshal(manifest, meta); err != nil {
+		return nil, err
+	}
+
+	return &manifestInfo{
+		key:       fmt.Sprintf("%s/%s/%s/%s", meta.APIVersion, meta.Kind, meta.Namespace, meta.Name),
```
lucasponce, line 99: Note that API machinery tends to return these fields empty on purpose.

Unsure if within the context of the execution of this flow we have a guarantee that these are populated.

I.e. OCM uses the same name for multiple resources (ManagedCluster / ManifestWork), if these fiealds are not populated that may introduce misleading results.

xuezhaojun:
@lucasponce Thanks for pointing this out! You're absolutely right - relying on these fields without validation could lead to incorrect duplicate detection if they were empty.

I've added defensive validation in `extractManifestInfo` to ensure `apiVersion`, `kind`, and `name` are all present before using them for duplicate detection:

```go
if meta.APIVersion == "" {
    return nil, fmt.Errorf("apiVersion must be set in manifest")
}
if meta.Kind == "" {
    return nil, fmt.Errorf("kind must be set in manifest")
}
if meta.Name == "" {
    return nil, fmt.Errorf("name must be set in manifest")
}
```

This ensures that different resource types won't be incorrectly identified as duplicates when metadata fields are missing.

Added corresponding test cases as well. Please take another look when you have a chance.

lucasponce:
Thanks, the changes look good, I'm validating that during the build of the ManifestWork, all "TypeMeta" are set in origin from OCM.

As those should be populated and this may cause any regression, I'll update the PR with my findings.

qiujian16:
I do not think this validation is needed, we already have this on API spec for each manifest
```
// +kubebuilder:validation:EmbeddedResource
```

xuezhaojun:
@qiujian16 You're right. The `+kubebuilder:validation:EmbeddedResource` marker on the Manifest field already ensures that apiVersion and kind are validated at the API server level before the webhook is invoked. The additional defensive validation was redundant.

I've reverted the commit. Thanks for the clarification!

PR #1318, opened 2026-01-07
🐛 Fix work rolebinding cleanup when hubAcceptsClient is set to false

ericogr:
# Summary

When a ManagedCluster is denied (`hubAcceptsClient: false`) while ManifestWorks exist, and those ManifestWorks are deleted, the work RoleBinding gets stuck with the finalizer `cluster.open-cluster-management.io/manifest-work-cleanup` and is never fully deleted.

**Steps to Reproduce:**

1. Accept a ManagedCluster (set `hubAcceptsClient: true`)
2. Create ManifestWorks in the cluster namespace
3. Deny the cluster (set `hubAcceptsClient: false`)
4. Delete all ManifestWorks
5. Wait and observe: work RoleBinding remains with `deletionTimestamp` and finalizer

**Expected:** Work RoleBinding should be completely deleted after ManifestWorks are removed
**Actual:** Work RoleBinding stuck indefinitely with finalizer `cluster.open-cluster-management.io/manifest-work-cleanup`

**Root Cause:**

The controller only attempted RBAC cleanup during the `accepted → denied` transition. Once the status was set to `HubAccepted=false`, subsequent reconciliations would return early without attempting cleanup, even after ManifestWorks were deleted.

In pkg/registration/hub/managedcluster/controller.go (lines 164-168):

```go
if !meta.IsStatusConditionTrue(..., HubAccepted) {
    return nil  // ← Bug: never retries cleanup
}
```

## Solution

When hubAcceptsClient is turned to false, we now remove the registration/work cluster rolebindings in the denial path before setting the condition, requeueing if needed so the work rolebinding’s finalizer can clear once manifestworks are gone. The integration test was updated to create a ManifestWork before flipping hubAcceptsClient from true to false and only delete it afterward, then assert the work rolebinding disappears, covering the stuck-finalizer case.

In pkg/registration/hub/managedcluster/controller.go:

```go
...
		// Moved RoleBinding removal for registration-agent and work-agent before setting the condition.
		err := c.removeClusterRbac(ctx, syncCtx, managedClusterName, managedCluster.Spec.HubAcceptsClient)
		if errors.Is(err, requeueError) {
			syncCtx.Queue().AddAfter(managedClusterName, requeueError.RequeueTime)
			return nil
...

		meta.SetStatusCondition(&newManagedCluster.Status.Conditions, metav1.Condition{
			Type:    v1.ManagedClusterConditionHubAccepted,
			Status:  metav1.ConditionFalse,
			Reason:  "HubClusterAdminDenied",
			Message: "Denied by hub cluster admin",
		})

...
```

## Testing

Added integration test `managedcluster_accept_with_manifestwork_test.go` that reproduces the scenario:
1. Accept cluster
2. Create ManifestWorks
3. Deny cluster
4. Delete ManifestWorks
5. Verify work RoleBinding is completely deleted

**Before fix:** Test passes - Because the check only verifies the presence of the finalizer, not the deletion of the RoleBinding.
**After fix:** Test passes - RoleBinding properly deleted

Run test:
```bash
make test-registration-integration ARGS="-ginkgo.focus='ManagedCluster set hubAcceptsClient from true to false'"
```

## Summary by CodeRabbit

* **Bug Fixes**
  * RBAC cleanup now runs earlier in the denial workflow, avoids duplicate cleanup calls, and better aggregates retryable and non-retryable errors to improve reliability.

* **Tests**
  * Hardened managed-cluster acceptance tests: added manifest-work handling, switched final checks to absence/IsNotFound polling, and extended wait timing for cleanup verification.

### DIFF
--- a/pkg/registration/hub/managedcluster/controller.go
+++ b/pkg/registration/hub/managedcluster/controller.go
@@ -188,6 +188,10 @@ func (c *managedClusterController) sync(ctx context.Context, syncCtx factory.Syn
 			return nil
 		}
 
+		if err != nil {
+			errs = append(errs, err)
+		}
+
 		if err = c.hubDriver.Cleanup(ctx, managedCluster); err != nil {
 			errs = append(errs, err)
 		}


qiujian16:
thanks, looks good to me, will leave to @elgnay for final review

/approve

PR #1319, opened 2026-01-07
🐛 fail fast on MTU mismatch before SelfSubjectAccessReview

Shrushti72:
Fix klusterlet TLS handshake timeouts caused by MTU mismatch
Klusterlet bootstrap and certificate rotation can fail when the managed
cluster MTU exceeds the hub path MTU due to large TLS handshake packets
during SelfSubjectAccessReview calls.
This change adds a pre-flight MTU validation to fail fast with a clear,
actionable error message instead of a silent TLS handshake timeout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added network MTU detection and validation that surfaces errors when detected MTU exceeds expected limits ( >1500 ), preventing connectivity and operational issues caused by incompatible network settings.

### DIFF
--- a/pkg/common/helpers/ssar.go
+++ b/pkg/common/helpers/ssar.go
@@ -1,3 +1,4 @@
+
 package helpers
 
 import (
@@ -164,6 +165,9 @@ func generateSelfSubjectAccessReviews(resource authorizationv1.ResourceAttribute
 					Verb:        verb,
 				},
 			},
+
+
+			
 		})
 	}
 	return reviews
--- a/pkg/common/helpers/ssar.go
+++ b/pkg/common/helpers/ssar.go
@@ -2,21 +2,58 @@ package helpers
 
 import (
 	"context"
+	"fmt"
+	"net"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
+func detectLocalMTU() (int, error) {
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return 0, err
+	}
+
+	maxMTU := 0
+	for _, iface := range ifaces {
+		if iface.Flags&net.FlagUp == 0 {
+			continue
+		}
+		if iface.MTU > maxMTU {
+			maxMTU = iface.MTU
+		}
+	}
+
+	if maxMTU == 0 {
+		return 0, fmt.Errorf("unable to detect MTU")
+	}
+
+	return maxMTU, nil
+}
+
 func CreateSelfSubjectAccessReviews(
 	ctx context.Context,
 	kubeClient kubernetes.Interface,
-	selfSubjectAccessReviews []authorizationv1.SelfSubjectAccessReview) (bool, *authorizationv1.SelfSubjectAccessReview, error) {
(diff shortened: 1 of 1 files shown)


Shrushti72:
actually pr was not verified ,it was creating issue so many commits was pushed because of this

pkg/common/helpers/ssar.go
```diff
@@ -1,22 +1,60 @@
...
+}
+
 func CreateSelfSubjectAccessReviews(
 	ctx context.Context,
 	kubeClient kubernetes.Interface,
-	selfSubjectAccessReviews []authorizationv1.SelfSubjectAccessReview) (bool, *authorizationv1.SelfSubjectAccessReview, error) {
+	selfSubjectAccessReviews []authorizationv1.SelfSubjectAccessReview,
+) (bool, *authorizationv1.SelfSubjectAccessReview, error) {
+
+	mtu, err := detectLocalMTU()
```
qiujian16, line 43: I would put this mtu detection in a separated func instead of calling inside CreateSelfSubjectAccessReviews

pkg/common/helpers/ssar.go
```diff
@@ -1,22 +1,60 @@
...
+	if err == nil && mtu > 1500 {
```
qiujian16, line 44: what if local MTU is larger than 1500 and remote is also larger?

pkg/common/helpers/ssar.go
```diff
@@ -1,22 +1,60 @@
...
+	"fmt"
+	"net"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
+func detectLocalMTU() (int, error) {
+	ifaces, err := net.Interfaces()
```
qiujian16, line 15: we need to clarify system permission it needs to execute this command.

qiujian16 (commented):
I would like this detection to be opt-in and disabled by default. It might break in existing env otherwise

pkg/common/helpers/ssar.go
Shrushti72: > we need to clarify system permission it needs to execute this command.

Good catch. net.Interfaces() relies on standard OS networking APIs and does not require elevated privileges in typical container environments, but this should be documented.

pkg/common/helpers/ssar.go
Shrushti72: > what if local MTU is larger than 1500 and remote is also larger?

You’re right, local MTU alone is only a heuristic.
The intent here is not to guarantee a mismatch, but to conservatively reduce SSAR payload size only when jumbo frames are detected locally and the opt-in flag is enabled.

pkg/common/helpers/ssar.go
Shrushti72: > I would put this mtu detection in a separated func instead of calling inside CreateSelfSubjectAccessReviews

I agree that MTU detection shouldn’t be embedded directly inside CreateSelfSubjectAccessReviews.
I’ll refactor this if wanted so MTU detection lives in a separate helper and is passed in only when the feature is enabled.

Shrushti72:
> I would like this detection to be opt-in and disabled by default. It might break in existing env otherwise

I’ll gate the MTU-aware behavior behind an explicit flag e.g. env var or config option, and keep the default path unchanged when it’s disabled.

PR #1332, opened 2026-01-16
🐛 MWRS Rollout Mechanism - Progressing status conditions, True Wins

annelaucg:
## Summary
The rollout mechanism described [here](https://open-cluster-management.io/docs/concepts/work-distribution/manifestworkreplicaset/#rollout-mechanism) depends on Progressing status.

If multiple resources have the Progressing status, the `newAggregateConditionFromRules` will have False wins, this means that Progressing = False will be set earlier than intended and the rollout will continue once 1 of the resources are Completed. 

This is counter to what is expected that the MWRS will rollout when all resources are Completed. 

## Related issue(s)

Fixes #

## Release Notes

* **Bug Fixes**
  * Fixed the Progressing condition to correctly report True when any resource is actively progressing, improving status accuracy.

* **Tests**
  * Added comprehensive test coverage for Progressing status scenarios with multiple resources in different states.


haoqing0110:
LGTM

PR #1342, opened 2026-01-19
:bug:  requeue work to re-create resource when resource is not available

zhiweiyin318:
## Related issue(s)

Fixes #
fix the issue that the resource applied by manifestWork cannot be re-created in time (have to wait for 4-6 min) after the resource is deleted. 

## Release Notes

* **Bug Fixes**
  * Improved reconciliation timing for applied but unavailable resources, enabling faster recovery.

* **Tests**
  * Added integration test for resource deletion and recreation scenarios.


mikeshng (commented):
@coderabbitai review

PR #1343, opened 2026-01-20
✨ add token driver for addon registration

elgnay:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Token-based addon authentication path added alongside CSR, with runtime selection and automatic fallback.
  * New addon token configuration to control token expiration.

* **Improvements**
  * Registration flow enhanced to choose auth method per-addon and reduce unnecessary restarts via improved config hashing.
  * New CLI/Helm flags to expose addon auth type and token expiration.

* **Tests**
  * Expanded unit and integration tests covering token/CSR flows, rotation, transitions, and edge cases.

### DIFF
--- a/pkg/registration/register/aws_irsa/aws_irsa.go
+++ b/pkg/registration/register/aws_irsa/aws_irsa.go
@@ -3,11 +3,15 @@ package aws_irsa
 import (
 	"context"
 	"fmt"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/klog/v2"
 
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	operatorv1 "open-cluster-management.io/api/operator/v1"
@@ -16,6 +20,8 @@ import (
 
 	"open-cluster-management.io/ocm/pkg/common/helpers"
 	"open-cluster-management.io/ocm/pkg/registration/register"
+	"open-cluster-management.io/ocm/pkg/registration/register/csr"
+	"open-cluster-management.io/ocm/pkg/registration/register/token"
 )
 
 //TODO: Remove these constants in once we have the function fully implemented for the AWSIRSADriver
@@ -36,6 +42,15 @@ type AWSIRSADriver struct {
 	managedClusterRoleSuffix string
 
 	awsIRSAControl AWSIRSAControl
+
+	// addonClients holds the addon clients and informers
+	addonClients *register.AddOnClients
+
+	// tokenControl is used for token-based addon authentication
+	tokenControl token.TokenControl
+
+	// csrControl is used for CSR-based addon authentication
+	csrControl csr.CSRControl
 }
 
 func (c *AWSIRSADriver) Process(
@@ -98,7 +113,7 @@ func (c *AWSIRSADriver) ManagedClusterDecorator(cluster *clusterv1.ManagedCluste
 	return cluster
 }
 
-func (c *AWSIRSADriver) BuildClients(_ context.Context, secretOption register.SecretOption, bootstrap bool) (*register.Clients, error) {
+func (c *AWSIRSADriver) BuildClients(ctx context.Context, secretOption register.SecretOption, bootstrap bool) (*register.Clients, error) {
 	clients, err := register.BuildClientsFromSecretOption(secretOption, bootstrap)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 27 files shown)


pkg/registration/register/interface.go
```diff
@@ -81,9 +81,31 @@ type RegisterDriver interface {
...
+// AddonRegistrationOption provides complete configuration for addon registration,
+// including authentication method and access to driver options.
+// Registration type (KubeClient vs CustomSigner) is determined from SecretOption.Signer.
+type AddonRegistrationOption interface {
+	// GetAddonKubeClientRegistrationAuth returns the authentication method for addons with registration type KubeClient.
+	// Possible values are "csr" (default) and "token".
+	GetAddonKubeClientRegistrationAuth() string
+
+	// GetCSROption returns the CSR driver option
+	GetCSROption() interface{}
```
qiujian16: why this is not typed.

pkg/registration/register/interface.go
```diff
@@ -81,9 +81,31 @@ type RegisterDriver interface {
...
+
+	// GetTokenOption returns the token driver option
+	GetTokenOption() interface{}
```
qiujian16: ditto

pkg/registration/register/aws_irsa/aws_irsa.go
```diff
@@ -107,9 +123,78 @@ func (c *AWSIRSADriver) BuildClients(_ context.Context, secretOption register.Se
...
+
 	return clients, nil
 }
 
+func (c *AWSIRSADriver) Fork(addonName string, option register.AddonRegistrationOption, secretOption register.SecretOption) (register.RegisterDriver, error) {
+	// Determine registration type from signer name
+	isKubeClientType := secretOption.Signer == certificatesv1.KubeAPIServerClientSignerName
+
+	// Check if we should use token-based authentication for KubeClient registration type
+	if isKubeClientType && option.GetAddonKubeClientRegistrationAuth() == "token" {
```
qiujian16: this seems like a common part for each driver, can we have a common function to start tokenDriver?

pkg/registration/register/factory/options.go
```diff
@@ -45,6 +65,18 @@ func (s *Options) Validate() error {
 	}
 }
 
+func (s *Options) GetAddonKubeClientRegistrationAuth() string {
+	return s.AddonKubeClientRegistrationAuth
+}
+
+func (s *Options) GetCSROption() interface{} {
+	return s.CSROption
```
qiujian16, line 73: why this has to be interface?

pkg/registration/register/interface.go
elgnay: To avoid Circular Dependency Occurs.

The `csr` and `token` packages need to import the parent register package to use:
  - register.SecretOption
  - register.RegisterDriver interface
  - register.AddonRegistrationOption interface
  - register.AddOnClients
  - Utility functions like register.KubeConfigFromSecretOption()

  If `register/interface.go` imports `register/csr` or `register/token`, Go will detect the circular dependency and fail to compile.

pkg/registration/register/factory/options.go
elgnay: factory.Options.GetCSROption() cannot return a concrete type because it implements the register.AddonRegistrationOption interface which requires interface{} return type.

pkg/registration/register/interface.go
qiujian16: If so why the interface needs to return csrOption and tokenOption for any driver. If it has to be returned on interface level, I think the returned type should also be defined in interface.

elgnay: > If so why the interface needs to return csrOption and tokenOption for any driver.
The cluster driver can fork add-on drivers at runtime, so the add-on registration needs to carry options for both driver types. Because csrOption and tokenOption are driver-specific, it’s debatable whether they should be part of the interface.

pkg/registration/spoke/addon/registration_controller.go
```diff
@@ -127,14 +132,27 @@ func (c *addOnRegistrationController) syncAddOn(ctx context.Context, syncCtx fac
 		return nil
 	}
 
-	cachedConfigs := c.addOnRegistrationConfigs[addOnName]
-	configs, err := getRegistrationConfigs(addOn)
+	// Ensure driver field is set for kubeClient type registrations
+	updated, err := c.ensureDriver(ctx, addOn)
+	if err != nil {
```
qiujian16: if updated || err != nil {
  return err
}

qiujian16:
looks good in general

pkg/registration/register/aws_irsa/aws_irsa.go
elgnay: A common function provided.

pkg/registration/spoke/addon/registration_controller.go
elgnay: Updated as suggested.

Issue #1163, opened 2025-09-04
Support ClusterProfile credentials plugin


haoqing0110:
**Describe the enhancement**
A clear and concise description of what the enhancement request is.

Today OCM already supports ClusterProfile API from SIG Multi-Cluster  https://github.com/open-cluster-management-io/ocm/issues/247.

This issue is to support the ClusterProfile credentials plugin. More details refer to https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/5339-clusterprofile-plugin-credentials.

PR #1351, opened 2026-01-26
🌱 sync clusterprofile based on managedclusterset and managedclustersetbinding

morvencao:
## Summary

refactor clusterprofile controller to sync nanagedclusters to clusterprofiles based on managedclustersets and managedclustersetbindings instead of using a hardcoded namespace.

## Related issue(s)

refer to https://github.com/open-cluster-management-io/ocm/issues/1163

## Summary by CodeRabbit

* **New Features**
  * Replaced the previous monolithic cluster-profile controller with two controllers: a lifecycle controller that creates/deletes ClusterProfiles based on bindings and cluster-set membership, and a status controller that syncs labels and status from managed clusters (deduplication and namespace scoping preserved).

* **Tests**
  * Added extensive unit and integration tests (queue-key logic, lifecycle reconciliation, status sync, end-to-end scenarios) and enabled the ClusterProfile feature gate in integration runs.
  * Removed the prior monolithic controller unit tests.

### DIFF
--- a/pkg/registration/hub/clusterprofile/lifecycle_controller.go
+++ b/pkg/registration/hub/clusterprofile/lifecycle_controller.go
@@ -27,11 +27,12 @@ import (
 	v1beta2 "open-cluster-management.io/api/cluster/v1beta2"
 	clustersdkv1beta2 "open-cluster-management.io/sdk-go/pkg/apis/cluster/v1beta2"
 	"open-cluster-management.io/sdk-go/pkg/basecontroller/factory"
+
+	"open-cluster-management.io/ocm/pkg/registration/hub/managedclustersetbinding"
 )
 
 const (
 	ClusterProfileManagerName = "open-cluster-management"
-	byClusterSet              = "by-clusterset"
 )
 
 // clusterProfileLifecycleController manages ClusterProfile creation and deletion
@@ -52,15 +53,6 @@ type clusterProfileLifecycleController struct {
 	clusterProfileLister     cplisterv1alpha1.ClusterProfileLister
 }
 
-// indexByClusterSet is the indexer function for ManagedClusterSetBinding by ClusterSet name
-func indexByClusterSet(obj interface{}) ([]string, error) {
-	binding, ok := obj.(*v1beta2.ManagedClusterSetBinding)
-	if !ok {
-		return []string{}, fmt.Errorf("obj is supposed to be a ManagedClusterSetBinding, but is %T", obj)
-	}
-	return []string{binding.Spec.ClusterSet}, nil
-}
-
 // NewClusterProfileLifecycleController creates a controller that manages ClusterProfile lifecycle
 func NewClusterProfileLifecycleController(
 	clusterInformer informerv1.ManagedClusterInformer,
@@ -69,13 +61,8 @@ func NewClusterProfileLifecycleController(
 	clusterProfileClient cpclientset.Interface,
 	clusterProfileInformer cpinformerv1alpha1.ClusterProfileInformer) factory.Controller {
 
-	// Add indexer for efficient lookup of bindings by clusterset
-	err := clusterSetBindingInformer.Informer().AddIndexers(cache.Indexers{
-		byClusterSet: indexByClusterSet,
-	})
-	if err != nil {
-		utilruntime.HandleError(err)
-	}
+	// Note: ByClusterSetIndex indexer is already added by managedclustersetbinding controller,
+	// so we don't need to add it again here. Informers are shared across controllers.
 
 	c := &clusterProfileLifecycleController{
 		clusterLister:            clusterInformer.Lister(),
@@ -190,7 +177,7 @@ func (c *clusterProfileLifecycleController) registerProfileEventHandler(
 
 // getBindingsByClusterSet efficiently retrieves all bindings for a given clusterset using the indexer
 func (c *clusterProfileLifecycleController) getBindingsByClusterSet(clusterSetName string) ([]*v1beta2.ManagedClusterSetBinding, error) {
-	objs, err := c.clusterSetBindingIndexer.ByIndex(byClusterSet, clusterSetName)
+	objs, err := c.clusterSetBindingIndexer.ByIndex(managedclustersetbinding.ByClusterSetIndex, clusterSetName)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/registration/hub/clusterprofile/status_controller.go
+++ b/pkg/registration/hub/clusterprofile/status_controller.go
@@ -208,11 +208,8 @@ func (c *clusterProfileStatusController) updateClusterProfile(
 
 	// Patch labels (patcher handles change detection)
 	_, err = profilePatcher.PatchLabelAnnotations(ctx, newProfile, newProfile.ObjectMeta, existing.ObjectMeta)
-	if err != nil {
-		return err
-	}
 
-	return nil
+	return err
 }
 
 func syncLabelsFromCluster(profile *cpv1alpha1.ClusterProfile, cluster *v1.ManagedCluster) {
--- a/pkg/registration/hub/clusterprofile/lifecycle_controller_queuekeys_test.go
+++ b/pkg/registration/hub/clusterprofile/lifecycle_controller_queuekeys_test.go
@@ -13,7 +13,19 @@ import (
 	clusterinformers "open-cluster-management.io/api/client/cluster/informers/externalversions"
 	v1 "open-cluster-management.io/api/cluster/v1"
 	v1beta2 "open-cluster-management.io/api/cluster/v1beta2"
+
+	"open-cluster-management.io/ocm/pkg/registration/hub/managedclustersetbinding"
 )
+
+// indexByClusterSet is a test helper for indexing ManagedClusterSetBindings by ClusterSet name.
+// In production, this indexer is registered by the managedclustersetbinding controller.
+func indexByClusterSet(obj interface{}) ([]string, error) {
+	binding, ok := obj.(*v1beta2.ManagedClusterSetBinding)
+	if !ok {
+		return []string{}, fmt.Errorf("obj is supposed to be a ManagedClusterSetBinding, but is %T", obj)
+	}
+	return []string{binding.Spec.ClusterSet}, nil
+}
 
 func TestClusterToQueueKeys(t *testing.T) {
 	cases := []struct {
@@ -681,7 +693,7 @@ func TestClusterToQueueKeys(t *testing.T) {
 			// Add indexer for bindings
 			bindingInformer := clusterInformers.Cluster().V1beta2().ManagedClusterSetBindings()
 			err := bindingInformer.Informer().AddIndexers(cache.Indexers{
-				byClusterSet: indexByClusterSet,
+				managedclustersetbinding.ByClusterSetIndex: indexByClusterSet,
 			})
 			if err != nil {
 				t.Fatal(err)
@@ -944,7 +956,7 @@ func TestClusterSetToQueueKeys(t *testing.T) {
 			// Add indexer for bindings
 			bindingInformer := clusterInformers.Cluster().V1beta2().ManagedClusterSetBindings()
 			err := bindingInformer.Informer().AddIndexers(cache.Indexers{
-				byClusterSet: indexByClusterSet,
+				managedclustersetbinding.ByClusterSetIndex: indexByClusterSet,
 			})
 			if err != nil {
 				t.Fatal(err)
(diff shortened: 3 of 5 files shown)


pkg/registration/hub/clusterprofile/controller.go
```diff
@@ -25,146 +32,435 @@ import (
...
+	clusterSetInformer clusterinformerv1beta2.ManagedClusterSetInformer,
+	clusterSetBindingInformer clusterinformerv1beta2.ManagedClusterSetBindingInformer,
 	clusterProfileClient cpclientset.Interface,
 	clusterProfileInformer cpinformerv1alpha1.ClusterProfileInformer) factory.Controller {
+
+	// Add indexer for ManagedClusterSetBinding by ClusterSet name
+	if err := clusterSetBindingInformer.Informer().AddIndexers(cache.Indexers{
+		bindingsByClusterSet: indexBindingByClusterSet,
+	}); err != nil {
+		klog.Errorf("Failed to add indexer for ManagedClusterSetBinding: %v", err)
```
qiujian16: use contextual logging, maybe just use HandleRuntimeErrorWithContext

morvencao: code updated.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
```diff
@@ -0,0 +1,324 @@
...
+// clusterSetToQueueKeys maps a ManagedClusterSet to all namespaces that have bindings to it
+func (c *clusterProfileLifecycleController) clusterSetToQueueKeys(obj runtime.Object) []string {
+	clusterSet, ok := obj.(*v1beta2.ManagedClusterSet)
+	if !ok {
+		return nil
+	}
+
+	// Find all bindings that reference this clusterset
+	// Constraint: binding.Name == binding.Spec.ClusterSet
+	allBindings, err := c.clusterSetBindingLister.List(labels.Everything())
```
qiujian16: it is better to have a index on this

pkg/registration/hub/clusterprofile/lifecycle_controller.go
```diff
@@ -0,0 +1,324 @@
...
+
+	// Find all clustersets containing this cluster
+	clusterSets, err := clustersdkv1beta2.GetClusterSetsOfCluster(cluster, c.clusterSetLister)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to get clustersets for cluster %s: %w", cluster.Name, err))
+		return nil
+	}
+
+	// List all bindings
+	allBindings, err := c.clusterSetBindingLister.List(labels.Everything())
```
qiujian16: same setting a index on clustersetbinding for clusterset will make this much easier.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
```diff
@@ -0,0 +1,324 @@
...
+			"clusterset", clusterSet.Name,
+			"clusterCount", len(clusters))
+
+		// Mark these clusters as desired
+		for _, cluster := range clusters {
+			// Skip clusters that are being deleted
+			if !cluster.DeletionTimestamp.IsZero() {
+				continue
+			}
+			desiredClusters[cluster.Name] = true
```
qiujian16: A easier approach is to put his into a set, and use Diff to compare the two sets.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	profilesDeleted := 0
+
+	// Create missing profiles
+	for clusterName := range desiredClusters {
+		if _, exists := existingClusters[clusterName]; !exists {
+			// Create profile
+			err := c.createClusterProfile(ctx, namespace, clusterName)
+			if err != nil {
+				logger.Error(err, "Failed to create ClusterProfile", "cluster", clusterName)
+				utilruntime.HandleError(fmt.Errorf("failed to create ClusterProfile %s/%s: %w", namespace, clusterName, err))
```
qiujian16: we should add err into errs and return aggregated err, so we can return faster with err.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
```diff
@@ -0,0 +1,324 @@
...
+func (c *clusterProfileLifecycleController) createClusterProfile(ctx context.Context, namespace, clusterName string) error {
+	logger := klog.FromContext(ctx)
+
+	clusterProfile := &cpv1alpha1.ClusterProfile{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      clusterName,
+			Namespace: namespace,
+			Labels: map[string]string{
+				cpv1alpha1.LabelClusterManagerKey:       ClusterProfileManagerName,
+				ClusterProfileForManagedClusterLabelKey: clusterName,
```
qiujian16: use this https://github.com/open-cluster-management-io/api/blob/main/cluster/v1/types.go#L271

pkg/registration/hub/clusterprofile/status_controller.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func (c *clusterProfileStatusController) profileToQueueKey(obj runtime.Object) []string {
+	profile, ok := obj.(*cpv1alpha1.ClusterProfile)
+	if !ok {
+		return nil
+	}
+
+	// Use cluster-name label if available
+	if clusterName, ok := profile.Labels[ClusterProfileForManagedClusterLabelKey]; ok {
```
qiujian16: we do not need to use as key, but this should be a filter that we only care about clusterprofile with this label exists, also we need to check if clustermanager is ocm.

pkg/registration/hub/clusterprofile/status_controller.go
```diff
@@ -0,0 +1,227 @@
...
+	// Sync labels
+	syncLabelsFromCluster(newProfile, cluster)
+
+	// Patch labels if modified
+	labelsModified := !reflect.DeepEqual(existing.Labels, newProfile.Labels)
+
+	if labelsModified {
+		_, err := profilePatcher.PatchLabelAnnotations(ctx, newProfile, newProfile.ObjectMeta, existing.ObjectMeta)
+		if err != nil {
+			return err
```
qiujian16: we should return here even without err as long as it is updated, there could be conflict otherwise

pkg/registration/hub/clusterprofile/status_controller.go
```diff
@@ -0,0 +1,227 @@
...
+		*cpv1alpha1.ClusterProfile, cpv1alpha1.ClusterProfileSpec, cpv1alpha1.ClusterProfileStatus](
+		c.clusterProfileClient.ApisV1alpha1().ClusterProfiles(existing.Namespace))
+
+	newProfile := existing.DeepCopy()
+
+	// Sync labels
+	syncLabelsFromCluster(newProfile, cluster)
+
+	// Patch labels if modified
+	labelsModified := !reflect.DeepEqual(existing.Labels, newProfile.Labels)
```
qiujian16: this is not needed, patcher will handle this.

pkg/registration/hub/clusterprofile/status_controller.go
```diff
@@ -0,0 +1,227 @@
...
+			return err
+		}
+		logger.V(4).Info("Updated profile labels")
+	}
+
+	// Sync status
+	syncStatusFromCluster(newProfile, cluster)
+
+	// Patch status
+	statusModified := !reflect.DeepEqual(existing.Status, newProfile.Status)
```
qiujian16: this is not needed, patcher will handle it.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
```diff
@@ -0,0 +1,324 @@
...
+	c := &clusterProfileLifecycleController{
+		clusterLister:           clusterInformer.Lister(),
+		clusterSetLister:        clusterSetInformer.Lister(),
+		clusterSetBindingLister: clusterSetBindingInformer.Lister(),
+		clusterProfileClient:    clusterProfileClient,
+		clusterProfileLister:    clusterProfileInformer.Lister(),
+	}
+
+	return factory.New().
+		WithInformersQueueKeysFunc(c.clusterToQueueKeys, clusterInformer.Informer()).
```
qiujian16: note this could be quite noisy, we only care about create/delete

pkg/registration/hub/clusterprofile/lifecycle_controller.go
```diff
@@ -0,0 +1,324 @@
...
+		WithInformersQueueKeysFunc(c.clusterSetToQueueKeys, clusterSetInformer.Informer()).
+		WithInformersQueueKeysFunc(c.bindingToQueueKey, clusterSetBindingInformer.Informer()).
+		WithInformersQueueKeysFunc(c.profileToQueueKey, clusterProfileInformer.Informer()).
```
qiujian16: same here, only need to care create/delete event

pkg/registration/hub/clusterprofile/lifecycle_controller.go
morvencao: added registerClusterEventHandler to catch managedcluster create/delete events only.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
morvencao: added registerProfileEventHandler to catch clusterprofile create/delete events only.

pkg/registration/hub/clusterprofile/status_controller.go
morvencao: removed this.

pkg/registration/hub/clusterprofile/status_controller.go
morvencao: removed this.

pkg/registration/hub/clusterprofile/status_controller.go
morvencao: updated this: If status has been updated, return early - labels will be updated in next reconcile.

pkg/registration/hub/clusterprofile/status_controller.go
```diff
@@ -0,0 +1,262 @@
...
+	if updated {
+		return err
+	}
+
+	// Status wasn't updated, now safe to patch labels
+	syncLabelsFromCluster(newProfile, cluster)
+
+	// Patch labels (patcher handles change detection)
+	_, err = profilePatcher.PatchLabelAnnotations(ctx, newProfile, newProfile.ObjectMeta, existing.ObjectMeta)
+	if err != nil {
```
qiujian16: nit, just return err

pkg/registration/hub/clusterprofile/lifecycle_controller.go
```diff
@@ -0,0 +1,451 @@
...
+	clusterlisterv1beta2 "open-cluster-management.io/api/client/cluster/listers/cluster/v1beta2"
+	v1 "open-cluster-management.io/api/cluster/v1"
+	v1beta2 "open-cluster-management.io/api/cluster/v1beta2"
+	clustersdkv1beta2 "open-cluster-management.io/sdk-go/pkg/apis/cluster/v1beta2"
+	"open-cluster-management.io/sdk-go/pkg/basecontroller/factory"
+)
+
+const (
+	ClusterProfileManagerName = "open-cluster-management"
+	byClusterSet              = "by-clusterset"
```
qiujian16: there is already an index (byClusterSet) added in clustersetbinding controller, which I think you can make it public and reuse. We do not need to add another indexer here.

pkg/registration/hub/clusterprofile/status_controller.go
morvencao: code updated.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
morvencao: yes, updated this to use byClusterSet index from managedclustersetbinding controller. We don't need to add it again here as informers are shared across controllers, right?

pkg/registration/hub/clusterprofile/status_controller.go
morvencao: added UnionFilter to filter clusterprofile with labels.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
morvencao: nice, code updated to use the label key.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
morvencao: code updated by using aggregated error.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
morvencao: code updated to use set and diff two sets to create/delete clusterprofiles.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
morvencao: added indexer to improve performance.

pkg/registration/hub/clusterprofile/lifecycle_controller.go
morvencao: added indexer to improve performance.

PR #1371, opened 2026-02-02
feat(chart): add importerConfiguration

guilhem:
## Summary

add a way to customize importerConfiguration in clusterm-manager chart

## Summary by CodeRabbit

* **New Features**
  * Added support for importer configuration in the cluster manager, allowing users to optionally configure importer settings as needed.


zhiweiyin318:
@guilhem  we also need to define  importerConfiguration in the values.yaml, I think that's why the ut is failed. 
```yaml
clusterManager:
  registrationConfiguration:
    importerConfiguration: {}
```

Issue #1323, opened 2026-01-09
PlacementDecision events are recorded even when there are no changes


haoqing0110:
**Describe the bug**
  The scheduling controller records Kubernetes Events for PlacementDecision even when there are no actual changes, causing unnecessary event accumulation.

  Current Behavior

  File: pkg/placement/controllers/scheduling/scheduling_controller.go
  Function: createOrUpdatePlacementDecision()
  Lines: 711-754

  updated, err := placementDecisionPatcher.PatchStatus(ctx, newPlacementDecision, newPlacementDecision.Status, existPlacementDecision.Status)
  if updated {
      return err  // Early return when status is updated
  }

  // Return value ignored - cannot tell if labels actually changed
  _, err = placementDecisionPatcher.PatchLabelAnnotations(ctx, newPlacementDecision, newPlacementDecision.ObjectMeta, existPlacementDecision.ObjectMeta)
  if err != nil {
      return err
  }

  // ❌ PROBLEM: Events are ALWAYS recorded here, even when nothing changed
  c.eventsRecorder.Eventf(..., "DecisionUpdate", "DecisionUpdated", ...)
  c.eventsRecorder.Eventf(..., "ScoreUpdate", "ScoreUpdated", ...)

**To Reproduce**

**Expected behavior**

**Environment ie: OCM version, Kubernetes version and provider:**

**Additional context**

PR #1376, opened 2026-02-09
:bug: improve event recording logic and test maintainbility

haoqing0110:
- Fix event recording to only record when decisions actually change
- Extract recordDecisionEvents helper to avoid duplicated event recording
- Add comprehensive event recording tests (100% coverage for new logic)
- Refactor test helpers to reduce code duplication:
  - Add newTestSchedulingController to create test controllers
  - Add assertion helpers (assertClusterSetBindingNames, assertClusterSetNames, assertClusterNames)
  - Add verifyEvents and containsEventType for event verification

🤖 Generated with [Claude Code](https://claude.ai/code)

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/1323 

## Summary by CodeRabbit

* **Bug Fixes**
  * Events now reliably reflect placement decision status changes; label-only updates no longer trigger status events.
  * Score events are emitted only when scores exist, use deterministic ordering, and are truncated when too long.
  * Event attribution updated to reference the specific placement decision for both decision and score events.

* **Tests**
  * Added comprehensive event-recording tests covering create/update/label/warning scenarios.
  * Refactored test setup with helpers to reduce duplication and simplify assertions.

### DIFF
--- a/pkg/placement/controllers/scheduling/scheduling_controller.go
+++ b/pkg/placement/controllers/scheduling/scheduling_controller.go
@@ -712,46 +712,60 @@ func (c *schedulingController) createOrUpdatePlacementDecision(
 	// If status has been updated, just return, this is to avoid conflict when updating the label later.
 	// Labels and annotations will still be updated in next reconcile.
 	if updated {
+		// Record events when status is updated
+		c.recordDecisionEvents(placement, existPlacementDecision, clusterScores, status)
 		return err
 	}
+	// Do not record events when label is updated
 	_, err = placementDecisionPatcher.PatchLabelAnnotations(ctx, newPlacementDecision, newPlacementDecision.ObjectMeta, existPlacementDecision.ObjectMeta)
-	if err != nil {
-		return err
-	}
+	return err
+}
 
-	// update the event with warning
+// recordDecisionEvents records DecisionUpdate and ScoreUpdate events for placement decision
+func (c *schedulingController) recordDecisionEvents(
+	placement *clusterapiv1beta1.Placement,
+	placementDecision *clusterapiv1beta1.PlacementDecision,
+	clusterScores PrioritizerScore,
+	status *framework.Status,
+) {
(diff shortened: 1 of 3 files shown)


pkg/placement/controllers/scheduling/scheduling_controller.go
```diff
@@ -712,46 +712,65 @@ func (c *schedulingController) createOrUpdatePlacementDecision(
...
 	}
-	_, err = placementDecisionPatcher.PatchLabelAnnotations(ctx, newPlacementDecision, newPlacementDecision.ObjectMeta, existPlacementDecision.ObjectMeta)
+	labelUpdated, err := placementDecisionPatcher.PatchLabelAnnotations(ctx, newPlacementDecision, newPlacementDecision.ObjectMeta, existPlacementDecision.ObjectMeta)
 	if err != nil {
 		return err
 	}
 
-	// update the event with warning
+	// Record events only when there are changes (status updated or labels updated)
+	if labelUpdated {
```
qiujian16: why do we need to record event when label is changed? Is anything different in event under this case?

haoqing0110: PlacementDecision label contains the index and decision group info; there's a possibility that select clusters do not change, but only the label (decision group name) changes, and this should be traced by event.

qiujian16: but we do not record decision group name in the event I think?

haoqing0110: Yes, we might need a different event to record the label changes.

qiujian16: if we do not have different events for now, could we just avoid event recording when label changes.

haoqing0110: Got it, removed the logic to call recordDecisionEvents when label changes.

PR #1394, opened 2026-02-21
🐛 Use generateName when creating ManifestWorks from ManifestWorkReplicaSet

### LINKED ISSUE #1380: should use generateName to create manifestwork from manifestworkreplicaset
**Describe the bug**
manifestworkreplicaset currently uses its name to create related manifestwork, which could conflict among mwrs from different namespace. It should use generateName intead.

**To Reproduce**
create two manifestwork in different namespace with the same name.

**Expected behavior**

**Additional context**

shivansh-source:
## Description

Fixes a naming conflict when ManifestWorkReplicaSets create
ManifestWorks.

Previously, ManifestWorks were created using the
ManifestWorkReplicaSet name directly. This could cause collisions
when multiple ManifestWorkReplicaSets from different namespaces
targeted the same managed cluster namespace.

This PR switches ManifestWork creation to use `generateName`,
allowing the API server to assign unique names.

## Changes
- Use `GenerateName` instead of `Name` when creating ManifestWorks
- Update deletion logic to locate ManifestWorks via labels
- Update unit tests to use label-based lookup instead of fixed names

## Related Issue
Fixes #1380

## Testing
- Updated existing unit tests
- `go test ./pkg/work/...` passes locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup when clusters are removed by listing and filtering placement-scoped manifest works to ensure all relevant works are deleted.

* **Behavior Changes**
  * Manifest work creation now uses server-generated names with a replica-set-derived prefix, altering final resource identifiers.

* **Tests**
  * Updated tests and helpers to use listing-based checks and to align fixtures with the new naming behavior.

### DIFF
--- a/pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_test.go
+++ b/pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_test.go
@@ -776,7 +776,7 @@ if len(works) != 0 {
 	works, err = listWorksByMWRS( context.TODO(), fWorkClient, "cls2", mwrSet.Namespace, mwrSet.Name, "place-test2",)
     if err != nil { t.Fatal(err) }
       if len(works) != 1 {
-       t.Fatalf("Expected ManifestWork for current placement to exist, but got %d", len(works))
+       t.Fatalf("Expected ManifestWork for current placement to exist, but got %d, works: %v", len(works), works)
     }
 currentMW := works[0]
 	// Verify the placement label is correct
@@ -864,7 +864,7 @@ func TestDeployReconcileWithMultiplePlacementChanges(t *testing.T) {
 		t.Fatal(err)
 	}
 	if len(works) != 1 {
-		t.Fatalf("Expected ManifestWork for placement2 to exist, but got error: %v", len(works))
+		t.Fatalf("Expected ManifestWork for placement2 to exist, but got error: %d, works: %v", len(works), works)
 	}
 	currentMW2 := works[0]
 	assert.Equal(t, currentMW2.Labels[workapiv1alpha1.ManifestWorkReplicaSetPlacementNameLabelKey], "place-test2")
@@ -874,7 +874,7 @@ func TestDeployReconcileWithMultiplePlacementChanges(t *testing.T) {
 		t.Fatalf("Expected ManifestWork for placement3 to exist, but got error: %v", err)
 	}
 	if len(works) != 1 {
-		t.Fatalf("Expected ManifestWork for placement3 to exist, but got error: %v", len(works))
+		t.Fatalf("Expected ManifestWork for placement3 to exist, but got error: %d, works: %v", len(works), works)
 	}
 	currentMW3 := works[0]
 	assert.Equal(t,currentMW3.Labels[workapiv1alpha1.ManifestWorkReplicaSetPlacementNameLabelKey], "place-test3")


pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_test.go
```diff
@@ -846,23 +850,34 @@ func TestDeployReconcileWithMultiplePlacementChanges(t *testing.T) {
...
 
-	currentMW3, err := fWorkClient.WorkV1().ManifestWorks("cls3").Get(context.TODO(), mwrSet.Name, metav1.GetOptions{})
+	works, err = listWorksByMWRS(context.TODO(), fWorkClient, "cls3", mwrSet.Namespace, mwrSet.Name, "place-test3")
 	if err != nil {
 		t.Fatalf("Expected ManifestWork for placement3 to exist, but got error: %v", err)
 	}
-	assert.Equal(t, currentMW3.Labels[workapiv1alpha1.ManifestWorkReplicaSetPlacementNameLabelKey], "place-test3")
+	if len(works) != 1 {
+		t.Fatalf("Expected ManifestWork for placement3 to exist, but got error: %v", err)
+	}
```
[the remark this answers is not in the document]
shivansh-source: i think printing the error would be better for maintainers than printing the val

mikeshng:
CC @youngbupark @annelaucg @qiujian16

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
```diff
@@ -381,7 +384,7 @@ func CreateManifestWork(
 
 	return &workv1.ManifestWork{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      mwrSet.Name,
+			GenerateName: mwrSet.Name + "-",
```
annelaucg, line 387: This generateName will generate a random hash behind right? How will we guarnatee its the namespace thats included in the MW name? 

I see this logic [below](https://github.com/open-cluster-management-io/ocm/pull/1394/changes#diff-20aa11fe464a70582c5b84caba83b1e06bf4d2027dc9629a3a324c3e787f5d2bR93) in the helper.go , but wondering how it ties to this change?

shivansh-source: generateName only ensures uniqueness

shivansh-source: and controller logic already relies on labels for lookup & deletion so thats why

shivansh-source: The name assignment in helper.go is only used for unit tests. 
@annelaucg

shivansh-source: > This generateName will generate a random hash behind right? How will we guarnatee its the namespace thats included in the MW name?
> 
Yes — `generateName` lets the API server append a unique suffix to the  ManifestWork name to avoid collisions

annelaucg: Got it, so it ensures unique-ness, but we will not append the namespace to the MW name. Any concerns about human readability of the knowing which ManifestWork was from which namespace?

qiujian16, line 387: I do not think adding "-" is necessary

qiujian16: I think the name/namespace of mwrs is set in the label.

qiujian16 (commented):
looks good in general, you might need to update integration/e2e accordingly

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
shivansh-source: got it will do the changes

shivansh-source:
@qiujian16  can i still finish this or u did the work ??

qiujian16:
#1421 is actually based on this PR and some refactor. Please take a look at that one. I think we can merge that one if it is ok which will be co-authored by both of us.

shivansh-source:
Sure sir no problem

PR #1409, opened 2026-03-04
:seedling: Replace legacy envtest setup with sdk-go shared script

xuezhaojun:
## Summary

Replace the legacy `ensure-kubebuilder-tools` Makefile target with the new shared `envtest-setup` target from [sdk-go](https://github.com/open-cluster-management-io/sdk-go/tree/main/ci/envtest).

### Changes

- **`test/integration-test.mk`**: Removed the legacy `ensure-kubebuilder-tools` target that manually installed `setup-envtest` via `go install` and downloaded kubebuilder binaries from Google Cloud Storage. Replaced it with the new `envtest-setup` target that uses the centralized `ensure-envtest.sh` script from sdk-go, which automatically detects the K8s version from `go.mod` and manages binary downloads.
- **`Makefile`**: Updated `test-unit` dependency from `ensure-kubebuilder-tools` to `envtest-setup`.
- All integration test targets (`test-registration-integration`, `test-work-integration`, `test-placement-integration`, `test-registration-operator-integration`, `test-addon-integration`, `test-cloudevents-work-mqtt-integration`, `test-cloudevents-work-grpc-integration`) now depend on `envtest-setup` instead of `ensure-kubebuilder-tools`.
- Simplified `clean-integration-test` by removing the `/tmp/kubebuilder` cleanup (no longer used).
- **`test/integration/work/conditionrules_test.go`**: Fixed all 3 Job status update blocks to comply with K8s 1.34 stricter Job status validation. The new envtest auto-detects K8s version from `go.mod` (`k8s.io/api v0.34.3` → K8s 1.34), which requires:
  - `status.startTime` for active/finished Jobs
  - `status.completionTime` for completed Jobs
  - `SuccessCriteriaMet=True` condition before setting `Complete=True`

  Affected test cases:
  - `should return well known completed condition rule`
  - `wildcard option should match all jobs`
  - `matched multi options only the first works`

### Motivation

The old approach used `storage.googleapis.com/kubebuilder-tools` which is deprecated. The new sdk-go shared script provides:
- Automatic K8s version detection from `go.mod`
- Centralized maintenance across all OCM repositories
- Fallback version resolution when exact versions aren't available

## Release Notes

* **Chores**
  * Streamlined test environment setup workflow for improved test execution reliability and maintainability.

### DIFF
--- a/test/integration/work/conditionrules_test.go
+++ b/test/integration/work/conditionrules_test.go
@@ -3,6 +3,7 @@ package work
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
@@ -97,6 +98,7 @@ var _ = ginkgo.Describe("ManifestWork Condition Rules", func() {
 				[]metav1.ConditionStatus{metav1.ConditionTrue}, eventuallyTimeout, eventuallyInterval)
 
 			// Update Job status on spoke
+			now := metav1.Now()
 			gomega.Eventually(func() error {
 				job, err := spokeKubeClient.BatchV1().Jobs(clusterName).Get(context.Background(), "job1", metav1.GetOptions{})
 				if err != nil {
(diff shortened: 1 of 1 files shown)


test/integration/work/conditionrules_test.go
```diff
@@ -405,7 +420,15 @@ var _ = ginkgo.Describe("ManifestWork Condition Rules", func() {
 				job.Status.Active = 0
 				job.Status.Ready = ptr.To(int32(0))
 				job.Status.Succeeded = 1
+				job.Status.StartTime = &now
```
qiujian16, line 423: why we need to change the test, is it because the apiserver upgrade?

xuezhaojun: Yes, that's the reason, latest apiserver has more strict check on job's condition, more details have been added in the PR description.

PR #1410, opened 2026-03-04
:seedling:  support install hosted addon via installStrategy

zhiweiyin318:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **New Features**
  * ManagedCluster annotations with addon-prefix are now propagated into ManagedClusterAddOns; hosting-cluster annotation is preserved for hosted clusters.
  * New controller watches ManagedCluster annotation changes and syncs addon-prefix annotations to cluster add-ons.

* **Tests**
  * Expanded unit and integration tests to verify annotation propagation, controller sync behavior, and reconciler wiring for managed clusters.

### DIFF
--- a/pkg/addon/controllers/addonannotation/controller.go
+++ b/pkg/addon/controllers/addonannotation/controller.go
@@ -1,0 +1,197 @@
+package addonannotation
+
+import (
+	"context"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	addonv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	addonv1alpha1client "open-cluster-management.io/api/client/addon/clientset/versioned"
+	addoninformerv1alpha1 "open-cluster-management.io/api/client/addon/informers/externalversions/addon/v1alpha1"
+	addonlisterv1alpha1 "open-cluster-management.io/api/client/addon/listers/addon/v1alpha1"
+	clusterinformersv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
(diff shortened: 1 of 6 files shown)


pkg/addon/controllers/addonmanagement/addon_install_reconciler.go
```diff
@@ -93,6 +105,18 @@ func (d *managedClusterAddonInstallReconciler) reconcile(
 	return cma, reconcileContinue, utilerrors.NewAggregate(errs)
 }
 
+// getHostingClusterName returns the hosting cluster name from the ManagedCluster's
+// addon.open-cluster-management.io/hosting-cluster-name annotation, or empty string if not set.
+func (d *managedClusterAddonInstallReconciler) getHostingClusterName(clusterName string, logger klog.Logger) string {
+	cluster, err := d.managedClusterLister.Get(clusterName)
+	if err != nil {
+		logger.Error(err, "failed to get cluster")
```
qiujian16: do we need to return error here? I think this check is not needed.

qiujian16:
please also confirm if the addon manager has the permission to update mca. Adding an e2e will be also helpful

zhiweiyin318:
> please also confirm if the addon manager has the permission to update mca. Adding an e2e will be also helpful

checked that addon manager has  the permission to update mca. 
e2e is added.

pkg/addon/controllers/addonannotation/controller.go
```diff
@@ -0,0 +1,197 @@
...
+	// Extract annotations with the addon prefix from the managed cluster
+	clusterAddonAnnotations := map[string]string{}
+	for k, v := range cluster.Annotations {
+		if strings.HasPrefix(k, addonv1alpha1.GroupName) {
+			clusterAddonAnnotations[k] = v
+		}
+	}
+
+	// List all addons in the cluster namespace
+	addons, err := c.managedClusterAddonLister.ManagedClusterAddOns(clusterName).List(labels.Everything())
```
haoqing0110, line 129: Do you expect the annotation to be synced to all the ManagedClusterAddOns (including manual created) or only the installStrategy selected mca?

haoqing0110: For example installStrategy selects cluster1, cluster2, and user manually creates mca for cluster3.

haoqing0110: Plz ignore this question. This won't happen since the install controller will remove the manually created mca.

haoqing0110:
LGTM

PR #1423, opened 2026-03-10
:bug: set tokenrequest UID to empty to avoid conflict with the ServiceAccount UID validation

zhiweiyin318:
## Related issue(s)

Fixes #

## Release Notes

* **Bug Fixes**
  * Resolved ServiceAccount UID validation conflicts affecting token request operations in Kubernetes 1.34+. Token requests will now process successfully without encountering validation errors when working with updated Kubernetes cluster environments.

### DIFF
--- a/pkg/server/services/tokenrequest/tokenreqeust.go
+++ b/pkg/server/services/tokenrequest/tokenreqeust.go
@@ -56,6 +56,9 @@ func (t *TokenRequestService) HandleStatusUpdate(ctx context.Context, evt *cloud
 
 	switch eventType.Action {
 	case types.CreateRequestAction:
+		// Clear the UID to avoid conflict with the ServiceAccount UID validation
+		// introduced by the TokenRequestServiceAccountUIDValidation feature gate (k8s 1.34+).
+		tokenRequest.UID = ""
 		// Create a token for the service account
 		tokenResponse, err := t.client.CoreV1().ServiceAccounts(tokenRequest.Namespace).CreateToken(ctx, tokenRequest.Name, tokenRequest, metav1.CreateOptions{})
 		if err != nil {


skeeey:
LGTM

pkg/server/services/tokenrequest/tokenreqeust.go
```diff
@@ -56,6 +56,9 @@ func (t *TokenRequestService) HandleStatusUpdate(ctx context.Context, evt *cloud
 
 	switch eventType.Action {
 	case types.CreateRequestAction:
+		// Clear the UID to avoid conflict with the ServiceAccount UID validation
+		// introduced by the TokenRequestServiceAccountUIDValidation feature gate (k8s 1.34+).
+		tokenRequest.UID = ""
```
qiujian16, line 61: how could this have an UID? I do not think we set this.

skeeey: we use this uid as cloudevent resource id, and we use this id on the agent side to identified the subscribed resource (https://github.com/open-cluster-management-io/sdk-go/blob/main/pkg/cloudevents/clients/serviceaccount/client.go#L116)

qiujian16: but when you create the tokenrequest, you should not be able to set a uid.

skeeey: yes, we should not set it when creating with kube api, but we need a way to record this uid for a tokenrequest when we Encode/Decode it, for now, we use the tokenrequest UID to record it, if we don't use UID, we may need other way to record it, maybe an annotation in the matedata, wdyt?

qiujian16: should this be done in the decoder?

skeeey: if we remove the uid in the decoder, we will lose this uid in the tokenrequest (we still need a way to record this on the decoded tokenrequest obj), but we need set this uid to tokenresponse

PR #1494, opened 2026-04-22
:seedling: Add CertRotationController support for PlacementDebugServer TLS

Randy424:
## Summary

Adds TLS serving certificate lifecycle management for the PlacementDebugServer feature gate using the existing CertRotationController infrastructure, replacing the earlier annotation-based approach.

When `PlacementDebugServer` is enabled in ClusterManager's `PlacementConfiguration.FeatureGates`:
- CertRotationController creates a `placement-debug-serving-cert` target rotation signed by the existing OCM self-signed CA (30-day validity, auto-rotated at 80%)
- The serving cert is mounted into the debug-server sidecar at `/var/run/secrets/serving-cert/` with `optional: true`
- Library-go's `hasServiceServingCerts()` detects and uses the cert automatically — no CLI flags needed

When the feature gate is disabled (default):
- The serving cert secret is deleted if it exists
- The target rotation is removed from the rotation map
- All other cert rotations (signer, CA bundle, webhooks) are unaffected

### Design Decisions

- **CertRotationController over annotation injection**: Uses the same self-signed CA infrastructure as webhooks and gRPC, keeping TLS management consistent across all OCM components. No dependency on platform-specific service-CA controllers.
- **`optional: true` on cert volume**: Allows the pod to start during the timing window before the cert is first generated. Library-go falls back to self-signed certificates and logs a warning.
- **RBAC `resourceNames`**: Added `placement-debug-serving-cert` to the named secrets list in the cluster-manager ClusterRole, required for get/update/patch/delete operations.

## Files Changed

| File | Change |
|------|--------|
| `manifests/config.go` | Added `PlacementDebugServerEnabled` and `PlacementServingCertSecret` to `HubConfig` |
| `manifests/.../placement/deployment.yaml` | Conditional serving-cert volume and mount in debug-server sidecar |
| `pkg/.../certrotation_controller.go` | PlacementDebug target rotation lifecycle (create when enabled, delete when disabled, cleanup on ClusterManager deletion) |
| `pkg/.../certrotation_controller_test.go` | `TestCertRotationPlacementDebug` — Enable and Disable cases |
| `pkg/.../clustermanager_controller.go` | Wires `PlacementDebugServerEnabled` and `PlacementServingCertSecret` into HubConfig |
| `pkg/.../options.go` | Adds informer for `placement-debug-serving-cert` secret |
| `pkg/operator/helpers/helpers.go` | `PlacementDebugServerEnabled()` helper |
| `pkg/operator/helpers/queuekey.go` | `PlacementDebugServingCertSecret` and `PlacementDebugService` constants |
| `deploy/.../cluster_role.yaml` (x2) | Added `placement-debug-serving-cert` to RBAC `resourceNames` |

## Test plan

- [x] Unit tests: `TestCertRotationPlacementDebug` — Enable and Disable cases pass
- [x] `gofmt`, `go vet`, `verify` all clean
- [x] E2E: `e2e`, `e2e-hosted`, `e2e-singleton` suites pass (feature disabled by default — validates the cleanup-when-nonexistent path)

### DIFF
--- a/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
+++ b/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
@@ -227,6 +227,13 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	}
 	_, placementFeatureMsgs = helpers.ConvertToFeatureGateFlags("Placement", placementFeatureGates, ocmfeature.DefaultHubPlacementFeatureGates)
 	config.PlacementDebugServerEnabled = helpers.FeatureGateEnabled(placementFeatureGates, ocmfeature.DefaultHubPlacementFeatureGates, ocmfeature.PlacementDebugServer)
+	if config.PlacementDebugServerEnabled {
+		secretName := config.ClusterManagerName + "-placement-serving-cert"
+		config.ServingCertSecret = secretName
+		config.Annotations = map[string]string{
+			"service.beta.openshift.io/serving-cert-secret-name": secretName,
+		}
+	}
 
 	featureGateCondition := helpers.BuildFeatureCondition(registrationFeatureMsgs, workFeatureMsgs, addonFeatureMsgs, placementFeatureMsgs)
 
--- a/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
+++ b/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
@@ -229,8 +229,8 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	config.PlacementDebugServerEnabled = helpers.FeatureGateEnabled(placementFeatureGates, ocmfeature.DefaultHubPlacementFeatureGates, ocmfeature.PlacementDebugServer)
 	if config.PlacementDebugServerEnabled {
 		secretName := config.ClusterManagerName + "-placement-serving-cert"
-		config.ServingCertSecret = secretName
-		config.Annotations = map[string]string{
+		config.PlacementServingCertSecret = secretName
+		config.PlacementAnnotations = map[string]string{
 			"service.beta.openshift.io/serving-cert-secret-name": secretName,
 		}
 	}
--- a/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
+++ b/pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
@@ -228,11 +228,7 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	_, placementFeatureMsgs = helpers.ConvertToFeatureGateFlags("Placement", placementFeatureGates, ocmfeature.DefaultHubPlacementFeatureGates)
 	config.PlacementDebugServerEnabled = helpers.FeatureGateEnabled(placementFeatureGates, ocmfeature.DefaultHubPlacementFeatureGates, ocmfeature.PlacementDebugServer)
 	if config.PlacementDebugServerEnabled {
-		secretName := config.ClusterManagerName + "-placement-serving-cert"
-		config.PlacementServingCertSecret = secretName
-		config.PlacementAnnotations = map[string]string{
-			"service.beta.openshift.io/serving-cert-secret-name": secretName,
-		}
+		config.PlacementServingCertSecret = helpers.PlacementDebugServingCertSecret
 	}
 
 	featureGateCondition := helpers.BuildFeatureCondition(registrationFeatureMsgs, workFeatureMsgs, addonFeatureMsgs, placementFeatureMsgs)
--- a/deploy/cluster-manager/chart/cluster-manager/templates/cluster_role.yaml
+++ b/deploy/cluster-manager/chart/cluster-manager/templates/cluster_role.yaml
@@ -36,6 +36,7 @@ rules:
     - "work-driver-config"
     - "open-cluster-management-image-pull-credentials"
     - "grpc-server-serving-cert"
+    - "placement-debug-serving-cert"
     - "cluster-import-config"
 - apiGroups: [""]
   resources: ["secrets"]
--- a/deploy/cluster-manager/config/rbac/cluster_role.yaml
+++ b/deploy/cluster-manager/config/rbac/cluster_role.yaml
@@ -38,6 +38,7 @@ rules:
     - "work-driver-config"
     - "open-cluster-management-image-pull-credentials"
     - "grpc-server-serving-cert"
+    - "placement-debug-serving-cert"
     - "cluster-import-config"
 - apiGroups: [""]
   resources: ["secrets"]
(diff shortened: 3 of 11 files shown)


Randy424:
/cc @mikeshng @jnpacker 

Community PR

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -227,6 +227,13 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 	}
 	_, placementFeatureMsgs = helpers.ConvertToFeatureGateFlags("Placement", placementFeatureGates, ocmfeature.DefaultHubPlacementFeatureGates)
 	config.PlacementDebugServerEnabled = helpers.FeatureGateEnabled(placementFeatureGates, ocmfeature.DefaultHubPlacementFeatureGates, ocmfeature.PlacementDebugServer)
+	if config.PlacementDebugServerEnabled {
+		secretName := config.ClusterManagerName + "-placement-serving-cert"
+		config.PlacementServingCertSecret = secretName
+		config.PlacementAnnotations = map[string]string{
+			"service.beta.openshift.io/serving-cert-secret-name": secretName,
```
mikeshng: Hi @qiujian16 @zhujian7

Do you have any good ideas to handle this OCP specific case more generically?

qiujian16: why do we need to use ocp annotation, can we just let registration-operator to generate CA and serving cert which has been used for webhooks.

Alternatively, can we consider patching the service in the downstream?

Randy424: The current convention for ACM Console is ocp annotation.

Console would need to change its trust configuration if we switch to the CertRotationController (via registration-operator):

The Console trust chain currently works like this.
 *Client requests placement prediction from console backend* -> 
(then, in console backend) placementDebug.ts → getServiceAgent() → getServiceCACertificate() → getCertificate('service-ca.crt', ...) → reads /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt (the service agent expects the CA that signed the placement server's cert must be present in the service ca cert)

That file is the OCP service CA — it's automatically mounted into pods on OpenShift, and it's also the CA that signs certs created by the **service.beta.openshift.io/serving-cert-secret-name** annotation. So the current PR's approach mostly fits with what console already trusts.

If we switch to the CertRotationController there are a few obstacles we face to get things working again:
  - The placement cert would be signed by OCM's own self-signed CA, not OCP's service CA
  - That OCM CA lives in a ConfigMap (e.g., cluster-manager-placement-serving-cert-ca-bundle)
  - The OCP service-ca.crt file would not contain this CA
  - Console's HTTPS agent would reject the connection as untrusted

Console would need to either:
  1. Mount the OCM CA bundle ConfigMap as an additional volume and point to it (new cert path or merge with existing)
  2. Or add the OCM CA to the SERVICE_CA_CERT env var
  3. Or something else...?
  
  Thoughts @kcormier @mikeshng @jnpacker?

qiujian16: if this is the case, can we use ca of service account to sign the serving cert for debug server instead of adding the annotation?

zhiweiyin318:
As a neutral community, we should avoid tight functional coupling with any single product. 
In PR descriptions, we aim to avoid specific product names and versions whenever possible—a community guideline we have consistently followed over the years.
Thanks.

rokej:
ocm runs a cert rotation controller that:

- Maintains a self-signed signing CA in a Secret (signer-secret).

- Maintains a CA bundle ConfigMap (ca-bundle-configmap, key ca-bundle.crt) that aggregates still-valid signing CAs (so rotations do not break in-flight clients at once).

- For each target (registration/work/addon webhook secrets), mints TLS serving certs signed by the current CA, with SANs set to in-cluster Service DNS names (e.g. registration-webhook.<namespace>.svc).

I am wondering if this existing cert rotation controller can be used for the placement service as well.

Randy424:
Hi @qiujian16, @rokej, (sorry for the hold up!)--It seems you and @rokej had very similar initial ideas here, i.e. leveraging the CertRotationController. [I am testing this idea now](https://github.com/stolostron/console/pull/6090) with the required changes to our console component, and with the recently pushed changed to this branch. Will let you know how it goes, but in the meantime please feel free to comment further on the PRs recent changes. Thank you.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
KevinFCormier: Thanks for looking into this @qiujian16. For the upcoming release, we did just implement the service signing cert approach in the downstream, but @mikeshng and @jnpacker suggested we open a PR in the upstream as well to get your feedback and see if we could achieve this in a non-vendor specific way so that we could avoid diverging the downstream from the upstream.

When I originally requested this feature, I asked for an actual Kubernetes API similar to the SelfSubjectAccessReview, where POSTing to it gives you a response but does not create a new resource. I wonder if it might be better to pursue that as a long-term approach rather than using a certificate signed by the `ca-bundle-configmap`?

Randy424: Thoughts on this @rokej?

rokej: How would a cert expiration and rotation work with that design?

Randy424: I believe the cert expiration and rotation will work the same between these two strategies, both can be managed by the CertRotationController. In the later situation though, it's no longer a concern for the downstream.

The difference between the strategies lies in who consumes the CA bundle. With Kevin's idea (let's call it strategy B), we would enable the Kube API to consume and inject it into its APIService.spec.caBundle, ultimately this has the benefit of saving us from having to build a method downstream to access the CA bundle. Downstream it would required a method for cross-namespace interaction between console pods and the CA bundle config map (and that's strategy A).

But the cost of this new strategy is that this api is more technically demanding, because (I believe) it means refactoring the debug server to function as a proper Kubernetes API server, not just an HTTP handler behind a Service. So as Kevin mentioned, this would be a long-term approach. 

Today the debug server is a raw net/http handler registered on a PathRecorderMux. To support API aggregation, I believe it would need to:

  1. Define a new API type (PlacementDebugReview)(?) with the full Kubernetes resource shape — TypeMeta, ObjectMeta, Spec, Status
  2. Implement the k8s.io/apiserver interface — content negotiation, API discovery responses, proper request/response marshaling
  3. Register and manage an APIService resource — the operator must create/update/delete it in sync with the feature gate and keep its caBundle current
  4. Commit to an API surface — a registered API group carries versioning and deprecation obligations (alpha → beta → GA) that a plain Service endpoint does not

The current handler just parses JSON, runs the scheduling logic, and returns a result. Strategy B wraps that same logic in significantly more Kubernetes API machinery. 

@rokej I've also shared a WIP document with you that has some more details about the implications.

mikeshng (approved):
/lgtm

Your changes makes sense to me.
Let's fix the PR title and maybe your PR description if it's needed. 
CI failure(s) needs to be look at as well.

PR #1500, opened 2026-04-27
✨ Apply resource in manifestwork ordered by kind

### LINKED ISSUE #374: Addon deployments fail due to missing ServiceAccount race condition
## Summary

When OCM addons are enabled on a managed cluster, the ManifestWork
generated by the addon-framework contains resources in an order that
does not account for dependencies. The Deployment is applied before the
ServiceAccount it references via `serviceAccountName`. The ReplicaSet
controller then fails to create pods with:

```
Error creating: pods "governance-policy-framework-7cd74848fb-" is forbidden:
  error looking up service account open-cluster-management-agent-addon/governance-policy-framework-sa:
  serviceaccount "governance-policy-framework-sa" not found
```

This causes unnecessary retries with exponential backoff. In most cases
the ServiceAccount appears within 1 second and the impact is minimal.
In the worst observed case, the ServiceAccount took **165 seconds** to
appear, with 15 failed attempts and exponential backoff adding
significant delay to the overall startup.

## Addon deployment timing

The `ocm-cluster/start` step joins a managed cluster to the hub,
enables addons (application-manager, governance-policy-framework,
config-policy-controller), and waits for their deployments to roll
out. It is the slowest and most variable step in the environment
setup.

Image

The `ocm-cluster/start` step (top) has a median of ~130 seconds and
a range of 90-170 seconds. The wide spread is caused by the
ServiceAccount race condition — when the ServiceAccount takes longer
to appear, the exponential backoff in the ReplicaSet controller
delays the entire addon deployment.

## Impact

- Adds 0-165 seconds of unnecessary latency to addon deployment
- Generates FailedCreate warning events on every managed cluster join
- In a 100-run stress test, **100% of runs** were affected on
  **all 4 addons** on **both clusters**

## Environment

- Kubernetes: minikube v1.38 (containerd)
- OCM: clusteradm v0.11.0
- Host: Intel Xeon Gold 6230 (10 cores), 47 GiB RAM, Fedora 43

## Affected addons

All addons enabled via `clusteradm addon enable` are affected:

- application-manager
- governance-policy-framework
- config-policy-controller
- klusterlet-addon-workmgr

## Statistics from 100 stress test runs

Each run creates 2 fresh minikube clusters (dr1, dr2), joins them to
a hub, and enables addons. Data collected from Kubernetes events in
the `open-cluster-management-agent-addon` namespace.

- **800/800** addon deployments (4 addons × 2 clusters × 100 runs)
  experienced FailedCreate errors
- Error count distribution per deployment:

  | Retries | Occurrences |
  |--------:|------------:|
  | 2       | 14          |
  | 3       | 28          |
  | 4       | 59          |
  | 5       | 207         |
  | 6       | 390         |
  | 7       | 94          |
  | 8       | 7           |
  | 15      | 1           |

- **76%** of cases resolved within 1 second (same-second timestamps)
- **24%** took longer, with the worst case taking 165 seconds

## Root cause

The addon-framework generates a ManifestWork containing all addon
resources (ServiceAccount, ClusterRole, ClusterRoleBinding, Deployment,
etc.). The work agent on the managed cluster applies these resources
sequentially in the order they appear in `spec.workload.manifests`.

When the Deployment appears before the ServiceAccount in this list,
the Deployment is created first and its ReplicaSet immediately tries
to create pods. Since the ServiceAccount doesn't exist yet, pod
creation fails. The work agent continues applying the remaining
manifests (including the ServiceAccount), but by then the ReplicaSet
controller has already entered exponential backoff.

This is similar to how Helm sorts resources by kind to ensure
dependency order (ServiceAccount before Deployment). The addon-framework
should apply the same ordering to the manifests it generates.

## Timeline of worst case (run 099, dr2)

Times in UTC. Logs are in local time (UTC+2).

### FailedCreate errors

| Addon                        | Errors | First          | Last           |
|------------------------------|-------:|----------------|----------------|
| application-manager          | 7      | 05:01:01       | 05:01:01       |
| config-policy-controller     | 4      | 05:01:01       | 05:01:01       |
| governance-policy-framework  | 15     | 05:01:01       | 05:02:24       |
| klusterlet-addon-workmgr     | 7      | 05:01:01       | 05:01:01       |

### Pod creation (SuccessfulCreate)

| Addon                        | Created at     | Delay  |
|------------------------------|----------------|--------|
| application-manager          | 05:01:02       | ~1s    |
| config-policy-controller     | 05:01:01       | <1s    |
| governance-policy-framework  | 05:03:46       | ~165s  |
| klusterlet-addon-workmgr     | 05:01:02       | ~1s    |

### governance-policy-framework detailed timeline

```
05:01:01  ReplicaSet created, first FailedCreate error (SA not found)
05:01:01  ... retries with exponential backoff ...
05:02:24  15th FailedCreate error (SA still missing)
05:03:46  ServiceAccount created by addon controller
05:03:46  Pod created successfully
05:03:46  Image pull started
05:03:52  Image pulled in 5.52s (67MB from local registry cache)
05:03:52  Container created and started
```

The image pull itself is fast (5.52s from local cache). All 165
seconds of delay were spent waiting for the ServiceAccount to appear.

## Comparison with fast run (run 084, dr2)

| Addon                        | Errors | Delay  |
|------------------------------|-------:|--------|
| application-manager          | 3      | ~1s    |
| config-policy-controller     | 2      | <1s    |
| governance-policy-framework  | 7      | ~1s    |
| klusterlet-addon-workmgr     | 7      | ~1s    |

Same race condition, but all ServiceAccounts appeared within 1 second.
Total addon deployment time: 80s vs 222s in the slow case.

## Suggested fix

Sort resources in the ManifestWork `spec.workload.manifests` array
by kind, ensuring dependency order:

1. Namespace
2. ServiceAccount
3. ClusterRole / Role
4. ClusterRoleBinding / RoleBinding
5. ConfigMap / Secret
6. Deployment / DaemonSet

This matches the ordering used by Helm's
[`InstallOrder`](https://github.com/helm/helm/blob/main/pkg/releaseutil/kind_sorter.go)
and would eliminate the race condition for all addons.

## Attached evidence

Attached a `099-gather.tar.gz` file (6.7MB) containing the complete
`drenv gather` output from the slowest run (222s), including pod specs,
container logs, events, deployments, and cluster state for all 3
clusters (hub, dr1, dr2).

Key paths inside the archive:

- `099.gather/dr2/namespaces/open-cluster-management-agent-addon/events.k8s.io/events/` -
  FailedCreate and SuccessfulCreate events for all addons
- `099.gather/dr2/namespaces/open-cluster-management-agent-addon/pods/` -
  Pod specs and container logs
- `099.gather/dr2/namespaces/open-cluster-management-agent-addon/apps/` -
  Deployment and ReplicaSet specs
- `099.gather/dr2/cluster/` - Cluster-level info (version, nodes)

[099-gather.tar.gz](https://github.com/user-attachments/files/25964872/099-gather.tar.gz)

qiujian16:
## Related issue(s)

Fixes https://github.com/open-cluster-management-io/addon-framework/issues/374

## Summary by CodeRabbit

* **Bug Fixes / Improvements**
  * Manifests are now pre-parsed and applied in a deterministic, kind-prioritized order so resource dependencies are handled consistently, while per-manifest status entries remain aligned to their original spec positions.

* **Tests**
  * Added unit and integration tests validating ordering semantics, stable tie-breaking, handling of unknown/invalid kinds, and preservation of per-manifest status indexing.

### DIFF
--- a/pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
+++ b/pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/util/retry"
@@ -31,6 +33,67 @@ type applyResult struct {
 	Error  error
 
 	resourceMeta workapiv1.ManifestResourceMeta
+}
+
+// resourceApplyOrder defines the priority rank for applying resources by kind.
+// Lower rank means applied first. Resources of the same rank preserve their spec order
+// (stable sort). Kinds not listed here are applied after all listed kinds (rank 1000),
+// preserving their relative spec order; nil/unparseable manifests sort last (rank 1001).
+//
+// Rank 0: Cluster-scoped foundations — no dependencies on other resources
+// Rank 1: RBAC — needed before workloads that reference service accounts
+// Rank 2: Namespace-scoped policies and identity — must exist before workloads
+// Rank 3: Config and storage — referenced by workloads as volumes or env
+// Rank 4: Networking — services, ingress, etc.
+// Rank 5: Workloads — pods, deployments, jobs, etc.
+// Rank 6: Post-workload — webhooks and API aggregation that may depend on running workloads
+var resourceApplyOrder = map[string]int{
+	// Rank 0: cluster-scoped, no deps
+	"PriorityClass":            0,
+	"Namespace":                0,
+	"CustomResourceDefinition": 0,
+	"StorageClass":             0,
+	"PersistentVolume":         0,
+	"PodSecurityPolicy":        0,
+
+	// Rank 1: RBAC
+	"ClusterRole":        1,
+	"ClusterRoleBinding": 1,
+	"Role":               1,
+	"RoleBinding":        1,
+
+	// Rank 2: namespace-scoped policies and identity
+	"NetworkPolicy":       2,
+	"ResourceQuota":       2,
+	"LimitRange":          2,
+	"PodDisruptionBudget": 2,
+	"ServiceAccount":      2,
+
+	// Rank 3: config and storage
+	"Secret":                3,
+	"ConfigMap":             3,
+	"PersistentVolumeClaim": 3,
+
+	// Rank 4: networking
+	"Service": 4,
+
+	// Rank 5: workloads
+	"DaemonSet":               5,
+	"Pod":                     5,
+	"ReplicationController":   5,
+	"ReplicaSet":              5,
+	"Deployment":              5,
+	"HorizontalPodAutoscaler": 5,
+	"StatefulSet":             5,
+	"Job":                     5,
+	"CronJob":                 5,
+
+	// Rank 6: post-workload
+	"APIService":                       6,
+	"MutatingWebhookConfiguration":     6,
+	"ValidatingWebhookConfiguration":   6,
+	"ValidatingAdmissionPolicy":        6,
+	"ValidatingAdmissionPolicyBinding": 6,
 }
 
 type manifestworkReconciler struct {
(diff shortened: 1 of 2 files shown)


pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
```diff
@@ -140,14 +216,17 @@ func (m *manifestworkReconciler) applyManifests(
...
+		needsApply := existingResults[om.specIndex].Result == nil || apierrors.IsConflict(existingResults[om.specIndex].Error)
+		if !needsApply {
+			continue
+		}
+		if om.err != nil {
+			existingResults[om.specIndex] = applyResult{Error: om.err}
+		} else {
+			existingResults[om.specIndex] = m.applyOneManifest(ctx, om.specIndex, om.obj, workSpec, workStatus, recorder, owner)
 		}
 	}
```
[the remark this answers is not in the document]
qiujian16: I think current implementation in general won't affect customresources, the order that user defined still mainained, only certain well known resources are prioritized, which should be fine

pkg/work/spoke/controllers/manifestcontroller/manifestwork_reconciler.go
```diff
@@ -131,6 +170,43 @@ func (m *manifestworkReconciler) reconcile(
...
+	sort.SliceStable(ordered, func(i, j int) bool {
+		return kindOrder(ordered[i].obj) < kindOrder(ordered[j].obj)
+	})
+
+	return ordered
+}
+
+func kindOrder(obj *unstructured.Unstructured) int {
+	if obj == nil {
+		return 1000
```
zhujian7: Should we set the nil obj larger than the obj that are not in the `resourceApplyOrder` list?

qiujian16: updated

mikeshng (commented):
This looks good to me. Just a question: is the ranking from somewhere like Hem? 

Hoping another Work maintainer can take a look as well.

mikeshng:
CC @dhaiducek

bhperry:
Order looks reasonable to me, although it is different from [helm's ordering](https://helm.sh/docs/intro/using_helm/#helm-install-installing-a-package).

I wonder if there should be an annotation or other config that allows ranks to be explicitly assigned manifests? Or perhaps a setting that disables the ranked ordering altogether and just follows manifest order instead.

dhaiducek:
> Just a question: is the ranking from somewhere like Helm?

In addition to the docs, here's the Helm ordering logic in their code for reference: 
- https://github.com/helm/helm/blob/fbc279188687f5278eef6dcfa741874ad61dc566/internal/release/v2/util/kind_sorter.go#L28

But this LGTM as it is! Out of curiosity, is there a reason to sort by Kind rather than API Group? Maybe a sort like:

```
<no-api-group>
apps
batch
autoscaling
policy
rbac.authorization.k8s.io
*.k8s.io
<everything-else>
admissionregistration.k8s.io
```

Either way there's a degree of maintainability, but I wonder if the API group might pick up new kinds more readily (assuming there isn't a risk or some other factor there)?

qiujian16:
> I wonder if there should be an annotation or other config that allows ranks to be explicitly assigned manifests? Or perhaps a setting that disables the ranked ordering altogether and just follows manifest order instead

Is there any potential risk of this change? I feel that since we only change order of known types, adding an additional flag might not be necessary.

qiujian16:
> Out of curiosity, is there a reason to sort by Kind rather than API Group

There is possible that two kinds in one group has dependencies. So kind seems like a more straightforward way.

qiujian16:
> In addition to the docs, here's the Helm ordering logic in their code for reference:

I think we might want to control the order by ourselves..

bhperry:
> > I wonder if there should be an annotation or other config that allows ranks to be explicitly assigned manifests? Or perhaps a setting that disables the ranked ordering altogether and just follows manifest order instead
> 
> Is there any potential risk of this change? I feel that since we only change order of known types, adding an additional flag might not be necessary.

The question is if there would ever be a case where a user wants to create a custom resource before a known kind (e.g. a deployment that expects a particular CR to exist before it starts), or a new/missed known kind that isn't handled here which is a dependency of another kind.

Fine in most cases, but just wanted to point out that this is taking control away from the user for niche cases.

dhaiducek (approved):
LGTM! Considering Kubernetes operates on a principle of eventual consistency and this addresses the particular issue at hand, I think enabling customization of the behavior here could be addressed later if deemed necessary.

qiujian16:
> Fine in most cases, but just wanted to point out that this is taking control away from the user for niche cases.

makes sense, I think we could have an annotation to control whether order is ignored.

PR #1506, opened 2026-04-30
:sparkles: feat: add "random" placement prioritizer

dhaiducek:
## Summary

feat: add "random" placement prioritizer

## Related issue(s)

- https://github.com/open-cluster-management-io/ocm/issues/1507

## Summary by CodeRabbit

* **New Features**
  * Added a Random prioritizer plugin to produce per-cluster placement scores.

* **Bug Fixes**
  * Corrected grammar and clarity in multiple placement plugin descriptions.
  * Updated misconfiguration error wording for clearer guidance.

* **Improvements**
  * Consistent implementation updates across placement plugins for reliability.
  * Added tests covering the new Random prioritizer's scoring and distribution.

### DIFF
--- a/pkg/placement/controllers/scheduling/schedule.go
+++ b/pkg/placement/controllers/scheduling/schedule.go
@@ -25,6 +25,7 @@ import (
 	"open-cluster-management.io/ocm/pkg/placement/plugins/addon"
 	"open-cluster-management.io/ocm/pkg/placement/plugins/balance"
 	"open-cluster-management.io/ocm/pkg/placement/plugins/predicate"
+	"open-cluster-management.io/ocm/pkg/placement/plugins/random"
 	"open-cluster-management.io/ocm/pkg/placement/plugins/resource"
 	"open-cluster-management.io/ocm/pkg/placement/plugins/steady"
 	"open-cluster-management.io/ocm/pkg/placement/plugins/tainttoleration"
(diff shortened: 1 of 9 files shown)


augustrh:
cc @mikeshng can you also review this one ...

mikeshng:
@augustrh I've reviewed the API change (which is the pre-req to this implementation, API PR must be merged first) https://github.com/open-cluster-management-io/api/pull/437

It's not clear to me and [Qiu Jian](https://github.com/open-cluster-management-io/ocm/issues/1507#issuecomment-4385675238) how this API will behave yet. We can discuss more on the API PR if there is more work on it.

qiujian16:
I've appened a comment on the related issue.

dhaiducek:
Thanks for the feedback, all! Sorry for the delay. I agree I need to solidify the use cases and logic here. I'll hopefully be able to spend more time on it next week.

augustrh:
@mikeshng and @qiujian16 yeah fair point. I wonder if we could instead propose a way to scope the balance priortiser in some way? the reason for random is, i think, to avoid things like balance considering unrelated placements - instead maybe we could make balance more flexible, or introduce something like ScopedBalance to let users say something like  "balance across clusters considering only placements in my namespace" or "only placements with this label." But, this may just be making things too hard (harder than need be :) ). True randomness could attempt to land workloads in places they will fail so ... wdyt about a scoping option?

PR #1527, opened 2026-05-15
:bug: OCM v1beta1 mca conversion fix issue 1526

### LINKED ISSUE #1526: OCM 1.3 v1alpha1 to v1beta1 can cause errors with addons using older addon-framework
**Describe the bug**

 Problem

Encountered when running code developed with older versions of the addon-framework, but deployed into OCM 1.3 environment - the issue appears to be around creating ClusterManagementAddon with `v1alpha1`, with supportedconfigs, but no default config specified.                                                                                                      
                                                                                                                           
- When a v1alpha1 ClusterManagementAddOn with supportedConfigs but no defaultConfig is converted to v1beta1, the sentinel  
  value "__reserved_no_default__" leaks into the CMA status fields (DefaultConfigReferences and InstallProgressions),
  creating entries with empty SpecHash values. This causes addons using older versions of addon-framework (v0.11.0 and     
  earlier) to fail with:                                                                           
         
  addon <name> deployment config desired spec hash is empty
                                                                                                                           
  Root Cause
                                                                                                                           
  The v1alpha1 → v1beta1 conversion webhook correctly uses the sentinel value ReservedNoDefaultConfigName                  
  ("__reserved_no_default__") to preserve the semantic "this config type is supported but has no default" for round-trip
  conversion.                                                                                                              
                                                                                                   
  However, two controller functions in pkg/addon/controllers/cmainstallprogression/controller.go fail to filter out this   
  sentinel value:
                                                                                                                           
  1. setDefaultConfigReference() (line 93): Only checks for empty string, not the sentinel                                 
  2. setInstallProgression() (line 135): Incorrectly checks Namespace field instead of Name field
                                                                                                                           
  This creates status entries like:                                                                                        
  status:                                                                                                                  
    defaultConfigReferences:                                                                                               
    - group: addon.open-cluster-management.io                                                      
      resource: addondeploymentconfigs       
      desiredConfig:                  
        name: "__reserved_no_default__"                                                                                    
        specHash: ""  # Empty!         
                                                                                                                           
  Impact                                                                                                                   
         
  - Backward compatibility broken: Old addons (using addon-framework < v1.3.0) fail when deployed to OCM v1.3.0+ clusters  
  - Affects: Any v1alpha1 CMA that declares a config type in supportedConfigs without providing defaultConfig
  - Example: cluster-proxy-addon's CMA declares addondeploymentconfigs as supported but optional.


This happens when an addon uses the helper function in addon-framework:
https://github.com/open-cluster-management-io/addon-framework/blob/929daf05d87b325d61bc1f291c429144f1e2e202/pkg/utils/addon_config.go#L77-L81

And will get a desiredConfig for the "__reserved_no_default__" that will have a SpecHash that is empty.

**To Reproduce**

**Expected behavior**

**Additional context**

tesshuflower:
## Summary

For CMA created with v1alpha1 that has supportedConfigs, but no default config selected, these will end up like:

```
  apiVersion: addon.open-cluster-management.io/v1beta1                                                                     
  kind: ClusterManagementAddOn                                                                                             
  spec:                                                                                                                    
    defaultConfigs:                                                                                                        
    - group: proxy.open-cluster-management.io                                                                              
      resource: managedproxyconfigurations                                                                                 
      name: cluster-proxy                                                                          
    - group: addon.open-cluster-management.io  # Entry for the supported config, even though there is no default
      resource: addondeploymentconfigs                                                                                     
      name: "__reserved_no_default__" 
```

And this can get added to the configReferences unless specifically ignored by the cmainstallprogression controller.  It seems these should be ignored as this special case "__reserved_no_default__" means there is no actual config that should be used.

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/1526

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration filtering in addon install progression now correctly excludes reserved "no default" configuration names from default selections and install progressions.

* **Tests**
  * Added test cases validating that reserved "no default" configuration names are ignored during addon status reconciliation and progression creation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-cluster-management-io/ocm/pull/1527)

### DIFF
--- a/pkg/addon/controllers/cmainstallprogression/controller.go
+++ b/pkg/addon/controllers/cmainstallprogression/controller.go
@@ -90,7 +90,7 @@ func setDefaultConfigReference(supportedConfigs []addonv1beta1.AddOnConfig,
 	existDefaultConfigReferences []addonv1beta1.DefaultConfigReference) []addonv1beta1.DefaultConfigReference {
 	newDefaultConfigReferences := []addonv1beta1.DefaultConfigReference{}
 	for _, config := range supportedConfigs {
-		if config.ConfigReferent.Name == "" {
+		if config.ConfigReferent.Name == "" || config.ConfigReferent.Name == addonv1beta1.ReservedNoDefaultConfigName {
 			continue
 		}
 		configRef := addonv1beta1.DefaultConfigReference{
@@ -132,7 +132,7 @@ func setInstallProgression(supportedConfigs []addonv1beta1.AddOnConfig, placemen
 		// set config references as default configuration
 		installConfigReferencesMap := map[addonv1beta1.ConfigGroupResource]sets.Set[addonv1beta1.ConfigReferent]{}
 		for _, config := range supportedConfigs {
-			if config.ConfigReferent.Name == "" || config.ConfigReferent.Namespace == addonv1beta1.ReservedNoDefaultConfigName {
+			if config.ConfigReferent.Name == "" || config.ConfigReferent.Name == addonv1beta1.ReservedNoDefaultConfigName {
 				continue
 			}
 			refs, ok := installConfigReferencesMap[config.ConfigGroupResource]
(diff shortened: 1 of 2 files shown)


mikeshng (commented):
Fixing the bad namespace check seems good to me.

Hopefully, the other maintainers can chime in regarding the default.

PR #1529, opened 2026-05-18
:seedling: chore: pin `addon-examples` image for tests

dhaiducek:
## Summary

Pin the `addon-examples` image for tests to the version specified in the `go.mod` file. This way, the `addon-framework` library can be updated independently from the `OCM` repo and changes here addressed when the dependency is updated.

## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **Tests**
  * E2E tests accept a configurable addon example image tag via a new CLI flag.
  * Manifests and test flows use a templated image placeholder so the selected tag is pulled and loaded into the local test cluster.
  * Failed-test output filenames now include timestamp and attempt index for clearer artifact tracing.

* **Chores**
  * Image tag defaults are derived from the repository/module version instead of always using "latest".

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-cluster-management-io/ocm/pull/1529?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/hack/e2e_flaky_error_test.sh
+++ b/hack/e2e_flaky_error_test.sh
@@ -34,8 +34,9 @@ do
     kind load docker-image --name=e2e quay.io/open-cluster-management/addon-manager:$IMAGE_TAG
 
     # This is for addon-manager test: /workspaces/OCM/test/e2e/manifests/addon/addon_template.yaml
-    docker pull quay.io/open-cluster-management/addon-examples:latest
-    kind load docker-image --name=e2e quay.io/open-cluster-management/addon-examples:latest
+    ADDON_EXAMPLE_IMAGE_TAG=$(go list -m -mod=readonly -f '{{ .Version }}' open-cluster-management.io/addon-framework | cut -d'-' -f1)
+    docker pull "quay.io/open-cluster-management/addon-examples:${ADDON_EXAMPLE_IMAGE_TAG}"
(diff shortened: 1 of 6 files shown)


hack/e2e_flaky_error_test.sh
```diff
@@ -34,8 +34,9 @@ do
     kind load docker-image --name=e2e quay.io/open-cluster-management/addon-manager:$IMAGE_TAG
 
     # This is for addon-manager test: /workspaces/OCM/test/e2e/manifests/addon/addon_template.yaml
-    docker pull quay.io/open-cluster-management/addon-examples:latest
-    kind load docker-image --name=e2e quay.io/open-cluster-management/addon-examples:latest
+    ADDON_EXAMPLE_IMAGE_TAG=$(go list -m -mod=readonly -f '{{ .Version }}' open-cluster-management.io/addon-framework | cut -d'-' -f1)
```
qiujian16, line 37: this might be problematic, because we could pin addon-framework to the latest commit. In that case, addon-example image should pinned to latest I think

dhaiducek: Hmmm, I see. In that case, is the `addon-framework` library versioning strictly aligned with OCM?

qiujian16: you might update addon-framework lib and upgrade the deps in ocm. It might need a latest addon-example image, It is not always the case though.

Issue #1538, opened 2026-05-21
Add imagePullSecrets for the OCM cluster-manager helm chart.


blueprismo:
**Describe the enhancement**
Right now, the way to use the helm chart with docker images in a private repo is with the IMHO "clumsy" username/password/dockerconfigjson hardcoded in the secret template:

[secret source](https://github.com/open-cluster-management-io/ocm/blob/main/deploy/cluster-manager/chart/cluster-manager/templates/image_pull_secret.yaml#L1-L12)
[secret template source](https://github.com/open-cluster-management-io/ocm/blob/main/deploy/cluster-manager/chart/cluster-manager/templates/_helpers.tpl#L2)

My proposal/suggestion is to add [ImagePullSecrets](https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod) to also allow users to provide an already existing secret to allow images to be pulled directly. Only a secret Key reference is needed (and an existing secret). No hardcoding credentials in files, no startup kirks.

Also it's the recommended k8s way :)

sudeshmu:
Hi @blueprismo and maintainers,
I'm exploring this issue and looked at the current implementation. I completely agree with @blueprismo -
hardcoding credentials in values.yaml is problematic, especially for GitOps workflows where the values file is
tracked in git.

I see the current approach creates a secret from template with username/password:
- [image_pull_secret.yaml](https://github.com/open-cluster-management-io/ocm/blob/main/deploy/cluster-manager/chart/cluster-manager/templates/image_pull_secret.yaml)
- [_helpers.tpl](https://github.com/open-cluster-management-io/ocm/blob/main/deploy/cluster-manager/chart/cluster-manager/templates/_helpers.tpl#L2)

A few questions before proposing a solution:
1. **Scope:** Just `cluster-manager` chart for now, or should `klusterlet` chart also get this? (I noticed
#1510 is improving klusterlet - should they have consistent behavior?)
2. **Backward compatibility:** Should we:
   - Keep the existing template for users who need it, AND support imagePullSecrets?
   - Or deprecate the old method entirely?
   - If both are supported, which takes precedence?
3. **Approach:** I'm thinking:
   ```yaml
   # values.yaml
   imagePullSecrets:
     - name: my-existing-secret
     - name: another-secret  # support multiple
   ```
   Then add to deployment specs:
   ```yaml
   spec:
[...]

PR #1539, opened 2026-05-21
✨ Add imagePullSecrets support to cluster-manager helm chart

blueprismo:
Allow referencing existing pull secrets via images.imagePullSecrets instead of embedding credentials in values.
Document that the operator syncs the default secret name to hub components.

## Summary

This PR adds the imagePullSecrets pod attribute to the operator pod. I've seen that then the operator uses that secret, with that specific name, to sync other deployments and items.

The goal is that people can also provide their imagePullSecret and then refer it like:

```yaml
images:
  imagePullSecrets:
    - name: open-cluster-management-image-pull-credentials  # MUST be this exact name!!!!!!
  imageCredentials:
    createImageCredentials: false
```

## Related issue(s)

https://github.com/open-cluster-management-io/ocm/issues/1538

Fixes #

## Release Notes

* **New Features**
  * Added `--image-pull-secret-name` CLI flag to configure which Kubernetes image pull secret is used for hub/registration syncing.
  * Added Helm/config support for `imagePullSecrets` to reference existing pull secrets.
* **Documentation**
  * Updated chart values guidance for using existing image pull credentials (including default secret name behavior).
* **Bug Fixes**
  * Ensured the configured image pull secret name is consistently applied across deployments and RBAC.
* **Tests**
  * Expanded coverage for custom and existing image pull secret name scenarios.

### DIFF
--- a/deploy/cluster-manager/chart/cluster-manager/templates/_helpers.tpl
+++ b/deploy/cluster-manager/chart/cluster-manager/templates/_helpers.tpl
@@ -3,7 +3,8 @@
      the full list is still applied to operator pods via clusterManager.imagePullSecrets. */}}
 {{- define "clusterManager.imagePullSecretName" -}}
 {{- if .Values.images.imagePullSecrets -}}
-{{- (index .Values.images.imagePullSecrets 0).name -}}
+{{- $name := (index .Values.images.imagePullSecrets 0).name | default "" -}}
+{{- required "images.imagePullSecrets[0].name must be set when imagePullSecrets is provided" $name -}}
 {{- else -}}
 open-cluster-management-image-pull-credentials
 {{- end -}}
--- a/deploy/cluster-manager/chart/cluster-manager/templates/_helpers.tpl
+++ b/deploy/cluster-manager/chart/cluster-manager/templates/_helpers.tpl
@@ -1,3 +1,23 @@
+{{/* Name of the image pull secret synced to hub namespaces and passed to the operator.
+     When imagePullSecrets lists multiple secrets, only the first name is used here;
+     the full list is still applied to operator pods via clusterManager.imagePullSecrets. */}}
+{{- define "clusterManager.imagePullSecretName" -}}
+{{- if .Values.images.imagePullSecrets -}}
+{{- (index .Values.images.imagePullSecrets 0).name -}}
+{{- else -}}
+open-cluster-management-image-pull-credentials
+{{- end -}}
+{{- end }}
+
+{{/* Image pull secrets for operator pods and service account. */}}
+{{- define "clusterManager.imagePullSecrets" -}}
+{{- if .Values.images.imagePullSecrets -}}
+{{ .Values.images.imagePullSecrets | toYaml }}
+{{- else -}}
+- name: {{ include "clusterManager.imagePullSecretName" . }}
+{{- end -}}
+{{- end }}
+
 {{/* Create secret to access docker registry */}}
 {{- define "imagePullSecret" }}
 {{- with .Values.images }}
--- a/deploy/cluster-manager/chart/cluster-manager/templates/cluster_role.yaml
+++ b/deploy/cluster-manager/chart/cluster-manager/templates/cluster_role.yaml
@@ -34,7 +34,7 @@ rules:
     - "addon-webhook-sa-kubeconfig"
     - "external-hub-kubeconfig"
     - "work-driver-config"
-    - "open-cluster-management-image-pull-credentials"
+    - {{ include "clusterManager.imagePullSecretName" . | quote }}
     - "grpc-server-serving-cert"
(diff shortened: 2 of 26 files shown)


qiujian16 (commented):
I think you will need to at least set a flag on operator to specify the secret name. Then the operator can sync the secret to open-cluster-management-hub or open-cluster-management-agent namespaces. 

I think currently the PR only change the pull secret for operator image itself, but not other controller or agent.

blueprismo:
> I think you will need to at least set a flag on operator to specify the secret name. Then the operator can sync the secret to open-cluster-management-hub or open-cluster-management-agent namespaces.
> 
> I think currently the PR only change the pull secret for operator image itself, but not other controller or agent.

Thanks for the review, you're right.

I can extend this with a flag like `--image-pull-secret-name`, similar to [--enable-sync-labels](https://github.com/open-cluster-management-io/ocm/blob/v1.3.1/deploy/cluster-manager/chart/cluster-manager/templates/operator.yaml#L41)), defaulting to the current name for backward compatibility. That would also require updating the RBAC templates in both the [cluster-manager](https://github.com/open-cluster-management-io/ocm/blob/v1.3.1/deploy/cluster-manager/chart/cluster-manager/templates/cluster_role.yaml#L37) and klusterlet charts, and wiring it through the cluster-manager and klusterlet secret sync paths so the secret propagates to open-cluster-management-hub and open-cluster-management-agent.

Before I proceed: should this be cluster-manager-only for now, or should the klusterlet chart get the same treatment in the same PR? Also, I think it's worth aligning with the people on https://github.com/open-cluster-management-io/ocm/issues/1538.

qiujian16:
we do not need container both of them in 1 PR, but I think we finally will need klusterlet to support this also.

blueprismo:
> we do not need container both of them in 1 PR, but I think we finally will need klusterlet to support this also.

Alrighty, added those changes in an additional commit. Is this enough for the PR of this scope? (and then extend the same functionality to clusterlet in a follow-up PR?)

blueprismo:
Sorry for the noise, friendly ping @qiujian16

mikeshng (commented):
This PR makes sense to me. It's clean and the behavior is optional. It looks good to me.

/assign @rokej 

@rokej could you PTAL at this one? It falls under the operator maintainer's domain. Thanks.

@blueprismo thank you for your PR! Could you please make sure the DCO check passes?

rokej:
https://github.com/open-cluster-management-io/ocm/blob/main/pkg/registration/hub/importer/renderers.go#L23 still hardcodes the pull secret name. I think you need to pass the configured name into RenderImagePullSecret (same pattern as the operator flag), or read it from HubConfig/registration config.

deploy/cluster-manager/chart/cluster-manager/templates/_helpers.tpl
```diff
@@ -1,3 +1,27 @@
+{{/* The name of the image pull secret created by the chart or referenced by default. */}}
+{{- define "clusterManager.imagePullSecretName" -}}
+{{- define "clusterManager.imagePullSecretName" -}}
+{{- if .Values.images.imagePullSecrets -}}
+{{- $name := (index .Values.images.imagePullSecrets 0).name | default "" -}}
```
rokej, line 6: What happens when there are multiple image pull secrets?

blueprismo: Thanks for the review!

When there are multiple secrets the full secrets array is applied to the operator SA and deployment (k8s behaviour). For hub sync and RBAC (ocm behaviour) we use the first entry name (or the default `open-cluster-management-image-pull-credentials` if `imagePullSecrets` isn't set). The operator syncs that one secret by name into hub namespaces and sets the pod `imagePullSecrets` accordingly, matching the existing sync path. 

Added some comments in both the values.yaml and _helpers.tpl to document this.

Klusterlet/agent namespace support is planned as a follow-up.

If users need multi-registry access, they can merge all credentials in one single secret. Otherwise we would need to allow cluster-manager to handle multiple secrets (sync, rbac, etc.) which adds more complexity, happy to revisit if that's a hard requirement

Let me know if this approach doesn't work for you

blueprismo:
@mikeshng DCO signed! (accepting the code review in the UI broke that check 😅 )
@rokej thanks for the `RenderImagePullSecret` catch and review, addressed

mikeshng:
@rokej @qiujian16 all the CIs are now passed, could you please take another look at this PR? Thanks!

deploy/cluster-manager/chart/cluster-manager/templates/_helpers.tpl
rokej: That sounds like a reasonable approach. Thank you.

rokej:
/lgtm
Thank you @blueprismo

PR #1367, opened 2026-02-02
🌱 Update base image to distroless for less CVE

qiujian16:
## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **Chores**
  * Optimized container base images across multiple deployment components for improved security posture and reduced resource footprint.
  * Enhanced runtime security through non-root user execution configuration for containerized environments.

PR #1556, opened 2026-06-03
:seedling: chore: update base image to distroless

dhaiducek:
## Summary

Reduces attack surface of the image and increase CVE maintainability

## Related issue(s)

Fixes # N/A

Closes #1367 

## Summary by CodeRabbit

* **Chores**
  * Switched final runtime base across container builds to a smaller UBI micro image and updated CI image-pull steps accordingly, affecting final runtime images.
* **Tests**
  * Added a configurable option to skip image builds when running end-to-end tests, making the test dependency chain deterministic and speeding CI when rebuilds aren’t required.

### DIFF
--- a/.github/workflows/post.yml
+++ b/.github/workflows/post.yml
@@ -60,7 +60,7 @@ jobs:
       - name: install imagebuilder
         run: go install github.com/openshift/imagebuilder/cmd/imagebuilder@v1.2.3
       - name: pull base image
-        run: docker pull gcr.io/distroless/base-debian13:latest --platform=linux/${{ matrix.arch }}
+        run: docker pull registry.access.redhat.com/ubi9/ubi-micro:latest --platform=linux/${{ matrix.arch }}
       - name: images
         run: |
           IMAGE_TAG=latest-${{ matrix.arch }} \
--- a/.github/workflows/post.yml
+++ b/.github/workflows/post.yml
@@ -60,7 +60,7 @@ jobs:
       - name: install imagebuilder
         run: go install github.com/openshift/imagebuilder/cmd/imagebuilder@v1.2.3
       - name: pull base image
-        run: docker pull registry.access.redhat.com/ubi9/ubi-minimal:latest --platform=linux/${{ matrix.arch }}
+        run: docker pull gcr.io/distroless/base-debian13:latest --platform=linux/${{ matrix.arch }}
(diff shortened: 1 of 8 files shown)


qiujian16:
this looks good to me. I wonder why we choose debian, do we have some deps on debian?

/approve

dhaiducek:
> this looks good to me. I wonder why we choose debian, do we have some deps on debian?

Not necessarily. I was just basing this PR off the previous one--I hadn't looked too deeply into other options. Though there is a `ubi-micro` image now, which would likely be closer to the original `ubi-minimal` image.

dhaiducek:
I pulled the images to do a trivial comparison. Since we need the `base` flavor (`static` is _really_ small at 3MB but don't have dependencies for `CGO_ENABLED`), it looks like `ubi-micro` is actually slightly smaller currently, and that `56 years ago` creation date doesn't exactly instill confidence since the image metadata is unpopulated (i.e. `"Created": "1970-01-01T00:00:00Z",`).

```
REPOSITORY                                 TAG     IMAGE ID      CREATED       SIZE
registry.access.redhat.com/ubi9/ubi-micro  latest  5f002f2f6a54  8 days ago    24.7 MB
gcr.io/distroless/base-debian13            latest  2d7208a29119  56 years ago  36.2 MB
```

So I've updated to use `ubi-micro` here.

rokej:
@dhaiducek I came across "Chainguard's entire business model revolves around maintaining Zero-CVE images."

dhaiducek:
> @dhaiducek I came across "Chainguard's entire business model revolves around maintaining Zero-CVE images."

Sure, but I'd be surprised if Chainguard is doing something particularly special and Google and Red Hat are any less CVE-aware. Moving to any distroless is beneficial and since the previous image was `ubi-minimal`, it'd be a more incremental step in the small chance any users are leveraging utilities for debugging or using the package manager.

dhaiducek:
Could someone re-run the CI here? I'm not sure the failures are related to the update to distroless.

tesshuflower:
> Could someone re-run the CI here? I'm not sure the failures are related to the update to distroless.

I like this change! Would be nice to keep the image as minimal (er, micro) as possible :)

Unfortunately I also can't re-run the CI - you may need to re-push your changes to trigger it all?

PR #1573, opened 2026-06-14
:bug: fix flaky e2e cleanup and addon conversion test

bot69dude:
## Summary
- Pre-clean ManagedClusterAddOns and ManifestWorks in `CleanKlusterletRelatedResources` before deleting the ManagedCluster (fixes AfterSuite cleanup timeout when pre-delete ManifestWorks block deletion)
- Merge status update and v1beta1 conversion verify into one `Eventually` loop in the failing CMA conversion e2e test (avoids race with `cmaInstallProgressionController`)

Fixes #1402

## Test plan
- [x] `go test -c ./test/e2e/`
- [x] `go vet ./test/e2e/... ./test/framework/...`
- [x] `go test ./pkg/addon/webhook/...`
- [ ] Full e2e validated via CI (not run locally due to resource constraints)

## Summary by CodeRabbit

* **Tests**
  * Improved conversion test reliability by validating resource specifications and status together.
  * Enhanced cleanup verification to ensure related addons and workload resources are fully removed before cluster deletion.
  * Added handling for expected resource-not-found responses during cleanup checks.
  * Preserved coverage for subsequent resource updates and deletion workflows.

Issue #1402, opened 2026-02-26
CI Checks Flaky Error


xuezhaojun:
Recently, ocm's CI checks fail on unit test, integration test and e2e test, need to invest why and how to fix

Example(e2e-grpc):

```
[AfterSuite] 
/home/runner/work/ocm/ocm/test/e2e/e2e_suite_test.go:225
  STEP: clean klusterlet e2e-universal-klusterlet resources after the test case @ 02/26/26 02:05:42.906
  [FAILED] in [AfterSuite] - /home/runner/work/ocm/ocm/vendor/github.com/onsi/gomega/internal/async_assertion.go:145 @ 02/26/26 02:08:12.914
[AfterSuite] [FAILED] [150.008 seconds]
[AfterSuite] 
/home/runner/work/ocm/ocm/test/e2e/e2e_suite_test.go:225

  [FAILED] Timed out after 150.000s.
  Expected success, but got an error:
      <*errors.errorString | 0xc001130d70>: 
      managed cluster e2e-universal-managedcluster still exists
      {
          s: "managed cluster e2e-universal-managedcluster still exists",
      }
  In [AfterSuite] at: /home/runner/work/ocm/ocm/vendor/github.com/onsi/gomega/internal/async_assertion.go:145 @ 02/26/26 02:08:12.914
```

```
[FAILED] Timed out after 150.001s.
  Unexpected error:
      <*errors.errorString | 0xc000ddebb0>: 
[...]

PR #1574, opened 2026-06-14
:sparkles: feat: add annotation to skip managed cluster clock sync check

### LINKED ISSUE #1509: Allowing ManagedClusterClockOutOfSync to be bypassed
**Describe the enhancement**
A clear and concise description of what the enhancement request is.

Hello, while testing to register a managd cluster that is out of sync with the hub cluster, I realized that the status of my managed cluster is `Unknown` due to the reason `ManagedClusterClockOutOfSync`

It seems like the status of the managed cluster is dependant with the synchronization of the hub and the managed cluster. CF :

https://github.com/open-cluster-management-io/ocm/blob/8690c37917b2e3358ad27c381d69bf7e4ec106f1/pkg/registration/hub/lease/clocksynccontroller.go#L114-L133

However, I think this synchronization should not be required all the time. For instance, if our hub cluster is in a region and one or many of our managed clusters are in different zones (eg, Europe and Asia). 

We should be able to say during managed cluster enrollment that the managed clustr is not synchronized and it is normal. I'm not exactly sure what are the implications of a managed cluster not being synchronized but if possible, it would be nice to have this option otherwise our managed clusters that are not in the same region as the hub cluster will never have a proper status

gauravshinde1729:
## Summary

  Adds an opt-out for the hub clock-sync check on a per-cluster basis.

When a `ManagedCluster` is annotated with  `registration.open-cluster-management.io/skip-clock-sync-check: "true"`,
the hub `ClockSyncController` treats the cluster's clock as synced regardless of the drift between the hub's clock and the agent's lease `RenewTime`. As a result the `ClockSynced` condition stays `True` and the agent is not blocked from updating the managed cluster status.

 This is useful for geo-distributed fleets where a hub and a spoke sit in distant regions and a real (harmless) clock offset would otherwise leave the managed cluster stuck in `Unknown`.

  **Behavior**
  - Default (no annotation): unchanged — clocks too far apart still set
    `ClockSynced=False` / reason `ManagedClusterClockOutOfSync`.
  - With the annotation set to `"true"`: `ClockSynced` is kept `True`.
 

  **Testing**
  - Added a unit test in `clocksynccontroller_test.go` covering an out-of-sync lease + annotation → `ClockSynced=True`.

  ## Related issue(s)

  Fixes #1509

## Summary by CodeRabbit

* **New Features**
  * Added a cluster-level setting to opt out of clock-synchronization validation, allowing managed clusters to bypass out-of-sync checks when required.

* **Tests**
  * Expanded clock-sync controller test coverage to verify the managed cluster “clock synced” condition is correctly set when the opt-out setting is enabled or disabled.


gauravshinde1729:
The failing e2e* checks are unrelated to this change — they fail at setup because engineerd/setup-kind@v0.6.2 is broken
(dist/main/index.js not found), which affects all PRs. This should be resolved by #1572.   All code checks (unit, integration,
build, verify, DCO) pass here. I'll rebase once #1572 merges.

cc @tesshuflower

gauravshinde1729:
@ldpliu @jaswalkiranavtar  can you please review this PR 
Thanks !

gauravshinde1729:
@qiujian16  The e2e failure is the known-flaky [addon-manager] "template type addon" spec (Eventually timeout), unrelated to this change. The other 3 e2e variants passed. Could a maintainer re-run the failed e2e job ? 

Following is the PR related to flaky test
(https://github.com/open-cluster-management-io/ocm/pull/1573)

gauravshinde1729:
Hi @qiujian16 👋
The CI failures on this PR are pre-existing flaky tests, not related to this change (it only touches the hub clock-sync controller).

I root-caused and fixed two flaky registration integration tests — ClusterProfile "should NOT create ClusterProfile when binding is not bound" and Addon Lease Resync — in a separate, test-only PR: #1589.

Plan: once #1589 merges, I'll rebase this PR and re-trigger CI so the integration job runs against the deflaked tests. (The separate e2e addon-conversion flake is tracked under #1402 / #1573.)

Can you please check this https://github.com/open-cluster-management-io/ocm/pull/1589

Thanks !

gauravshinde1729:
Hey @ldpliu @jaswalkiranavtar @qiujian16  , The E2E test seems to be completed , Can you please review this ! 
Thanks !

PR #1575, opened 2026-06-14
:sparkles: feat: add extraVolumes and extraVolumeMounts support to klusterlet helm chart

### LINKED ISSUE #1510: Improve helm chart for klusterlet
**Describe the enhancement**
A clear and concise description of what the enhancement request is.

Hello, currently the klusterlet chart does not allow us to mount custom volumes during deployment. This leads to not beind able to mount custom `ca` for instance. So if we have a hub cluster with unknown certificate authority we are blocked.

I propose that we can have the possibility to mount custom volumes that will be mounted during deployment. Example for the values yaml in the chart :

```
# extra volumes
extraVolumes: []
# extra volume mounts
extraVolumeMounts: []
```

In the templates, for the deployment, it could look similar to that

```
[...]
        volumeMounts:
        - name: tmpdir
          mountPath: /tmp
          {{- with .Values.extraVolumeMounts }}
            {{ toYaml . | nindent 10 }}
          {{- end }}
[...]
```

It would also require to  create a new template in addition to that to range on the extra volumes to create them, something like `pvc.yml`

In addition to that, maybe having the possibility to set that ca is custom and might not be trusted could be great for dev environments.

Also, this skip tls option should appear in the `hub-kubeconfig-secret` that is created once csr has been approved on hub cluster side.

digvijay-y:
## Summary
Adds `extraVolumes` and `extraVolumeMounts` support to the klusterlet Helm chart operator Deployment.

- `values.yaml`: added `extraVolumes: []` and `extraVolumeMounts: []`  
- `templates/operator.yaml`: wired both into the Deployment template

## Related issue(s)

Fixes #1510

## Summary by CodeRabbit

* **New Features**
  * Added support for custom volume and volume mount configuration in the klusterlet operator deployment.
  * Users can now provide additional volumes and mounts via configuration values for greater deployment flexibility.


deploy/klusterlet/chart/klusterlet/templates/operator.yaml
```diff
@@ -69,11 +69,17 @@ spec:
...
         emptyDir: { }
+      {{- with .Values.extraVolumes }}
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
 {{- end }}
-{{- end }}
+{{- end }}
```
qiujian16: eof

digvijay-y:
@jnpacker @qiujian16  I have updated the PR, could you review it whenever possible.

PR #1585, opened 2026-06-17
:bug: fix(cmainstallprogression): wrap PatchStatus calls in RetryOnConflict

### LINKED ISSUE #1582: `cmainstallprogression` controller PatchStatus calls missing RetryOnConflict — MCAs stuck Unknown after hub restart
## Describe the bug

After a hub cluster restart, `ManagedClusterAddOn` objects for enrolled clusters remain in `Unknown` or `ConfigurationUnsupported` state indefinitely. The only recovery is manual intervention (patch `status.supportedConfigs` by hand or restart the addon-manager).

## Root cause

`pkg/addon/controllers/cmainstallprogression/controller.go` contains three bare `PatchStatus` calls (one per code path: manual, self-managed, and placement install strategies) with no `RetryOnConflict` around them.

On hub restart, the informer cache is rebuilt and a reconcile event fires for every `ClusterManagementAddOn` simultaneously. Multiple reconcile goroutines race to patch the same CMA's `status.supportedConfigs`. Because the sdk-go patcher makes a single `client.Patch()` call with no retry, concurrent attempts fail with `409 Conflict` — the resourceVersion has already changed between the GET (lister cache) and the PATCH.

The 409 causes the controller to requeue the CMA, but by that time the other goroutines have all failed too. The field never gets written.

**Downstream effect:** `ClusterManagementAddOn.status.supportedConfigs` stays empty → any `spec.configs` patch on a `ManagedClusterAddOn` is rejected with `ConfigurationUnsupported` → the `addontemplate` controller cannot render ManifestWorks → all addons remain `Dead`/`Unknown` on every spoke.

## Affected versions

- Confirmed on **1.3.1**
- Likely also affects 1.2.x (same code path)

## Reproduction steps

1. Enroll 3+ spoke clusters to a hub (enough to generate concurrent CMA reconcile events)
2. Restart the `cluster-manager` deployment: `kubectl rollout restart deployment/cluster-manager -n open-cluster-management`
3. Observe: `kubectl get managedclusteraddon -A` shows many addons in `Unknown` / `ConfigurationUnsupported`
4. Observe: `kubectl get clustermanagementaddon -o json | jq '.[].status.supportedConfigs'` shows `null` on affected CMAs
5. Observe cluster-manager logs for lines like: `failed to patch cma status ... the object has been modified; please apply your changes to the latest version`

## Proposed fix

Wrap each of the three `PatchStatus` call sites in `retry.RetryOnConflict(retry.DefaultRetry, ...)`. Because `PatchStatus` takes both `newObject` and `oldObject` and creates a merge patch (it does not re-fetch internally), each retry closure must re-fetch a fresh object from the API server before re-applying the status mutation.

A fix with a unit test is available at: *(add PR link once opened)*

The unit test (`TestPatchStatusRetriesOn409Conflict`) injects 2 consecutive 409 Conflicts via a `PrependReactor`, runs `sync()`, and asserts the reconciler succeeds and writes `status.defaultConfigReferences`.

## Workaround

Until this is fixed upstream, operators can add a "status healer" controller that bootstraps `status.supportedConfigs` from `spec.supportedConfigs` after detecting the empty-status condition:

```bash
kubectl patch clustermanagementaddon <name> \
  --subresource=status --type=merge \
  --patch '{"status":{"supportedConfigs":[{"group":"addon.open-cluster-management.io","resource":"addondeploymentconfigs"},{"group":"addon.open-cluster-management.io","resource":"addontemplates"}]}}'
```

(Repeat for each affected CMA.)

mcphersonwhite-axon:
## What this PR does

Wraps the three bare `PatchStatus` call sites in `cmainstallprogression/controller.go` in `retry.RetryOnConflict`, fixing a thundering-herd race that leaves `ClusterManagementAddOn.status.supportedConfigs` unpopulated after hub restart.

Fixes #1582

## Root cause

On hub restart the informer cache rebuilds and fires a reconcile event for every `ClusterManagementAddOn` simultaneously. Multiple goroutines race to patch the same CMA status. The sdk-go patcher makes a single `client.Patch()` with no retry — concurrent attempts fail with `409 Conflict`. The field is never written, so every subsequent `MCA.spec.configs` patch is rejected with `ConfigurationUnsupported` and ManifestWorks are never created.

## Changes

- Extracts each of the three `sync()` code paths into dedicated private methods: `patchManualStatus`, `patchSelfManagedStatus`, `patchInstallProgressionStatus`
- Each wraps its `PatchStatus` call in `retry.RetryOnConflict(retry.DefaultRetry, ...)`
- First attempt reuses the lister-cached object (no extra API call on the happy path); any retry re-fetches a fresh object from the API server to get the current `resourceVersion`
- Adds `addonClient` field to the struct for use in retry re-fetches

## Test

`TestPatchStatusRetriesOn409Conflict` injects 2 consecutive `409 Conflict` errors via `PrependReactor`, calls `sync()`, and asserts no error is returned, exactly 2 conflicts fired, and `status.defaultConfigReferences` is populated. The test fails to build on unpatched main, proving it is coupled to the fix.

## Testing

```
go test ./pkg/addon/controllers/cmainstallprogression/... -v -count=1
make test-addon-integration
```

Both pass on this branch (27/27 integration specs pass).

## New Features
* None.

## Bug Fixes
* Improved resilience when updating cluster management add-on status by retrying on concurrent modification (409) conflicts until the update succeeds.

## Refactor
* Reworked install progression reconciliation to use separate helper flows based on installation strategy, with updated status recalculation rules per path.

## Tests
* Added a unit test that simulates repeated 409 conflicts during status patching and verifies the controller succeeds and updates default configuration references.

### DIFF
--- a/pkg/addon/controllers/cmainstallprogression/controller.go
+++ b/pkg/addon/controllers/cmainstallprogression/controller.go
@@ -5,6 +5,8 @@ import (
 	"sort"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 
 	addonv1beta1 "open-cluster-management.io/api/addon/v1beta1"
@@ -20,7 +22,8 @@ import (
 // cmaInstallProgressionController reconciles instances of clustermanagementaddon the hub
 // based to update related object and status condition.
 type cmaInstallProgressionController struct {
-	patcher patcher.Patcher[
+	addonClient addonclient.Interface
+	patcher     patcher.Patcher[
 		*addonv1beta1.ClusterManagementAddOn, addonv1beta1.ClusterManagementAddOnSpec, addonv1beta1.ClusterManagementAddOnStatus]
 	clusterManagementAddonLister addonlisterv1beta1.ClusterManagementAddOnLister
 	addonFilterFunc              factory.EventFilterFunc
@@ -33,6 +36,7 @@ func NewCMAInstallProgressionController(
 	addonFilterFunc factory.EventFilterFunc,
 ) factory.Controller {
 	c := &cmaInstallProgressionController{
+		addonClient: addonClient,
 		patcher: patcher.NewPatcher[
 			*addonv1beta1.ClusterManagementAddOn, addonv1beta1.ClusterManagementAddOnSpec, addonv1beta1.ClusterManagementAddOnStatus](
 			addonClient.AddonV1beta1().ClusterManagementAddOns()),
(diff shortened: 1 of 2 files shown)


pkg/addon/controllers/cmainstallprogression/controller.go
```diff
@@ -59,31 +63,101 @@ func (c *cmaInstallProgressionController) sync(ctx context.Context, syncCtx fact
...
+
+// patchManualStatus updates the CMA status for manual install strategy addons, retrying on conflict.
+func (c *cmaInstallProgressionController) patchManualStatus(
+	ctx context.Context,
+	seed *addonv1beta1.ClusterManagementAddOn,
+) error {
+	// needFresh tracks whether we need to re-fetch from the API server on subsequent retry iterations.
+	// The first attempt reuses the lister-cached object so no extra Get is issued on the happy path.
+	needFresh := false
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
qiujian16, line 92: why do we need to retry here? we commonly let error caught by the controller and controller will retry

mcphersonwhite-axon: Thank you for the review. The standard controller framework requeue handles transient errors, but 409 Conflicts from concurrent PatchStatus calls need specific treatment. 

When the controller returns a 409 error to the framework:

1. The item is requeued with exponential backoff (typically 5-10s minimum)
2. On retry, the controller re-reads from the lister cache — which may still have the stale resourceVersion because the informer sync lag hasn't caught up yet
3. The next PatchStatus call uses the same stale resourceVersion → another 409 → requeues again

This creates a feedback loop on hub restart when all CMAs reconcile simultaneously (the thundering herd). The framework requeue doesn't break the cycle because:

- All N reconcile events are in the queue simultaneously
- They all fail with 409, all requeue, all come back after backoff, all conflict again
- The lister cache is updated infrequently enough that the stale-version problem persists across requeues

RetryOnConflict is not for transient errors (where the framework retry is appropriate), it's specifically for optimistic concurrency conflicts where the resolution requires a fresh resource version from the API server. The controller framework doesn't re-fetch from the API on requeue — it reads from the lister cache — which is precisely what makes it insufficient here.

A reference pattern: [k8s.io/client-go/util/retry.RetryOnConflict ](k8s.io/client-go/util/retry.RetryOnConflict) is widely used in controllers where optimistic locking conflicts are expected during high-contention writes (see staging/src/k8s.io/client-go/util/retry/util.go in the Kubernetes core repo).

mikeshng (commented):
CC @tesshuflower @kahirokunn 

Tagging the msa and cp addons maintainers.

@mcphersonwhite-axon thank you for the PR, could you please sign-off on your commit?

pkg/addon/controllers/cmainstallprogression/controller.go
qiujian16: The reason I ask is it may cause the flip flop, another actor might set different values and retryOnConflict basically override the change from other actor in this field. Have we considered using server side apply which may have a better field level ownership control?

mcphersonwhite-axon:
> CC @tesshuflower @kahirokunn
> 
> Tagging the msa and cp addons maintainers.
> 
> @mcphersonwhite-axon thank you for the PR, could you please sign-off on your commit?

yes will do. Sorry forgot to add.

kahirokunn:
@mcphersonwhite-axon I'm looking forward to your sign-off 🙏

pkg/addon/controllers/cmainstallprogression/controller.go
rokej: This PR addresses the immediate bug although it seems @qiujian16's concern about concurrent controllers write to the same CMA status arrays is valid as well. I think this PR is a big improvement from the current code and the server-side apply should be dealt with a separate re-design effort.

mcphersonwhite-axon:
> @mcphersonwhite-axon I'm looking forward to your sign-off 🙏 
> 
> 

We should be all signed-off

mcphersonwhite-axon:
@kahirokunn Question, are there any other items I need to tackle. I see the blocking failed PR check, but I am not seeing how this was affected by the code change. Also I am not able to trigger another run. Is there anything I am missing?

mikeshng (commented):
Thank you for the PR @mcphersonwhite-axon . I can handle the CI retry but the main reason the PR is staling is I think @qiujian16 is waiting for your reply to his comment: https://github.com/open-cluster-management-io/ocm/pull/1585/changes#r3456501386

pkg/addon/controllers/cmainstallprogression/controller.go
mcphersonwhite-axon: @qiujian16 for this particular, PR I did not consider server side apply. The main goal here was to allow CMA install progression controller to be more resilient to changes and retry if there is a conflict due to the stale versions of the CMA objects. 

We could look into server-side applies, just was not top of mind.

PR #1589, opened 2026-06-20
:seedling: deflake registration integration tests (ClusterProfile, addon lease resync)

gauravshinde1729:
PR title

  :seedling: deflake registration integration tests (ClusterProfile and addon lease resync)

  PR description

  ## Summary

  Fixes two flaky tests in the registration integration suite (test-only changes).

  **1. ClusterProfile — "should NOT create ClusterProfile when binding is not bound"**
  The test created the `ManagedClusterSet`, which made the binding become *bound* — so it never really tested the "not bound" case and raced while the binding flipped.
 
 Fix: don't create the clusterset, so the binding stays unbound for the whole test,
  then assert no ClusterProfile is created.

  **2. Addon Lease Resync — "should update addon status to unavailable after addon lease controller resync"**
  The addon lease was created once and never renewed, so the "available" window was only ~5s and a slow runner could miss it.
  Fix: keep the lease renewed until the addon is confirmed "available", then stop renewing to test the transition to "unavailable".

  ## Related issue(s)

  Part of #1402

## Release Notes

* **Tests**
  * Added addon lease renewal test with background refresh cycles to validate availability state transitions and status changes.
  * Improved ClusterProfile binding test determinism by ensuring referenced resources are absent, preventing race conditions and reliably verifying that profiles are not created when bindings remain unbound.

### DIFF
--- a/test/integration/registration/addon_lease_test.go
+++ b/test/integration/registration/addon_lease_test.go
@@ -193,6 +193,32 @@ var _ = ginkgo.Describe("Addon Lease Resync", func() {
 	ginkgo.It("should update addon status to unavailable after addon lease controller resync", func() {
 		assertSuccessClusterBootstrap()
 		assertAddOn()
+
+		// The addon lease is created once and is not renewed by any agent in this test.
+		// With AddOnLeaseControllerLeaseDurationSeconds=1 the lease goes stale after a
+		// ~5s grace period, so the "available" window is too short to observe reliably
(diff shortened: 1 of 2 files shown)


mikeshng (commented):
I like this PR. Closing and reopening it to retrigger the CI to see how it goes.

mikeshng (approved):
/approve

/lgtm

Thank you for your contribution!

PR #1593, opened 2026-06-22
:bug: fix(placement): return correct HTTP status codes from debug endpoint

### LINKED ISSUE #1592: Bug: PlacementDebugServer returns HTTP 200 for non-existent Placement
## Description

When querying the PlacementDebugServer debug endpoint with a GET request for a non-existent Placement, the server returns HTTP **200 OK** even though the JSON response body contains an error message indicating the placement was not found. The response should use HTTP **404 Not Found** instead.

This makes it difficult for callers (scripts, operators, or monitoring) to distinguish a successful debug response from a client error without parsing the JSON body.

## Steps to reproduce

1. Install the cluster-manager with the placement debug server enabled.
2. Exec into the placement controller pod and query a placement that does not exist:

```bash
oc exec -n open-cluster-management-hub \
  deploy/cluster-manager-placement-controller \
  -- curl -k -s -w "\nHTTP_CODE: %{http_code}\n" \
  -X GET "https://cluster-manager-placement.open-cluster-management-hub.svc:9443/debug/placements/default/does-not-exist"
```

## Actual results

```
{"error":"[placement.cluster.open-cluster-management.io|http://placement.cluster.open-cluster-management.io] \"does-not-exist\" not found"}
HTTP_CODE: 200
```

## Expected results

```
{"error":"[placement.cluster.open-cluster-management.io|http://placement.cluster.open-cluster-management.io] \"does-not-exist\" not found"}
HTTP_CODE: 404
```

## Root cause

Two compounding issues in `pkg/placement/debugger/debugger.go`:

1. **`reportErr` never sets an HTTP status code** (`debugger.go:185-191`): The function writes a JSON error body via `w.Write()` without calling `w.WriteHeader()`. In Go's `net/http`, calling `Write` without a preceding `WriteHeader` implicitly sends `200 OK`. Every error path through `Handler` — including the "placement not found" branch — therefore returns HTTP 200.

2. **No error-type discrimination** (`debugger.go:98-102`): When `placementLister.Placements(namespace).Get(name)` fails because the placement does not exist, it returns a `k8s.io/apimachinery/pkg/api/errors` Not Found error. The handler passes this error directly to `reportErr` without inspecting it (e.g. with `apierrors.IsNotFound`).

## Suggested fix

1. Refactor `reportErr` to accept an HTTP status code and call `w.WriteHeader(statusCode)` before writing the body. Set `Content-Type: application/json` on error responses.

2. In the GET branch of `Handler`, classify lister errors:

```go
placement, err = d.placementLister.Placements(namespace).Get(name)
if err != nil {
    if apierrors.IsNotFound(err) {
        d.reportErr(w, http.StatusNotFound, fmt.Errorf("placement '%s' not found", name))
    } else {
        d.reportErr(w, http.StatusInternalServerError, err)
    }
    return
}
```

3. Add unit test coverage asserting the HTTP status code for the not-found case (existing tests only check the JSON error string, not the status code).

## Additional context

The same underlying `reportErr` deficiency affects other error paths in the handler (e.g. permission failures should return 403, parse/validation errors should return 400). A single refactor of `reportErr` can address all of these consistently.

rokej:
## Summary

- Refactor `reportErr` to set `Content-Type`, call `WriteHeader`, and return appropriate HTTP status codes instead of always defaulting to 200 OK
- Classify handler errors: bad request (400), unauthorized (401), forbidden (403), placement not found (404), internal server error (500)
- Add unit test coverage asserting HTTP 404 for non-existent placements and status codes for permission failures

Fixes #1592

## Test plan

- [x] `go test ./pkg/placement/debugger/... -v`
- [ ] Query a non-existent placement via the debug endpoint and confirm HTTP 404

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the placement debugger API to return accurate HTTP status codes: `400` for invalid requests, `401/403` for permission issues, `404` for missing placements, and `500` for server errors.
  * Standardized error responses to always be structured JSON with `application/json` content type.
* **Tests**
  * Enhanced automated tests to validate status codes and JSON error payloads across bad requests, missing placements, and permission scenarios.

### DIFF
--- a/pkg/placement/debugger/debugger.go
+++ b/pkg/placement/debugger/debugger.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


qiujian16 (commented):
/approve
/lgtm

PR #1601, opened 2026-06-26
:bug: Fix config references order

### LINKED ISSUE #1600: Re-ordering ManagedClusterAddOn configs has no effect
**Describe the bug**

Re-ordering entries in `ManagedClusterAddOn.spec.configs` does not update
the order in `status.configReferences`. The order is determined by the
first reconcile and never changes, making it impossible to alter config
precedence after initial creation.

Config precedence depends on index ordering — the documented contract on
`GetAddOnDeploymentConfigValues` (in addon-framework) states that the
highest-index entry takes precedence. However, changing which config
appears last in `spec.configs` has no effect because
`status.configReferences` retains its original order.

**To Reproduce**

Prerequisites: a hub cluster with OCM addon-manager running. This example
uses `cluster1` as the managed cluster name, consistent with the
[OCM contributing guidelines](https://github.com/open-cluster-management-io/ocm/blob/main/CONTRIBUTING.md)
for kind-based test environments.

1. Create an `AddOnTemplate` with a variable placeholder:

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: AddOnTemplate
metadata:
  name: ordering-demo
spec:
  addonName: ordering-demo
  agentSpec:
    workload:
      manifests:
        - apiVersion: v1
          kind: ConfigMap
          metadata:
            name: ordering-demo
            namespace: open-cluster-management-agent-addon
          data:
            greeting: "{{GREETING}}"
EOF
```

2. Create two `AddOnDeploymentConfig` resources with different values:

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: AddOnDeploymentConfig
metadata:
  name: greeting-hello
  namespace: cluster1
spec:
  customizedVariables:
    - name: GREETING
      value: "hello"
---
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: AddOnDeploymentConfig
metadata:
  name: greeting-goodbye
  namespace: cluster1
spec:
  customizedVariables:
    - name: GREETING
      value: "goodbye"
EOF
```

3. Create a `ClusterManagementAddOn` referencing the template (no global
   ADC default, so per-cluster configs are the only source):

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ClusterManagementAddOn
metadata:
  name: ordering-demo
  annotations:
    addon.open-cluster-management.io/lifecycle: "addon-manager"
spec:
  addOnMeta:
    displayName: Config Ordering Demo
  installStrategy:
    type: Manual
  supportedConfigs:
    - group: addon.open-cluster-management.io
      resource: addontemplates
      defaultConfig:
        name: ordering-demo
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
EOF
```

4. Create the `ManagedClusterAddOn` with two ADC references — `hello`
   first, `goodbye` second:

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ManagedClusterAddOn
metadata:
  name: ordering-demo
  namespace: cluster1
spec:
  configs:
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
      name: greeting-hello
      namespace: cluster1
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
      name: greeting-goodbye
      namespace: cluster1
EOF
```

5. Wait for reconciliation, then inspect `status.configReferences`:

```bash
kubectl get managedclusteraddons ordering-demo -n cluster1 \
  -o jsonpath='{range .status.configReferences[*]}{.resource}: {.desiredConfig.name}{"\n"}{end}'
```

Note the order — `greeting-hello` appears before `greeting-goodbye`.

6. Re-order the configs to put `goodbye` first:

```bash
kubectl apply -f - <<'EOF'
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ManagedClusterAddOn
metadata:
  name: ordering-demo
  namespace: cluster1
spec:
  configs:
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
      name: greeting-goodbye
      namespace: cluster1
    - group: addon.open-cluster-management.io
      resource: addondeploymentconfigs
      name: greeting-hello
      namespace: cluster1
EOF
```

7. Wait for reconciliation and inspect again:

```bash
kubectl get managedclusteraddons ordering-demo -n cluster1 \
  -o jsonpath='{range .status.configReferences[*]}{.resource}: {.desiredConfig.name}{"\n"}{end}'
```

The order in `status.configReferences` is unchanged — `greeting-hello`
still appears before `greeting-goodbye`, despite the spec now listing
`goodbye` first.

Cleanup:

```bash
kubectl delete clustermanagementaddons ordering-demo
kubectl delete addontemplates ordering-demo
kubectl delete addondeploymentconfigs greeting-hello greeting-goodbye -n cluster1
```

**Expected behavior**

When the user re-orders entries in `ManagedClusterAddOn.spec.configs`,
the order in `status.configReferences` should be updated to reflect the
new ordering, so that precedence changes take effect.

- OCM addon-manager (all versions with multi-GK config support)
- Kubernetes 4.x / OpenShift 4.x

**Additional context**

The root cause is in `mergeAddonConfig` in
`pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler.go`:

```go
mergedConfigs := make(addonConfigMap)
// First go through the configReferences listed in mca status,
// if the existing config (gvk + namespace + name) is also in the desiredConfigMap,
// append it to mergedConfigs,
// this will save the LastAppliedConfig and LastObservedGeneration from mca status.
for _, configRef := range mcaCopy.Status.ConfigReferences {
    gr := configRef.ConfigGroupResource
    if _, ok := mergedConfigs[gr]; !ok {
        mergedConfigs[gr] = []addonv1beta1.ConfigReference{}
    }
    if _, ok := desiredConfigMap.containsConfig(gr, configRef.DesiredConfig.ConfigReferent); ok {
        mergedConfigs[gr] = append(mergedConfigs[gr], configRef)
    }
}
// Then go through the desiredConfigMap ...
```

Because the old `status.configReferences` are iterated first and their
order is preserved, any re-ordering in the desired config map is ignored
for entries that already exist. New entries are appended at the end, but
existing entries keep their original positions.

The fix is to use the `desiredConfigMap` as the source of truth for
ordering, while preserving `LastAppliedConfig` and
`LastObservedGeneration` via a lookup map keyed by
group+resource+name+namespace:

```go
// Build a lookup from old status to preserve rollout tracking fields
oldStatusByKey := map[configKey]addonv1beta1.ConfigReference{}
for _, configRef := range mcaCopy.Status.ConfigReferences {
    key := configKey{...}
    oldStatusByKey[key] = configRef
}

// Iterate desiredConfigMap in sorted GR order — source of truth for ordering
configRefs := []addonv1beta1.ConfigReference{}
for _, gr := range desiredConfigMap.orderedKeys() {
    for _, configRef := range desiredConfigMap[gr] {
        key := configKey{...}
        if old, ok := oldStatusByKey[key]; ok {
            old.DesiredConfig = configRef.DesiredConfig.DeepCopy()
            configRefs = append(configRefs, old)
        } else {
            configRefs = append(configRefs, configRef)
        }
    }
}
```

**Related issue:** There is a separate bug in addon-framework where
`GetAddOnConfigRef` returns the first match instead of the last, which
compounds this issue. See https://github.com/open-cluster-management-io/addon-framework/issues/387

KevinFCormier:
## Summary
After ManagedClusterAddOn status.configReferences are initially computed, the order is not updated if the order of configs in the spec changes. This PR corrects that.

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/ocm/issues/1600

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved rollout tracking details when addon configuration references are refreshed, reordered, or updated.
  * Improved stability of `status.configReferences` ordering: if the observed order doesn’t match the desired order within a config group, the addon is set to require applying updates instead of reporting success.
  * Prevented potential nil-related failures by avoiding updates when a desired config reference is not present.

* **Tests**
  * Added unit tests covering config reference merge behavior and reconciliation patching when `spec.configs` ordering changes.

### DIFF
--- a/pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler.go
+++ b/pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler.go
@@ -86,44 +86,46 @@ func (d *managedClusterAddonConfigurationReconciler) mergeAddonConfig(
 	mca *addonv1beta1.ManagedClusterAddOn, desiredConfigMap addonConfigMap) *addonv1beta1.ManagedClusterAddOn {
 	mcaCopy := mca.DeepCopy()
 
-	mergedConfigs := make(addonConfigMap)
-	// First go through the configReferences listed in mca status,
-	// if the existing config (gvk + namespace + name) is also in the desiredConfigMap, append it to mergedConfigs,
-	// this will save the LastAppliedConfig and LastObservedGeneration from mca status.
+	type configKey struct {
+		gr      addonv1beta1.ConfigGroupResource
+		refName string
+		refNs   string
+	}
+
+	// Build a lookup from old status to preserve LastAppliedConfig and LastObservedGeneration
+	oldStatusByKey := map[configKey]addonv1beta1.ConfigReference{}
 	for _, configRef := range mcaCopy.Status.ConfigReferences {
-		gr := configRef.ConfigGroupResource
-		if _, ok := mergedConfigs[gr]; !ok {
-			mergedConfigs[gr] = []addonv1beta1.ConfigReference{}
+		if configRef.DesiredConfig == nil {
+			continue
 		}
-		if _, ok := desiredConfigMap.containsConfig(gr, configRef.DesiredConfig.ConfigReferent); ok {
-			mergedConfigs[gr] = append(mergedConfigs[gr], configRef)
+		key := configKey{
+			gr:      configRef.ConfigGroupResource,
+			refName: configRef.DesiredConfig.ConfigReferent.Name,
+			refNs:   configRef.DesiredConfig.ConfigReferent.Namespace,
 		}
+		oldStatusByKey[key] = configRef
 	}
 
-	// Then go through the desiredConfigMap, for each configReference,
-	// if the desired config (gvk + namespace + name) is aleady in the mergedConfigs,
-	// update the ConfigReferent and DesiredConfig (including desired spechash) to mergedConfigs.
-	// else just append it to the mergedConfigs.
-	for gr, configReferences := range desiredConfigMap {
-		for _, configRef := range configReferences {
-			if _, ok := mergedConfigs[gr]; !ok {
-				mergedConfigs[gr] = []addonv1beta1.ConfigReference{}
-			}
-
-			referent := configRef.DesiredConfig.ConfigReferent
-			if i, exist := mergedConfigs.containsConfig(gr, referent); exist {
-				mergedConfigs[gr][i].DesiredConfig = configRef.DesiredConfig.DeepCopy()
-			} else {
-				mergedConfigs[gr] = append(mergedConfigs[gr], configRef)
-			}
-		}
-	}
-
-	// sort by gvk and set the final config references
+	// Iterate desiredConfigMap in sorted GR order — this is the source of truth for ordering.
+	// Within each GR, the slice order from the desired config map is preserved.
 	configRefs := []addonv1beta1.ConfigReference{}
-	for _, gvk := range mergedConfigs.orderedKeys() {
-		configRefs = append(configRefs, mergedConfigs[gvk]...)
+	for _, gr := range desiredConfigMap.orderedKeys() {
+		for _, configRef := range desiredConfigMap[gr] {
+			key := configKey{
+				gr:      gr,
+				refName: configRef.DesiredConfig.ConfigReferent.Name,
+				refNs:   configRef.DesiredConfig.ConfigReferent.Namespace,
+			}
+			if old, ok := oldStatusByKey[key]; ok {
+				// Preserve rollout tracking fields, update desired config
+				old.DesiredConfig = configRef.DesiredConfig.DeepCopy()
+				configRefs = append(configRefs, old)
+			} else {
+				configRefs = append(configRefs, configRef)
+			}
+		}
 	}
+
 	mcaCopy.Status.ConfigReferences = configRefs
 	return mcaCopy
 }
--- a/pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler_test.go
+++ b/pkg/addon/controllers/addonconfiguration/addon_configuration_reconciler_test.go
@@ -461,6 +461,109 @@ func TestAddonConfigReconcile(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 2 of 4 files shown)


mikeshng (commented):
@KevinFCormier sorry I am late to the party. Trying to get this PR merge and I think it's ready to go in.

The only question I have is "why does it matter"? Ok the order in the status is "wrong" based on the spec but why does that matter? Did the API say it should be the correct order? Usually, we want to avoid unnecessary updates for performance so that's why I am asking.

KevinFCormier:
> @KevinFCormier sorry I am late to the party. Trying to get this PR merge and I think it's ready to go in.
> 
> The only question I have is "why does it matter"? Ok the order in the status is "wrong" based on the spec but why does that matter? Did the API say it should be the correct order? Usually, we want to avoid unnecessary updates for performance so that's why I am asking.

Thanks for the question @mikeshng. I think I left out a important detail in the example outlined in the ticket: https://github.com/open-cluster-management-io/ocm/issues/1600

The example creates a `ConfigMap` called `ordering-demo` in the `open-cluster-management-agent-addon` namespace. Initially, it will have the greeting "hello" in its data, because that `AddOnDeploymentConfig` appears first in the spec of the `ManagedClusterAddOn`. When we reverse the order in the spec and the status does not update, it's not only that status that does not update, but the contents of the `ConfigMap` is not updated either. The actual configs used by the `AddOnTemplate` will come from the status, not the spec. You would have to actually delete the `AddOnDeploymentConfig` with the "hello" greeting to have the "goodbye" greeting appear in the `ConfigMap`. So it is a real functional bug - you could have to setups with identical specs but different results because of the history of how they were created.

mikeshng (approved):
> You would have to actually delete the AddOnDeploymentConfig with the "hello" greeting to have the "goodbye" greeting appear in the ConfigMap. So it is a real functional bug

:O thank you for the clarification and the fix!

/approve

/lgtm

Issue #1596, opened 2026-06-25
ManifestWorkReplicaSet silently fails when `namespace.name` exceeds the 63-char label limit


krishnankm:
**Describe the enhancement**

The MWRS controller stamps every generated `ManifestWork` with `work.open-cluster-management.io/manifestworkreplicaset = <namespace>.<name>` (set in `buildManifestWork`, value from `manifestWorkReplicaSetKey`). This value is applied verbatim with no length validation or truncation. When an MWRS's `namespace` + `.` + `name` exceeds 63 characters, the kube-apiserver rejects the `ManifestWork` create with an invalid-label-value error.

**Impact**
- The `ManifestWork` is never created on the affected placement(s), so the workload silently never deploys.
- The error is aggregated and requeued, but is **not** reflected in the MWRS status. `ManifestworkApplied` shows only a generic `NotAsExpected`/`Processing`. The root cause (invalid label length) appears only in controller logs, which most consumers cannot access.

**Reproduce**
Create an MWRS where `len(namespace) + 1 + len(name) > 63` and a Placement that selects ≥1 cluster. The MWRS reports not-ready; no ManifestWork appears; the reason is only in the controller log.

**Requests**
1. **Don't break on long refs.** Derive the owner-label value so it always fits 63 chars — e.g. a deterministic hash suffix (`<truncated>-<hash>`), with the full `namespace/name` preserved in an annotation for traceability. Keep selectors working via the hashed value.
2. **Surface the failure.** Validate name/namespace length up front (admission webhook on MWRS) and/or write a specific MWRS  status condition with the actual apply error (e.g. `Reason: InvalidManifestWorkName`, message naming the 63-char limit), so users don't need controller logs to diagnose it.

**Environment:** `open-cluster-management.io/api v1.2.0`

krishnankm:
I'd like to contribute the observability half (Request 2) of this since it is self-contained and does not affect the wire format.

My proposal is to fail fast in `deployReconciler.reconcile`. Before building or applying ManifestWorks, we can validate `manifestWorkReplicaSetKey(mwrSet)` using `validation.IsValidLabelValue`. If the value is invalid (for example, the >63-character case), we would set the `ManifestworkApplied` condition to `False` with an actionable message that identifies the offending value and references the 63-character limit, then `reconcileStop` to avoid repeatedly attempting failed applies. This keeps the entire change within the ocm repository and avoids any API changes.

[...]

PR #1602, opened 2026-06-26
:sparkles: feat: surface invalid ManifestWork owner label on MWRS status

krishnankm:
## Summary

The ManifestWorkReplicaSet controller stamps every generated ManifestWork with the ownervlabel `work.open-cluster-management.io/manifestworkreplicaset = "<namespace>.<name>"` and applies it without length validation. When `namespace.name` exceeds the 63-char label-value limit, the API server rejects each ManifestWork apply, but the MWRS status only shows a generic `ManifestworkApplied=False` (`NotAsExpected`/`Processing`), the cause is reachable only in controller logs.

This validates the owner value at the start of `deployReconciler.reconcile`. If it is not a valid label value, it sets `ManifestworkApplied=False` with a new `InvalidManifestWorkName` reason and an actionable message, then stops (namespace and name are immutable, so retrying can never succeed). The reason is defined locally in the controller package, so no `open-cluster-management.io/api` change is needed.

Implements **request 2** of the linked issue. Request 1 (hashing the owner label) will follow in a separate PR.

## Related issue(s)

Part of #1596 

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for the owner label value used during ManifestWorkReplicaSet reconciliation.
  * If the label value is invalid (for example, exceeds Kubernetes length limits), reconciliation stops without making changes and updates the `ManifestworkApplied` status condition with a clear error message.
* **Tests**
  * Added a unit test covering the “owner label too long” case to ensure no ManifestWork objects are created.

### DIFF
--- a/pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
+++ b/pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
@@ -3,12 +3,14 @@ package manifestworkreplicasetcontroller
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation"
 
 	clusterlister "open-cluster-management.io/api/client/cluster/listers/cluster/v1beta1"
 	worklisterv1 "open-cluster-management.io/api/client/work/listers/work/v1"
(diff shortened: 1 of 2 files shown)


pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
```diff
@@ -37,6 +42,16 @@ func (d *deployReconciler) reconcile(
 	minRequeue := maxRequeueTime
 	count, total, succeededCount := 0, 0, 0
 
+	// Report invalid ManifestWork owner labels in status.
+	ownerValue := manifestWorkReplicaSetKey(mwrSet)
+	if verrs := validation.IsValidLabelValue(ownerValue); len(verrs) > 0 {
```
qiujian16, line 49: can we add a TODO here? my understanding is we will not need this validation and the label when we use hash value

krishnankm: Thanks @qiujian16 , added a TODO right there marking this validation (and the label) for removal once we switch the owner label to a hash value (#1596). Pushed.

krishnankm:
Hi @elgnay @qiujian16 @zhujian7 Could you review the PR and also take a look at the failing integration check ? I dont think its because of the PR. 🙇

PR #585, opened 2024-08-01
[WIP] :sparkles: addon: add support for multiple GVK

haoqing0110:
## Related issue(s)

Ref: https://github.com/open-cluster-management-io/ocm/pull/547

PR #1610, opened 2026-06-30
:bug: fix: preserve config reference ordering in cmainstallprogression controller

### LINKED ISSUE #1609: :bug: Nondeterministic ordering of same-GVK ConfigReferences in CMA InstallProgressions
The `cmaInstallProgressionController` uses a `sets.Set[ConfigReferent]` (a Go map) to collect per-placement configs in `setInstallProgression` (`controller.go:133`). When a placement has multiple configs of the same GVK, iterating the set produces a random order on each reconcile.

This causes:
- **Status thrashing** — the patcher sees a different slice order and issues unnecessary PATCH requests every reconcile
- **Nondeterministic config consumption** — which config is "first" in the MCA status changes randomly, affecting addon managers that read configs by position

### Example

A CMA placement with two `prometheusrules` (e.g., alerting-rules and recording-rules) will have their order in `status.installProgressions[].configReferences` change randomly between reconciles.

### Context

Introduced in dfa835d67 (#585, "addon: add support for multiple GVK"). Before that commit, only one config per GVK was supported so ordering was not an issue. The equivalent code in the addonconfiguration controller (`graph.go:644-670`) already uses a slice with dedup instead of a set, which preserves insertion order.

tesshuflower:
Fixes https://github.com/open-cluster-management-io/ocm/issues/1609

## Problem

The `cmaInstallProgressionController` uses a `sets.Set[ConfigReferent]` (a Go map) to collect per-placement configs in `setInstallProgression`. When a placement has multiple configs of the same GVK, iterating the set produces a random order on each reconcile.

This causes:
- **Status thrashing** — the patcher uses `equality.Semantic.DeepEqual` (order-sensitive on slices) and issues unnecessary PATCH requests every reconcile
- **Nondeterministic config consumption** — which config is "first" in the MCA status changes randomly, affecting addon managers that iterate same-GVK configs in order

Introduced in dfa835d67 (#585, "addon: add support for multiple GVK"). The equivalent code in the addonconfiguration controller (`graph.go:644-670`) already uses a slice with dedup instead of a set, which preserves insertion order.

## Solution

Replace `sets.Set[ConfigReferent]` with `[]ConfigReferent` and a manual dedup check before appending in `overrideConfigMapByAddOnConfigs`, matching the pattern already used in `pkg/addon/controllers/addonconfiguration/graph.go`. This preserves the spec-defined ordering from the CMA placement `configs` list.

## Changes

- `pkg/addon/controllers/cmainstallprogression/controller.go` — replace set with slice+dedup, add `containsConfigReferent` helper, remove `sets` import
- `pkg/addon/controllers/cmainstallprogression/controller_test.go` — add ordering assertions to existing multi-same-GVK test cases
- `test/integration/addon/addon_configs_test.go` — new integration test verifying the full pipeline (CMA spec → CMA status → MCA status) preserves config ordering for multiple same-GVK configs in a placement

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the order of install progression config references when multiple entries share the same resource type, ensuring status output matches the configured sequence.
  * Improved duplicate config handling to prevent repeated config references in install progression/status results.

* **Tests**
  * Strengthened controller tests to assert exact ordering of config references (and clearer failure conditions) for placements with multiple same-GVK configs.
  * Added an integration test verifying ordering is preserved end-to-end, including on managed cluster add-on status.

### DIFF
--- a/pkg/addon/controllers/cmainstallprogression/controller.go
+++ b/pkg/addon/controllers/cmainstallprogression/controller.go
@@ -5,7 +5,6 @@ import (
 	"sort"
 
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
 	addonv1beta1 "open-cluster-management.io/api/addon/v1beta1"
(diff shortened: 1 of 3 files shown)


tesshuflower:
@coderabbitai review

rokej (approved):

tesshuflower:
@qiujian16 Would you mind reviewing this one?

PR #1623, opened 2026-07-13
:bug: fix(placement): use correct SAR verb on debug permission checks

### LINKED ISSUE #1622: Bug: PlacementDebugServer returns HTTP 200 when user lacks permission
## Describe the bug

When querying the PlacementDebugServer debug endpoint (`/debug/placements/`) as a user who lacks sufficient RBAC permissions, the server returns HTTP **200 OK** even though the JSON response body contains a permission-denied error. Callers should receive **403 Forbidden** so scripts, operators, and monitoring can detect authorization failures without parsing the response body.

A related problem: permission checks always evaluate the **`create`** verb on `placements`, regardless of HTTP method. A GET request for an existing placement is therefore authorized against placement **create** permission, producing a misleading error message (`user does not have permission to create placements in namespace …`) instead of checking **get** access.

## Steps to reproduce

1. Install cluster-manager with the placement debug server enabled.
2. Create a ServiceAccount and grant it access only to the debug endpoint (not to create/get Placement CRs):

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: placement-debug-only
rules:
- nonResourceURLs: ["/debug/placements/*"]
  verbs: ["get", "post"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: placement-debug-only
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: placement-debug-only
subjects:
- kind: ServiceAccount
  name: debug-user
  namespace: open-cluster-management-hub
```

3. Obtain the ServiceAccount token (`${TOKEN}`).
4. Exec into the placement controller pod and send a POST request:

```bash
oc exec -n open-cluster-management-hub \
  deploy/cluster-manager-placement-controller \
  -- curl -k -s -w "\nHTTP: %{http_code}\n" -X POST \
  "https://cluster-manager-placement.open-cluster-management-hub.svc:9443/debug/placements/" \
  -H "Authorization: Bearer ${TOKEN}" \
  -H "Content-Type: application/json" \
  -d '{"apiVersion":"cluster.open-cluster-management.io/v1beta1","kind":"Placement","metadata":{"name":"repro-test","namespace":"default"},"spec":{"numberOfClusters":1}}'
```

5. Repeat with a GET request for an existing placement:

```bash
oc exec -n open-cluster-management-hub \
  deploy/cluster-manager-placement-controller \
  -- curl -k -s -w "\nHTTP: %{http_code}\n" -X GET \
  "https://cluster-manager-placement.open-cluster-management-hub.svc:9443/debug/placements/default/my-placement" \
  -H "Authorization: Bearer ${TOKEN}"
```

## Actual results

```
{"error":"user does not have permission to create placements in namespace default: "}
HTTP: 200
```

The same misleading `create` error message and HTTP 200 are returned for both POST and GET.

## Expected results

```
{"error":"user does not have permission to create placements in namespace default: "}
HTTP: 403
```

For GET requests, the permission check and error message should refer to **get** access (e.g. `user does not have permission to get placements in namespace default`).

## Root cause

In `pkg/placement/debugger/debugger.go`:

1. **Missing HTTP status on permission errors**: The permission-denied path writes a JSON error body without setting an appropriate HTTP status code first. In Go's `net/http`, writing the body without a preceding `WriteHeader` defaults to **200 OK**.

2. **Hardcoded SAR verb**: `checkPermission` always submits a SubjectAccessReview with `Verb: "create"`, even for GET requests. The HTTP method is not mapped to the corresponding Kubernetes RBAC verb.

3. **Test gap**: Integration tests in `test/integration/placement/debugger_test.go` assert only the JSON `error` field for permission-denied cases, not `resp.StatusCode`, so an incorrect 200 response is not caught in CI.

## Suggested fix

1. Return **403 Forbidden** (or **401 Unauthorized** when user context is missing) on permission-denied responses, setting the status code before writing the JSON body.

2. Map HTTP method to SAR verb in `checkPermission`:
   - `GET` → `get`
   - `POST` → `create`

3. Add status-code assertions to integration and unit tests for permission-denied scenarios.

## Environment

- OCM cluster-manager placement controller (placement debug server)
- Kubernetes / OpenShift 4.x

rokej:
## Summary

- Map PlacementDebugServer permission checks to the correct Kubernetes RBAC verb: `GET` requests use `get`, `POST` requests use `create`
- Return accurate permission-denied error messages (e.g. `get placements` vs `create placements`)
- Reject unsupported HTTP methods with **405 Method Not Allowed**
- Add test coverage asserting **403 Forbidden** on permission denial in unit and integration tests
- Free unused disk on e2e runners (dotnet/android/CodeQL/etc.) before image builds to reduce `no space left on device` flakes

Fixes #1622

## Test plan

- [x] `go test ./pkg/placement/debugger/... -count=1`
- [ ] Integration: `test/integration/placement/debugger_test.go` permission-denied case
- [ ] E2E jobs reclaim disk and complete without runner disk exhaustion

## Summary by CodeRabbit

* **Bug Fixes**
  * Debug endpoints now handle `GET` and `POST` with method-specific permission checks.
  * Unsupported HTTP methods now return `405 Method Not Allowed`.
  * Permission-denied responses more precisely indicate the required operation.
* **Tests**
  * Expanded/updated unit and integration coverage for verb-specific authorization behavior and handler rejection.
* **Chores**
  * E2E workflows now reclaim runner disk space before running tests to improve reliability.

### DIFF
--- a/pkg/placement/debugger/debugger.go
+++ b/pkg/placement/debugger/debugger.go
@@ -80,16 +80,15 @@ func (d *Debugger) Handler(w http.ResponseWriter, r *http.Request) {
 	var placement *clusterv1beta1.Placement
 	var err error
 
-	// Support both GET (fetch from API) and POST (accept JSON body)
-	if r.Method == http.MethodPost {
+	switch r.Method {
+	case http.MethodPost:
 		// POST: Parse Placement from request body
 		placement, err = d.parsePlacementFromBody(r)
 		if err != nil {
 			d.reportErr(w, http.StatusBadRequest, err)
 			return
 		}
-	} else {
-		// GET: Fetch Placement from API (original behavior)
+	case http.MethodGet:
 		namespace, name, err := d.parsePath(r.URL.Path)
 		if err != nil {
 			d.reportErr(w, http.StatusBadRequest, err)
@@ -109,9 +108,12 @@ func (d *Debugger) Handler(w http.ResponseWriter, r *http.Request) {
 			}
 			return
 		}
+	default:
+		d.reportErr(w, http.StatusMethodNotAllowed, fmt.Errorf("method %s not allowed", r.Method))
+		return
 	}
 
-	// Check if user has permission to create placements in this namespace
+	// Check if user has permission to access placements in this namespace
 	if err := d.checkPermission(r, placement.Namespace); err != nil {
 		d.reportPermissionErr(w, err)
 		return
@@ -209,11 +211,24 @@ func (d *Debugger) reportPermissionErr(w http.ResponseWriter, err error) {
 		statusCode = http.StatusUnauthorized
 	case strings.Contains(msg, "does not have permission"):
 		statusCode = http.StatusForbidden
+	case strings.Contains(msg, "unsupported method"):
+		statusCode = http.StatusMethodNotAllowed
 	}
 	d.reportErr(w, statusCode, err)
 }
 
-// checkPermission checks if the user has permission to create placements in the namespace using SAR
+func placementPermissionVerb(r *http.Request) (string, error) {
+	switch r.Method {
+	case http.MethodPost:
+		return "create", nil
+	case http.MethodGet:
+		return "get", nil
+	default:
+		return "", fmt.Errorf("unsupported method %s", r.Method)
+	}
+}
+
+// checkPermission checks if the user has permission to access placements in the namespace using SAR
 func (d *Debugger) checkPermission(r *http.Request, namespace string) error {
 	// Get user from request context (authenticated by GenericAPIServer)
 	ctx := r.Context()
@@ -229,7 +244,12 @@ func (d *Debugger) checkPermission(r *http.Request, namespace string) error {
 		extra[k] = authorizationv1.ExtraValue(v)
 	}
 
-	// Create SubjectAccessReview to check if user can create placements
+	verb, err := placementPermissionVerb(r)
+	if err != nil {
+		return err
+	}
+
+	// Create SubjectAccessReview to check if user can access placements
 	sar := &authorizationv1.SubjectAccessReview{
 		Spec: authorizationv1.SubjectAccessReviewSpec{
 			User:   username,
@@ -237,7 +257,7 @@ func (d *Debugger) checkPermission(r *http.Request, namespace string) error {
 			Extra:  extra,
 			ResourceAttributes: &authorizationv1.ResourceAttributes{
 				Namespace: namespace,
-				Verb:      "create",
+				Verb:      verb,
 				Group:     "cluster.open-cluster-management.io",
 				Version:   "v1beta1",
 				Resource:  "placements",
@@ -254,8 +274,8 @@ func (d *Debugger) checkPermission(r *http.Request, namespace string) error {
 	}
 
 	if !result.Status.Allowed {
-		return fmt.Errorf("user does not have permission to create placements in namespace %s: %s",
-			namespace, result.Status.Reason)
+		return fmt.Errorf("user does not have permission to %s placements in namespace %s: %s",
+			verb, namespace, result.Status.Reason)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


pkg/placement/debugger/debugger.go
```diff
@@ -213,7 +213,14 @@ func (d *Debugger) reportPermissionErr(w http.ResponseWriter, err error) {
 	d.reportErr(w, statusCode, err)
 }
 
-// checkPermission checks if the user has permission to create placements in the namespace using SAR
+func placementPermissionVerb(r *http.Request) string {
+	if r.Method == http.MethodPost {
+		return "create"
+	}
+	return "get"
```
mikeshng: I don't know the full context but this feels wrong.
We shouldn't default to get. It should map Get to get directly at minimal. 
What if it's Delete or Patch or other methods that is not Post or Get.

rokej: @mikeshng The placement dry-run is designed to support POST and GET only https://github.com/open-cluster-management-io/ocm/blob/main/pkg/placement/debugger/debugger.go#L83.

dhaiducek: I agree. I'd have to look at the surrounding logic, but it seems like the permission (get/create) should be explicitly passed into the `checkPermission()` function rather than assumed.

rokej: The whole point of this dry-run API is so that it does not actually create any kubernetes resources in the cluster.

mikeshng: I don't understand why the AI review bot is not picking this up. If there are only two possible options, the standard code convention for this type of stuff is:

```
if Post 
// do something
else if Get
// do something
else
//error
```

I mean we are literally fixing an issue that was cause by wrong assumptions so the fix to me should be let's be explicit whenenver possible.

@coderabbitai what's up with you? why are you not pointing out something like this?

rokej: @mikeshng @dhaiducek Can you please review the code again? Thanks.

dhaiducek:
LGTM! The current test failure was actually raised by CodeRabbit in its "outside the diff" comment. (You have to download the logs to see the failure.) The ServiceAccount is missing the `get` permission it needs for the test to pass.

https://github.com/open-cluster-management-io/ocm/blob/86f6019ccd2804729488a88ed3082b09a1c2da34/test/integration/placement/debugger_test.go#L103-L105
https://github.com/open-cluster-management-io/ocm/blob/86f6019ccd2804729488a88ed3082b09a1c2da34/test/integration/placement/debugger_test.go#L185

mikeshng (commented):
> The ServiceAccount is missing the get permission it needs for the test to pass.

This mean a fix is still needed? I am a bit confused. 

@coderabbitai why would you "hide" that critical fix needed by default? What do you even mean "outside the diff", if this PR causes something outside of the diff to break then we still need to know that so the PR can address it properly? Is something wrong or do I have the wrong assumptions?

dhaiducek:
The test RBAC needs to be fixed for the test to pass with the new logic flows since before this the `get` permission wasn't returning any error--only `create`, which is already part of the RBAC.

And I think the CodeRabbit thing is a GitHub limitation. I suspect CodeRabbit can't programmatically leave comments on lines that aren't part of the PR.

dhaiducek:
Interestingly, CodeRabbit also noted our current issues around not returning the correct HTTP status as a "nit" in a previous PR, but it went unnoticed:

>`195-201`: **Error responses always return HTTP 200 status.**
>
>The `reportErr` function returns errors with HTTP 200 OK status. Consider setting appropriate HTTP status codes (e.g., 400 for bad requests, 403 for permission denied, 500 for internal errors) to follow HTTP conventions and enable clients to handle errors correctly without parsing the JSON body.

ref: https://github.com/open-cluster-management-io/ocm/pull/1461#pullrequestreview-4018554554

mikeshng (approved):
/lgtm

/assign @qiujian16

PR #1627, opened 2026-07-20
:sparkles: feat: add NetworkPolicies for open-cluster-management-agent namespace

mihirlele:
## Summary

Ship 4 NetworkPolicy manifests with the klusterlet operator to restrict ingress/egress in the `open-cluster-management-agent` namespace, meeting CIS Kube benchmark 5.3.2.

### Policies added

| Policy | Purpose |
|--------|---------|
| `default-deny-all` | Baseline deny for all ingress and egress |
| `allow-dns-and-api` | DNS egress (OpenShift DNS on 5353, kube-dns on 53) + ports-only API server egress (TCP 443/6443) |
| `klusterlet` | Operator egress to intra-namespace pods, addon namespaces, and `kubernetes.default.svc` |
| `klusterlet-agent` | Agent egress to `kubernetes.default.svc`, hub webhooks (port 9443), and intra-namespace pods |

### Other changes

- **RBAC**: Added `networking.k8s.io/networkpolicies` permissions to the klusterlet ClusterRole (both static and Helm chart)
- **Cleanup**: Added `*networkingv1.NetworkPolicy` case to `CleanUpStaticObject` and `GenerateRelatedResource` in `helpers.go`
- **Static resource registration**: Added 4 NP file paths to `managementStaticResourceFiles` in the management reconciler
- **Tests**: Updated action counts in `TestSyncDeploy`, `TestSyncDeploySingleton`, `TestSyncDeployHosted`, `TestSyncDelete`, and `TestSyncDeleteHosted`

### Design decisions

- **Ports-only egress (no `to` selector) for API server access**: The `allow-dns-and-api` policy uses ports-only rules for TCP 443/6443 without restricting destination IPs or hostnames. This ensures the policy works on both hub (local-cluster) and spoke clusters without needing to know the hub API server address, and avoids breakage from cloud load balancer IP rotation.
- **Follows existing static resource pattern**: NPs are embedded via `//go:embed`, templated with `{{ .AgentNamespace }}`, and applied via `ApplyDirectly()` / `ApplyNetworkPolicy()` — consistent with how all other management static resources are handled.
- **Works across all deploy modes**: Singleton (combined `klusterlet-agent`), Default (separate registration/work agents), and Hosted modes are all covered by the pod selector labels used in each policy.

### Testing

- **Hub cluster (local-cluster)**: Custom operator image deployed, all 4 NPs created automatically via reconciliation, all pods healthy, zero errors
- **Spoke cluster**: NPs applied manually (same YAML), verified all agents maintained hub connectivity (`HubConnectionDegraded: False`), zero network errors in logs
- **Unit tests**: All updated test counts pass

## Summary by CodeRabbit

* **New Features**
  * Added optional NetworkPolicy support for klusterlet components, including a default-deny policy and targeted allow rules for DNS, API, and required agent/klusterlet communications.
* **Bug Fixes**
  * Updated operator and OLM RBAC so NetworkPolicy resources can be managed end-to-end, and reconciliation now applies/removes these policies only when the feature is enabled.
* **Tests**
  * Extended integration/unit coverage to enable the feature gate and validate the increased set of created related resources.

### DIFF
--- a/manifests/klusterlet/management/klusterlet-allow-dns-networkpolicy.yaml
+++ b/manifests/klusterlet/management/klusterlet-allow-dns-networkpolicy.yaml
@@ -1,0 +1,36 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-dns-and-api
+  namespace: {{ .AgentNamespace }}
+spec:
+  podSelector: {}
+  policyTypes:
+  - Egress
+  egress:
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-dns
+    ports:
+    - protocol: UDP
+      port: 5353
+    - protocol: TCP
+      port: 5353
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: kube-system
+      podSelector:
+        matchLabels:
+          k8s-app: kube-dns
+    ports:
+    - protocol: UDP
+      port: 53
+    - protocol: TCP
+      port: 53
+  - ports:
+    - protocol: TCP
+      port: 443
+    - protocol: TCP
+      port: 6443
--- a/manifests/klusterlet/management/klusterlet-networkpolicy.yaml
+++ b/manifests/klusterlet/management/klusterlet-networkpolicy.yaml
@@ -8,7 +8,11 @@ spec:
     matchLabels:
       app: klusterlet
   policyTypes:
+  - Ingress
   - Egress
+  ingress:
+  - from:
+    - podSelector: {}
   egress:
   - to:
     - podSelector: {}
--- a/manifests/klusterlet/management/klusterlet-networkpolicy.yaml
+++ b/manifests/klusterlet/management/klusterlet-networkpolicy.yaml
@@ -1,0 +1,28 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: klusterlet
+  namespace: {{ .AgentNamespace }}
+spec:
+  podSelector:
+    matchLabels:
+      app: klusterlet
+  policyTypes:
+  - Egress
+  egress:
+  - to:
+    - podSelector: {}
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          addon.open-cluster-management.io/namespace: "true"
+    ports:
+    - protocol: TCP
+      port: 443
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: default
+    ports:
+    - protocol: TCP
+      port: 443
--- a/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_management_recocile.go
+++ b/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_management_recocile.go
@@ -70,7 +70,8 @@ func (r *managementReconcile) reconcile(ctx context.Context, klusterlet *operato
 		return klusterlet, reconcileStop, err
 	}
 
-	files := managementStaticResourceFiles
+	var files []string
+	files = append(files, managementStaticResourceFiles...)
 	if config.NetworkPoliciesEnabled {
 		files = append(files, networkPolicyFiles...)
 	}
@@ -129,7 +130,8 @@ func (r *managementReconcile) clean(ctx context.Context, klusterlet *operatorapi
 	}
 
 	// remove static file on the management cluster
-	files := managementStaticResourceFiles
+	var files []string
+	files = append(files, managementStaticResourceFiles...)
 	if config.NetworkPoliciesEnabled {
 		files = append(files, networkPolicyFiles...)
 	}
--- a/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_management_recocile.go
+++ b/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_management_recocile.go
@@ -36,6 +36,10 @@ var (
 		"klusterlet/management/klusterlet-work-role.yaml",
 		"klusterlet/management/klusterlet-work-rolebinding.yaml",
 		"klusterlet/management/klusterlet-work-rolebinding-extension-apiserver.yaml",
+		"klusterlet/management/klusterlet-default-deny-all-networkpolicy.yaml",
+		"klusterlet/management/klusterlet-allow-dns-networkpolicy.yaml",
+		"klusterlet/management/klusterlet-networkpolicy.yaml",
+		"klusterlet/management/klusterlet-agent-networkpolicy.yaml",
 	}
 )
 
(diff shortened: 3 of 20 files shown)


pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_management_recocile.go
```diff
@@ -36,6 +36,10 @@ var (
 		"klusterlet/management/klusterlet-work-role.yaml",
 		"klusterlet/management/klusterlet-work-rolebinding.yaml",
 		"klusterlet/management/klusterlet-work-rolebinding-extension-apiserver.yaml",
+		"klusterlet/management/klusterlet-default-deny-all-networkpolicy.yaml",
```
qiujian16, line 42: to ensure backward compatible, I'd suggest we add a flag to enable network policies

mihirlele: @qiujian16 Thanks for the suggestion. However, this is a CIS Kubernetes Benchmark 5.3.2 requirement (OCPSTRAT-819) that needs to be enforced for OpenShift 5.0. Making it opt-in would mean clusters are non-compliant by default, which defeats the security hardening goal.

If there are concerns about backward compatibility, we could default the feature gate to Enable so that NPs are applied automatically but admins can disable them if needed. Would that address your concern?

qiujian16: I think you could add a flag or feature gate on operator controller and disable by default, you could enable it in openshift install

mikeshng (commented):
Thank you for the PR @mihirlele !

Please verify and address all of the coderabbitai review comments. Going forward, please avoid referencing internal tracking IDs such as OCPSTRAT-xyz, since not everyone has access to those systems in OCM-io.

pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_management_recocile.go
rokej: I agree we should find a way to have these NPs disabled by default.

mihirlele:
Thanks for the review feedback! Here's a summary of the changes I'm working on to address all comments:

### 1. Feature gate (per @qiujian16 and @rokej)

Added a `NetworkPolicies` feature gate, **disabled by default**. Users opt in via the Klusterlet CR:

```yaml
spec:
  registrationConfiguration:
    featureGates:
    - feature: NetworkPolicies
      mode: Enable
```

Implementation follows the existing `AboutAPIEnabled` / `ClusterProperty` pattern:
- Defined `NetworkPolicies` as an operator-internal feature gate constant
- Added `NetworkPoliciesEnabled` bool to `klusterletConfig`, set via `helpers.FeatureGateEnabled()` in `sync()`
- Moved the 4 NP files into a separate `networkPolicyFiles` slice, conditionally appended in both `reconcile()` and `clean()` based on the config flag
- Added `filterOperatorFeatureGates()` to strip operator-internal features before `ConvertToFeatureGateFlags`, so they don't produce invalid agent CLI flags or trigger a `ValidFeatureGates: False` condition

For OpenShift/ACM, the import controller can enable this feature gate by default when deploying the Klusterlet CR to managed clusters.

### 2. Missing ingress rules (per @coderabbitai)

Added `Ingress` to `policyTypes` and an intra-namespace ingress rule (`from: [{podSelector: {}}]`) to both `klusterlet-networkpolicy.yaml` and `klusterlet-agent-networkpolicy.yaml`. Without this, the `default-deny-all` policy blocks inbound traffic even though the sending pod has an egress rule — NetworkPolicies are directional.

### 3. Ports-only egress (per @coderabbitai)

The ports-only 443/6443 rule in `allow-dns-and-api` is intentional — it must work on both hub and spoke clusters without knowing the hub API server address, and avoids breakage from cloud load balancer IP rotation. The per-pod policies' port 443 rules to `kubernetes.default.svc` provide defense-in-depth and are harmless (NetworkPolicies are additive).

### 4. Internal tracking IDs (per @mikeshng)

Will remove OCPSTRAT references from the PR description.

### 5. Tests

- Reverted hardcoded test counts to pre-NP values (since the default is now disabled)
- Added `TestSyncDeployWithNetworkPolicies` that enables the feature gate and asserts the higher counts (24 + 4 NPs = 28)

All controller and helper tests pass.

mihirlele:
@qiujian16 can you PTAL?

qiujian16:
looks good in general, I did not look into details on the policies. I wonder if our e2e covers the policy cases, since I am not quite sure that the network plugin of kind cluster for e2e picks network policies.

rokej:
@mihirlele The network policies and the way they are disabled by default but can be enabled through klusterlet spec look good to me.

qiujian16:
please enable the feature gate in integration and e2e so it can be tested.

mihirlele:
@qiujian16 Thanks! its done.

mihirlele:
@qiujian16 can you please approve the testing? I think it should go through this time.

mihirlele:
seems like no pending action items. We may need to retry.

mikeshng (commented):
@mihirlele for the coderabbit review items could you reply to them after you addressed them? Especially the ones marked critical and major. Thanks!

manifests/klusterlet/management/klusterlet-networkpolicy.yaml
```diff
@@ -0,0 +1,28 @@
...
+  namespace: {{ .AgentNamespace }}
+spec:
+  podSelector:
+    matchLabels:
+      app: klusterlet
+  policyTypes:
+  - Egress
+  egress:
+  - to:
+    - podSelector: {}
```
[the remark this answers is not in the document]
mihirlele: Addressed in commit 63c2ebed3. Both `klusterlet-networkpolicy.yaml` and `klusterlet-agent-networkpolicy.yaml` now include `Ingress` in `policyTypes` and have an `ingress: [{from: [{podSelector: {}}]}]` block to allow intra-namespace communication.

manifests/klusterlet/management/klusterlet-allow-dns-networkpolicy.yaml
```diff
@@ -0,0 +1,36 @@
...
+    ports:
+    - protocol: UDP
+      port: 53
+    - protocol: TCP
+      port: 53
+  - ports:
+    - protocol: TCP
+      port: 443
+    - protocol: TCP
+      port: 6443
```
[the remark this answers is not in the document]
mihirlele: Intentional — no action taken. The hub API server address varies across deployments (cloud LBs, external clusters, EKS/GKE), so we can't pin a destination in the `to:` block. The per-workload policies (`klusterlet` and `klusterlet-agent`) provide defense in depth by scoping `kubernetes.default.svc` egress on port 443 to specific namespaces.

mihirlele:
Finding: Remove unconditional NP feature gate from CreateKlusterlet
  File: test/framework/klusterlet.go
  Verdict: Skip
  Reason: Intentional — enables broad e2e coverage across all klusterlet
  scenarios. Feature is opt-in in production.
  ────────────────────────────────────────
  Finding: Assert agent args don't contain NetworkPolicies=true
  File: klusterlet_test.go:80-87
  Verdict: Skip
  Reason: Existing args count check (== 7) already catches leakage — if the gate
  leaked, count would be 8 and the test would fail.
  ────────────────────────────────────────
  Finding: Fetch actual NPs and verify rules
  File: Multiple test files
  Verdict: Skip
  Reason: relatedResources count already validates the 4 NPs were applied. NP
  manifests are static YAML with no conditional logic worth asserting on at
  integration level.
  ────────────────────────────────────────
  Finding: Fix comment arithmetic
  File: klusterlet_test.go:154
  Verdict: Fixed
  Reason: Made the -2 duplicated SAs deduction explicit so the sum visibly equals

PR #1634, opened 2026-07-22
:sparkles: feat: add NetworkPolicies for open-cluster-management-hub namespace

amancse:
Added NetworkPolicy enforcement for open-cluster-management-hub namespace

Adds least-privilege NetworkPolicy enforcement for the open-cluster-management-hub namespace, applied and reconciled by the cluster-manager operator as part of the ClusterManager CR lifecycle.

  5 policies are managed:
  - default-deny-all (always-on baseline)
  - allow-egress (DNS + kube-apiserver combined, always-on)
  - allow-intra-namespace (always-on)
  - allow-webhook-ingress (always-on)
  - allow-prometheus-ingress (always-on)

## Related issue(s)

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Added feature-gated NetworkPolicy protection for the Cluster Manager hub namespace, including default-deny, internal communication, ingress, DNS/API access, and Prometheus scraping.
* **Security**
  * Granted required permissions to manage Kubernetes NetworkPolicy resources.
  * Simplified klusterlet egress rules while retaining required DNS and HTTPS access.
* **Bug Fixes**
  * Improved feature-gate handling so policies are applied only when enabled.
* **Tests**
  * Expanded rendering, reconciliation, integration, and end-to-end coverage.

### DIFF
--- a/deploy/cluster-manager/chart/cluster-manager/templates/cluster_role.yaml
+++ b/deploy/cluster-manager/chart/cluster-manager/templates/cluster_role.yaml
@@ -81,6 +81,10 @@ rules:
 - apiGroups: ["admissionregistration.k8s.io"]
   resources: ["validatingwebhookconfigurations", "mutatingwebhookconfigurations"]
   verbs: ["create", "get", "list", "update", "watch", "patch", "delete"]
+# Allow the registration-operator to manage NetworkPolicies
+- apiGroups: ["networking.k8s.io"]
+  resources: ["networkpolicies"]
+  verbs: ["create", "get", "list", "update", "watch", "patch", "delete"]
 # Allow the nuclues to manage clustermanager apis.
 - apiGroups: ["operator.open-cluster-management.io"]
   resources: ["clustermanagers"]
--- a/deploy/cluster-manager/config/rbac/cluster_role.yaml
+++ b/deploy/cluster-manager/config/rbac/cluster_role.yaml
@@ -83,6 +83,10 @@ rules:
 - apiGroups: ["admissionregistration.k8s.io"]
   resources: ["validatingwebhookconfigurations", "mutatingwebhookconfigurations"]
   verbs: ["create", "get", "list", "update", "watch", "patch", "delete"]
+# Allow the registration-operator to manage NetworkPolicies
+- apiGroups: ["networking.k8s.io"]
+  resources: ["networkpolicies"]
+  verbs: ["create", "get", "list", "update", "watch", "patch", "delete"]
 # Allow the nuclues to manage clustermanager apis.
 - apiGroups: ["operator.open-cluster-management.io"]
   resources: ["clustermanagers"]
--- a/manifests/cluster-manager/hub/networkpolicies/01-hub-ns-default-deny.yaml
+++ b/manifests/cluster-manager/hub/networkpolicies/01-hub-ns-default-deny.yaml
@@ -1,0 +1,22 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: default-deny-all
+  namespace: {{ .ClusterManagerNamespace }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: app
+        operator: In
+        values:
+          - clustermanager-registration-controller
+          - "{{ .ClusterManagerName }}-registration-webhook"
+          - "{{ .ClusterManagerName }}-work-controller"
+          - "{{ .ClusterManagerName }}-work-webhook"
+          - clustermanager-placement-controller
+          - clustermanager-addon-manager-controller
+          - "{{ .ClusterManagerName }}-addon-webhook"
+          - "{{ .ClusterManagerName }}-grpc-server"
+  policyTypes:
+    - Ingress
+    - Egress
--- a/manifests/cluster-manager/hub/networkpolicies/02-hub-ns-egress.yaml
+++ b/manifests/cluster-manager/hub/networkpolicies/02-hub-ns-egress.yaml
@@ -1,0 +1,45 @@
+# Combined egress policy for all hub controller and webhook pods.
+# Rule 1: DNS resolution (CoreDNS port 53, node-local-dns/openshift-dns port 5353)
+# Rule 2: kube-apiserver egress (ports 443, 6443)
+#
+# No "to" selector is specified on either rule — egress is restricted by port only.
+# DNS and apiserver endpoints vary by platform (kube-system vs openshift-dns,
+# ClusterIP vs static pod vs external control plane), and locking "to" would
+# require platform-specific namespace or IP configuration that OCM intentionally
+# avoids. The port restriction limits the blast radius to DNS and HTTPS traffic.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-egress
+  namespace: {{ .ClusterManagerNamespace }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: app
+        operator: In
+        values:
+          - clustermanager-registration-controller
+          - "{{ .ClusterManagerName }}-registration-webhook"
+          - "{{ .ClusterManagerName }}-work-controller"
+          - "{{ .ClusterManagerName }}-work-webhook"
+          - clustermanager-placement-controller
+          - clustermanager-addon-manager-controller
+          - "{{ .ClusterManagerName }}-addon-webhook"
+          - "{{ .ClusterManagerName }}-grpc-server"
+  policyTypes:
+    - Egress
+  egress:
+    - ports:
+        - port: 53
+          protocol: UDP
+        - port: 53
+          protocol: TCP
+        - port: 5353
+          protocol: UDP
+        - port: 5353
+          protocol: TCP
+    - ports:
+        - port: 443
+          protocol: TCP
+        - port: 6443
+          protocol: TCP
--- a/manifests/cluster-manager/hub/networkpolicies/03-hub-ns-intra-namespace.yaml
+++ b/manifests/cluster-manager/hub/networkpolicies/03-hub-ns-intra-namespace.yaml
@@ -1,0 +1,32 @@
+# Intra-namespace pod-to-pod communication.
+# Allows pods within open-cluster-management-hub to reach each other
+# (e.g. placement-controller debug server, inter-controller calls).
+# API-server webhook ingress is handled separately by 04-hub-ns-webhook-ingress.yaml.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-intra-namespace
+  namespace: {{ .ClusterManagerNamespace }}
+spec:
+  podSelector: &ocmPods
+    matchExpressions:
+      - key: app
+        operator: In
+        values:
+          - clustermanager-registration-controller
+          - "{{ .ClusterManagerName }}-registration-webhook"
+          - "{{ .ClusterManagerName }}-work-controller"
+          - "{{ .ClusterManagerName }}-work-webhook"
+          - clustermanager-placement-controller
+          - clustermanager-addon-manager-controller
+          - "{{ .ClusterManagerName }}-addon-webhook"
+          - "{{ .ClusterManagerName }}-grpc-server"
+  policyTypes:
+    - Ingress
+    - Egress
+  ingress:
+    - from:
+        - podSelector: *ocmPods
+  egress:
+    - to:
+        - podSelector: *ocmPods
--- a/manifests/cluster-manager/hub/networkpolicies/04-hub-ns-webhook-ingress.yaml
+++ b/manifests/cluster-manager/hub/networkpolicies/04-hub-ns-webhook-ingress.yaml
@@ -1,0 +1,42 @@
+# Ingress policy for all externally-reachable hub pods.
+# Always applied when NetworkPolicies feature gate is enabled — restricted by pod label and port only.
+# Intentionally allows ingress from all sources: webhook sources vary by platform (kube-system,
+# openshift-kube-apiserver, external control planes), and OCM cannot assume a platform-specific
+# source namespace. Port restriction limits the blast radius to webhook, gRPC, and debug ports.
+#
+# Covers:
+#   - registration/work/addon webhooks (apiserver admission calls)
+#   - grpc-server (spoke cluster connections, ports 8090/443)
+#   - placement debug server (port 9443)
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-hub-ingress
+  namespace: {{ .ClusterManagerNamespace }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: app
+        operator: In
+        values:
+          - "{{ .ClusterManagerName }}-registration-webhook"
+          - "{{ .ClusterManagerName }}-work-webhook"
+          - "{{ .ClusterManagerName }}-addon-webhook"
+          - "{{ .ClusterManagerName }}-grpc-server"
+          - clustermanager-placement-controller
+  policyTypes:
+    - Ingress
+  ingress:
+    - ports:
+        - port: {{ .RegistrationWebhook.Port }}
+          protocol: TCP
+        - port: {{ .WorkWebhook.Port }}
+          protocol: TCP
+        - port: {{ .AddonWebhook.Port }}
+          protocol: TCP
+        - port: 8090
+          protocol: TCP
+        - port: 443
+          protocol: TCP
+        - port: 9443
+          protocol: TCP
--- a/manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
+++ b/manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
@@ -1,0 +1,29 @@
+# Prometheus scrape ingress for all hub controller and webhook pods.
+# Allows ingress on metrics ports from any source within the cluster.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-prometheus-ingress
+  namespace: {{ .ClusterManagerNamespace }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: app
+        operator: In
+        values:
+          - clustermanager-registration-controller
+          - "{{ .ClusterManagerName }}-registration-webhook"
+          - "{{ .ClusterManagerName }}-work-controller"
+          - "{{ .ClusterManagerName }}-work-webhook"
+          - clustermanager-placement-controller
+          - clustermanager-addon-manager-controller
+          - "{{ .ClusterManagerName }}-addon-webhook"
+          - "{{ .ClusterManagerName }}-grpc-server"
+  policyTypes:
+    - Ingress
+  ingress:
+    - ports:
+        - port: 8080
+          protocol: TCP
+        - port: 8443
+          protocol: TCP
--- a/manifests/klusterlet/management/klusterlet-agent-networkpolicy.yaml
+++ b/manifests/klusterlet/management/klusterlet-agent-networkpolicy.yaml
@@ -14,18 +14,10 @@ spec:
   - from:
     - podSelector: {}
   egress:
-  - to:
-    - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: default
-    ports:
+  - ports:
     - protocol: TCP
       port: 443
-  - to:
-    - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: open-cluster-management-hub
-    ports:
+  - ports:
     - protocol: TCP
       port: 9443
   - to:
--- a/manifests/klusterlet/management/klusterlet-networkpolicy.yaml
+++ b/manifests/klusterlet/management/klusterlet-networkpolicy.yaml
@@ -16,17 +16,6 @@ spec:
   egress:
   - to:
     - podSelector: {}
-  - to:
-    - namespaceSelector:
-        matchLabels:
-          addon.open-cluster-management.io/namespace: "true"
-    ports:
-    - protocol: TCP
-      port: 443
-  - to:
-    - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: default
-    ports:
+  - ports:
     - protocol: TCP
       port: 443
--- a/pkg/operator/helpers/helpers.go
+++ b/pkg/operator/helpers/helpers.go
@@ -805,6 +805,28 @@ func FeatureGateEnabled(features []operatorapiv1.FeatureGate,
 	return defaultFeature.Default
 }
 
+// NetworkPolicies is an operator-internal feature gate that controls whether NetworkPolicy
+// manifests are applied to the hub/agent namespace. Not passed through to agent binaries.
+const NetworkPolicies featuregate.Feature = "NetworkPolicies"
+
+// operatorOnlyFeatureGates are feature gates consumed by the operator itself,
+// not passed through to agent binaries as CLI flags.
+var operatorOnlyFeatureGates = map[featuregate.Feature]bool{
+	NetworkPolicies: true,
+}
+
+// FilterOperatorFeatureGates removes operator-internal feature gates from the list
+// so they are not forwarded as CLI flags to hub/agent binaries.
+func FilterOperatorFeatureGates(features []operatorapiv1.FeatureGate) []operatorapiv1.FeatureGate {
+	var filtered []operatorapiv1.FeatureGate
+	for _, f := range features {
+		if !operatorOnlyFeatureGates[featuregate.Feature(f.Feature)] {
+			filtered = append(filtered, f)
+		}
+	}
+	return filtered
+}
+
 // IsSingleton returns if agent is deployed in singleton mode either hosted or not
 func IsSingleton(mode operatorapiv1.InstallMode) bool {
 	return mode == operatorapiv1.InstallModeSingleton || mode == operatorapiv1.InstallModeSingletonHosted
--- a/pkg/operator/helpers/helpers_test.go
+++ b/pkg/operator/helpers/helpers_test.go
@@ -17,6 +17,7 @@ import (
 	admissionv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	fakeapiextensions "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/fake"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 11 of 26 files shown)


pkg/operator/helpers/helpers.go
```diff
@@ -1028,3 +1031,23 @@ func GRPCServerEndpointType(cm *operatorapiv1.ClusterManager) string {
...
+	nodes, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	var cidrs []string
+	for _, node := range nodes.Items {
+		for _, addr := range node.Status.Addresses {
+			if addr.Type == corev1.NodeInternalIP {
+				cidrs = append(cidrs, addr.Address+"/32")
+			}
```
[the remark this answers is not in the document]
amancse: This part is removed

manifests/cluster-manager/hub/networkpolicies/03-hub-ns-kubelet-probe.yaml
```diff
@@ -0,0 +1,24 @@
...
+  ingress:
+    - ports:
+        - port: 8000
+          protocol: TCP
+        - port: 8443
+          protocol: TCP
+      from:
+{{- range .NodeCIDRs }}
+        - ipBlock:
+            cidr: {{ . }}
```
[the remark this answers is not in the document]
amancse: this implementation is removed

manifests/cluster-manager/hub/networkpolicies/02-hub-ns-egress.yaml
```diff
@@ -0,0 +1,28 @@
...
+          protocol: TCP
+        - port: 5353
+          protocol: UDP
+        - port: 5353
+          protocol: TCP
+    - ports:
+        - port: 443
+          protocol: TCP
+        - port: 6443
+          protocol: TCP
```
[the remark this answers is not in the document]
amancse: fixed documentation

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_hub_reconcile.go
```diff
@@ -147,6 +168,41 @@ func (c *hubReconcile) reconcile(ctx context.Context, cm *operatorapiv1.ClusterM
...
+			helpers.SetRelatedResourcesStatusesWithObj(ctx, &cm.Status.RelatedResources, objData)
+			return objData, nil
+		},
+		npFiles...,
+	)
+	for _, result := range npResults {
+		if result.Error != nil {
+			appliedErrs = append(appliedErrs, fmt.Errorf("%q (%T): %v", result.File, result.Type, result.Error))
+		}
+	}
```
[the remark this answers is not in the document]
amancse: added code to cleanup

manifests/cluster-manager/hub/networkpolicies/01-hub-ns-default-deny.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: default-deny-all
+  namespace: {{ .ClusterManagerNamespace }}
+spec:
+  podSelector: {}
```
rokej: This should select the cluster-manager related pods only.

manifests/cluster-manager/hub/networkpolicies/04-hub-ns-webhook-ingress.yaml
```diff
@@ -0,0 +1,34 @@
+# Combined webhook ingress policy for registration, work and addon webhooks.
+# All three webhooks are called by kube-apiserver from the same source namespace
+# on the same port — a single policy with matchExpressions covers all three.
+# Only applied when APIServerNamespace is set in ClusterManagerSpec.NetworkPolicyConfiguration.
```
rokej: With always-on deny, webhooks may not work if APIServerNamespace is not set.

manifests/cluster-manager/hub/networkpolicies/04-hub-ns-webhook-ingress.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - Ingress
+  ingress:
+    - ports:
+        - port: {{ .RegistrationWebhook.Port }}
+          protocol: TCP
+        - port: {{ .WorkWebhook.Port }}
+          protocol: TCP
+        - port: {{ .AddonWebhook.Port }}
+          protocol: TCP
+      from:
```
rokej: Wouldn't it be simpler to use port-only allow? Based on my research, the API server namespace often does not represent admission traffic the way the field implies.

manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: allow-prometheus-ingress
+  namespace: {{ .ClusterManagerNamespace }}
+spec:
+  podSelector:
+    matchExpressions:
+      - key: app
+        operator: In
+        values:
+          - "{{ .ClusterManagerName }}-registration-controller"
```
rokej: This pod label is wrong. https://github.com/open-cluster-management-io/ocm/blob/main/manifests/cluster-manager/management/registration/deployment.yaml#L21 It should be `clustermanager-registration-controller`.

manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - "{{ .ClusterManagerName }}-registration-webhook"
+          - "{{ .ClusterManagerName }}-work-controller"
+          - "{{ .ClusterManagerName }}-work-webhook"
+          - "{{ .ClusterManagerName }}-placement-controller"
```
rokej: https://github.com/open-cluster-management-io/ocm/blob/main/manifests/cluster-manager/management/placement/deployment.yaml#L21. It should be `clustermanager-placement-controller`

manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - "{{ .ClusterManagerName }}-addon-manager-controller"
```
rokej: https://github.com/open-cluster-management-io/ocm/blob/main/manifests/cluster-manager/management/addon-manager/deployment.yaml#L21 It should be `clustermanager-addon-manager-controller`

manifests/cluster-manager/hub/networkpolicies/03-hub-ns-intra-namespace.yaml
```diff
@@ -0,0 +1,20 @@
...
+# Allows pods within open-cluster-management-hub to reach each other
+# (e.g. placement-controller debug server, inter-controller calls).
+# API-server webhook ingress is handled separately by 04-hub-ns-webhook-ingress.yaml.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-intra-namespace
+  namespace: {{ .ClusterManagerNamespace }}
+spec:
+  podSelector: {}
```
rokej: Empty pod selector!

manifests/cluster-manager/hub/networkpolicies/04-hub-ns-webhook-ingress.yaml
amancse: The intent was to restrict webhook ingress to only the namespace where kube-apiserver pods run, so that arbitrary pods on the cluster couldn't directly reach the webhook endpoints.

Will remove the "from" block and make this policy always-on with port-only ingress which will resolve opt-in webhook NP as well for empty ns field (above comment).

rokej:
No network policies are defined for these optional pods. 

- grpc-server — Service ports 8090 / 443 (when gRPC auth is enabled)

- placement debug — Service port 9443 (when PlacementDebugServer is enabled)

Also, all these NetworkPolicies should be disabled by default and there should be a knob somewhere to opt in.

rokej:
@dhaiducek @mikeshng @qiujian16 @tesshuflower Could you please review this as well? Thank you.

vendor/open-cluster-management.io/api/operator/v1/types_clustermanager.go
tesshuflower, line 1: This will require changes to open-cluster-management-io/api to add the new NetworkPolicyiConfigturation to the ClusterManager CR, correct?  I didn't see a PR there for this yet.  I assume that change would be required before we merge this PR?

vendor/open-cluster-management.io/api/operator/v1/types_clustermanager.go
```diff
@@ -85,6 +85,32 @@ type ClusterManagerSpec struct {
...
+	MonitoringNamespace string `json:"monitoringNamespace,omitempty"`
+
+	// apiServerNamespace is the namespace from which kube-apiserver originates webhook calls.
+	// Used to restrict ingress on registration, work, and addon webhook pods to only accept
+	// requests from the API server. Leave unset to skip namespace-scoped webhook ingress restriction.
+	// Common values: "kube-system" (vanilla Kubernetes), "openshift-kube-apiserver" (OpenShift).
+	// +optional
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([a-z0-9\-]*[a-z0-9])?$`
+	APIServerNamespace string `json:"apiServerNamespace,omitempty"`
```
tesshuflower: We should probably be consistent with the way our other projects handle this (not saying this is not the correct way however).

From our downstream side I haven't seen anything specifically asking users to specify an API server namespace in order to set NetworkPolicies yet - is this something we should adopt everywhere? Or is it possible we don't need this information?

amancse: APIServerNamespace is not needed now, I will be removing this completly in order to address Roke's [comment](https://github.com/open-cluster-management-io/ocm/pull/1634#discussion_r3630685282)  but we still have MonitoringNamespace field for the same purpose as with diff downstream have different namespace for monitoring viz: "openshift-monitoring" (OpenShift), "monitoring" (kube-prometheus-stack)

deploy/cluster-manager/olm-catalog/latest/manifests/operator.open-cluster-management.io_clustermanagers.yaml
```diff
@@ -868,6 +868,29 @@ spec:
                 description: workImagePullSpec represents the desired image configuration
                   of work controller/webhook installed on hub.
                 type: string
+              networkPolicyConfiguration:
```
mikeshng: Thank you for your contribution!

Before I dig deeper into the PR, I want to point out that this API change cannot be made directly in the CRD. The change needs to be introduced in the API definition first, in https://github.com/open-cluster-management-io/api, and then propagated through the generated CRDs via go mod import, vendor import etc.

Most importantly, changes to the OCM core API are not trivial and require careful consideration. Before making any API changes, please create an issue (in the OCM repo, API repo, or enhancement repo) that clearly describes the problem you are trying to solve, the proposed solution, and why an API change is necessary. Seek the API maintainers approval first before the API PR.

/hold

vendor/open-cluster-management.io/api/operator/v1/types_clustermanager.go
mikeshng: +1

This is a significant change. Please see my comment https://github.com/open-cluster-management-io/ocm/pull/1634/changes#r3632256362

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -47,8 +48,24 @@ const (
...
 	clusterManagerReSyncTime = 5 * time.Second
+
+	NetworkPolicies featuregate.Feature = "NetworkPolicies"
 )
 
+var operatorOnlyFeatureGates = map[featuregate.Feature]bool{
+	NetworkPolicies: true,
+}
+
+func filterOperatorFeatureGates(features []operatorapiv1.FeatureGate) []operatorapiv1.FeatureGate {
```
qiujian16: we have defined this in klusterlet, I think we should share the same common util func.

amancse: added NetworkPolicies constant, operatorOnlyFeatureGates map, and FilterOperatorFeatureGates as the single shared implementation,removed local duplicates

deploy/cluster-manager/olm-catalog/latest/manifests/operator.open-cluster-management.io_clustermanagers.yaml
amancse: removed NamespaceSelectors from prometheus NP, thus this API change is not valid as well. I have removed the api change done previously

vendor/open-cluster-management.io/api/operator/v1/types_clustermanager.go
amancse: I have removed the api change done previously for MonitoringNamespace

vendor/open-cluster-management.io/api/operator/v1/types_clustermanager.go
amancse: Addressed

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_hub_reconcile.go
```diff
@@ -147,6 +160,39 @@ func (c *hubReconcile) reconcile(ctx context.Context, cm *operatorapiv1.ClusterM
...
+				return objData, nil
+			},
+			npFiles...,
+		)
+		for _, result := range npResults {
+			if result.Error != nil {
+				appliedErrs = append(appliedErrs, fmt.Errorf("%q (%T): %v", result.File, result.Type, result.Error))
+			}
+		}
+	}
```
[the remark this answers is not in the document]
mikeshng: It was not done in https://github.com/open-cluster-management-io/ocm/pull/1627 so this may not be necessary for now. It can be done in the future, if it becomes a problem.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_hub_reconcile.go
```diff
@@ -147,6 +158,33 @@ func (c *hubReconcile) reconcile(ctx context.Context, cm *operatorapiv1.ClusterM
 		}
 	}
 
+	if config.NetworkPoliciesEnabled {
+		npFiles := append([]string{}, hubNetworkPolicyBaseFiles...)
+		npFiles = append(npFiles, hubPrometheusNPFile)
+		npResults := helpers.ApplyDirectly(
```
mikeshng: How come https://github.com/open-cluster-management-io/ocm/pull/1627 doesn't need to manually call this ApplyDirectly function and you have to?

mikeshng (commented):
Good job removing the API dep. 

/unhold

Added a review feedback. Please address.

Also, since you are hitting the DCO CI error and you will have to fix your commits, might as well squash your commits into one and make sure you sign off on that commit with `git commit -s ...`

qiujian16:
verify fails, I do not think you should have diff on vendor. Also please enable feature gate in integration and e2e test.

manifests/cluster-manager/hub/networkpolicies/01-hub-ns-default-deny.yaml
amancse: Added selected pod names

manifests/cluster-manager/hub/networkpolicies/04-hub-ns-webhook-ingress.yaml
amancse: removed ApiServerNamespace dependency

manifests/cluster-manager/hub/networkpolicies/04-hub-ns-webhook-ingress.yaml
amancse: removed NS dep

manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
amancse: Added selected pod names

manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
amancse: Added selected pod names

manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
amancse: Added selected pod names

manifests/cluster-manager/hub/networkpolicies/03-hub-ns-intra-namespace.yaml
amancse: Added selected pod names

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_hub_reconcile.go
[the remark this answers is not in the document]
amancse: removed for now

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_hub_reconcile.go
amancse: removed

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go
```diff
@@ -195,8 +196,9 @@ func (n *clusterManagerController) sync(ctx context.Context, controllerContext f
 		registrationFeatureGates = clusterManager.Spec.RegistrationConfiguration.FeatureGates
 		config.AutoApproveUsers = strings.Join(clusterManager.Spec.RegistrationConfiguration.AutoApproveUsers, ",")
 	}
+	config.NetworkPoliciesEnabled = helpers.FeatureGateEnabled(registrationFeatureGates, ocmfeature.DefaultHubRegistrationFeatureGates, helpers.NetworkPolicies)
 	config.RegistrationFeatureGates, registrationFeatureMsgs = helpers.ConvertToFeatureGateFlags("Registration",
-		registrationFeatureGates, ocmfeature.DefaultHubRegistrationFeatureGates)
+		helpers.FilterOperatorFeatureGates(registrationFeatureGates), ocmfeature.DefaultHubRegistrationFeatureGates)
```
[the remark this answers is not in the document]
amancse: FeatureGateEnabled checks the explicit feature gates list first — before ever touching the defaults map. So when a user sets NetworkPolicies: Enable in their ClusterManager spec, the function finds it in the
  loop and returns true immediately. It never reaches the defaults map fallback.

  The defaults map only matters when the user sets nothing — and in that case falling through to "key not found → false" is exactly the right behavior for an opt-in gate.

  The code works correctly in all three cases:
  - User enables it → true ✓
  - User disables it explicitly → false ✓
  - User sets nothing → false (opt-in default) ✓

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller_test.go
```diff
@@ -1709,3 +1709,88 @@ func containsArg(args []string, prefix string) bool {
...
+						hasBaseNP = true
+					case "allow-prometheus-ingress":
+						hasPrometheus = true
+					}
+				}
+			}
+
+			if hasBaseNP != c.expectBaseNPs {
+				t.Errorf("expectBaseNPs=%v but hasBaseNP=%v", c.expectBaseNPs, hasBaseNP)
+			}
```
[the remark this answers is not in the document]
amancse: The single hasBaseNP boolean is sufficient here. The reconciler applies all four base policies atomically from the same manifest loop — there is no code path that creates a subset

deploy/cluster-manager/config/rbac/cluster_role.yaml
```diff
@@ -49,6 +49,9 @@ rules:
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["networking.k8s.io"]
+  resources: ["networkpolicies"]
+  verbs: ["create", "get", "list", "watch", "update", "patch", "delete"]
```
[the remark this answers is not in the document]
amancse: Scoping networkpolicies to a namespaced Role while the same ClusterRole already grants cluster-wide namespaces, clusterroles, and clusterrolebindings management would be inconsistent and add maintenance burden (new Role + RoleBinding files, kustomization wiring) for no meaningful security gain.  An operator with namespace-creation rights can already affect NetworkPolicies indirectly. The appropriate scope for this privilege reduction is the ClusterRole as a whole, which is a much larger, pre-existing design decision in this repo, not scoped to this PR.

manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
```diff
@@ -0,0 +1,29 @@
...
+          - "{{ .ClusterManagerName }}-addon-webhook"
+          - "{{ .ClusterManagerName }}-grpc-server"
+  policyTypes:
+    - Ingress
+  ingress:
+    - ports:
+        - port: 8080
+          protocol: TCP
+        - port: 8443
+          protocol: TCP
```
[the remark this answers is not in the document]
amancse: The open ingress on metrics ports is intentional for the initial implementation, as we are not adding NamespaceSelector as this is not a fixed name

deploy/cluster-manager/chart/cluster-manager/templates/cluster_role.yaml
```diff
@@ -81,6 +81,10 @@ rules:
 - apiGroups: ["admissionregistration.k8s.io"]
   resources: ["validatingwebhookconfigurations", "mutatingwebhookconfigurations"]
   verbs: ["create", "get", "list", "update", "watch", "patch", "delete"]
+# Allow the registration-operator to manage NetworkPolicies
+- apiGroups: ["networking.k8s.io"]
+  resources: ["networkpolicies"]
+  verbs: ["create", "get", "list", "update", "watch", "patch", "delete"]
```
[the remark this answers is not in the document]
amancse: Same reasoning as the config/rbac finding: the operator's ClusterRole already grants cluster-wide namespaces, clusterroles, clusterrolebindings, CRD, and webhook configuration management

test/e2e/e2e_suite_test.go
```diff
@@ -170,6 +170,11 @@ var _ = BeforeSuite(func() {
 		return hub.EnableHubRegistrationFeature("ClusterImporter")
 	}).Should(Succeed())
 
+	By("Enable NetworkPolicies Feature")
+	Eventually(func() error {
+		return hub.EnableHubRegistrationFeature("NetworkPolicies")
+	}).Should(Succeed())
```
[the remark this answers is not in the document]
amancse: CheckHubReady() calls CheckClusterManagerStatus() which asserts ObservedGeneration == Generation. The reconciler only advances ObservedGeneration after fully processing the updated spec — including applying all NetworkPolicy resources. If any apply fails, the Applied condition is set to False, which CheckClusterManagerStatus also catches. An explicit NetworkPolicy existence check would be redundant with this existing synchronization.

manifests/cluster-manager/hub/networkpolicies/04-hub-ns-webhook-ingress.yaml
```diff
@@ -0,0 +1,40 @@
+# Ingress policy for all externally-reachable hub pods.
+# Always applied — restricted by pod label and port only.
+# No source namespace filter — traffic origin varies by platform.
```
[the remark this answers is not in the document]
amancse: Previously addressed. The policy covers both webhook ingress (apiserver source varies by platform) and gRPC server ingress (ports 8090/443) from spoke agents on remote clusters. gRPC traffic cannot be restricted with namespaceSelector since spoke agents are not in any hub-cluster namespace. Making the policy opt-in would silently break gRPC connectivity for users who don't configure a webhook source namespace, because policy 01 (default-deny) would block it.

pkg/operator/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller_test.go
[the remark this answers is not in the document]
amancse: Addressed

manifests/klusterlet/management/klusterlet-agent-networkpolicy.yaml
```diff
@@ -14,18 +14,10 @@ spec:
...
-    ports:
+  - ports:
     - protocol: TCP
       port: 443
-  - to:
-    - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: open-cluster-management-hub
-    ports:
+  - ports:
```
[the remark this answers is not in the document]
amancse: namespaceSelector are not being added for any NP, as these are vendor specific ns names

manifests/klusterlet/management/klusterlet-networkpolicy.yaml
```diff
@@ -16,17 +16,6 @@ spec:
...
-    - protocol: TCP
-      port: 443
-  - to:
-    - namespaceSelector:
-        matchLabels:
-          kubernetes.io/metadata.name: default
-    ports:
+  - ports:
     - protocol: TCP
       port: 443
```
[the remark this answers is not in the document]
amancse: namespaceSelector are being removed for NP, this is intentional

pkg/operator/helpers/helpers_test.go
```diff
@@ -2442,6 +2442,105 @@ func TestNormalizeImagePullSecretName(t *testing.T) {
...
+		},
+		{
+			name: "egress policy scoped to OCM pods",
+			config: manifests.HubConfig{
+				ClusterManagerNamespace: "open-cluster-management-hub",
+				ClusterManagerName:      "cluster-manager",
+			},
+			manifestFile: "cluster-manager/hub/networkpolicies/02-hub-ns-egress.yaml",
+			expectContains: []string{
+				"allow-egress", "port: 53", "port: 443", "port: 6443", "port: 5353",
```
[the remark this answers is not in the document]
amancse: addressed

mikeshng:
> verify fails, I do not think you should have diff on vendor. Also please enable feature gate in integration and e2e test.

Addressed by Aman.

Qiu Jian, could you PTAL again? Thanks. 

/assign @qiujian16

manifests/cluster-manager/hub/networkpolicies/05-hub-ns-prometheus.yaml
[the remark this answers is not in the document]
mikeshng: It's already allowed by the allow-hub-ingress rule in 04-hub-ns-webhook-ingress.yaml, so this one would just be a repeat.

mikeshng (approved):
/lgtm

amancse:
@qiujian16 All the requested changes are done, can you please have a look

qiujian16 (commented):
/approve

PR #1639, opened 2026-07-23
:bug: Apply TLS profile to operator and agent serving endpoints

### LINKED ISSUE #1638: :bug: Operator and agent serving endpoints do not apply TLS profile from ocm-tls-profile ConfigMap
**Describe the bug**

The cluster-manager operator (hub), klusterlet operator (spoke), and spoke agent binaries
(klusterlet-agent, registration-agent, work-agent) serve a health/metrics endpoint on port 8443
via library-go's GenericAPIServer. This endpoint does not respect the TLS profile configured in
the ocm-tls-profile ConfigMap -- it continues to accept TLSv1.2 even when the cluster TLS
profile is set to require TLSv1.3.

The spoke agents receive --tls-min-version and --tls-cipher-suites flags from their deployment
templates (injected by the klusterlet operator), but the flags are not wired to the serving
endpoint because ApplyTLSToCommand is not called.

The operators do not receive TLS flags from their deployment manifests at all -- they read the
ConfigMap at runtime, but only after the serving endpoint is already started with library-go
defaults.

All other hub commands (registration-controller, placement, work, addon-manager, grpc-server)
correctly call ApplyTLSToCommand and do not have this issue.

**To Reproduce**

**Expected behavior**

**Additional context**

tesshuflower:
The operator and agent binaries serve health/metrics on port 8443 via library-go's GenericAPIServer but do not apply the TLS profile from the ocm-tls-profile ConfigMap. Port 8443 accepts TLSv1.2 even when the cluster requires TLSv1.3.

- Add `ApplyTLSToCommand` to spoke agent commands (they receive TLS flags from deployment templates but were not wiring them to the serving endpoint)
- Add `ApplyTLSFromConfigMapToCommand` for operator commands (they read the ConfigMap directly since no external component injects TLS flags into their deployments)

Fixes #1638

## Summary by CodeRabbit

* **New Features**
  * Added automatic TLS profile configuration for hub and spoke operators from the cluster TLS profile.
  * Applied TLS settings to spoke agent, registration, and work-agent health and metrics servers.
  * TLS settings are refreshed when components restart after profile updates.

* **Bug Fixes**
  * Preserved explicitly provided TLS configuration and handled unavailable in-cluster configuration gracefully.

* **Tests**
  * Added coverage for TLS flags, profile loading, defaults, and configuration hooks.

### DIFF
--- a/pkg/common/options/options.go
+++ b/pkg/common/options/options.go
@@ -11,6 +11,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/utils/clock"
 
 	tlslib "open-cluster-management.io/sdk-go/pkg/tls"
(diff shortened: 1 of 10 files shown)


pkg/common/options/options.go
```diff
@@ -73,16 +75,42 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
...
+// writeTLSServingConfig writes a minimal GenericOperatorConfig YAML to a temp
+// file in /tmp and sets library-go's --config flag to point at it.
+// /tmp is mounted as an emptyDir in all operator/agent deployments so writing
+// is safe even with readOnlyRootFilesystem: true.
 // Inside cmd.Run, StartController calls Config() which reads the file; the TLS
-// values survive SetRecommendedHTTPServingInfoDefaults because DefaultString only
-// sets fields that are empty.
+// values survive SetRecommendedHTTPServingInfoDefaults because DefaultString
+// only sets fields that are empty.
+func writeTLSServingConfig(cmd *cobra.Command, minVersion, cipherSuites string) error {
```
qiujian16, line 85: why we need to write the configuration from the code? shouldn't we just have a config yaml as a placeholder somewhere? And what if we do not specify this configuration at all?

tesshuflower: @qiujian16 - sorry for the delay, I went on PTO shortly after creating this PR.

This code that writes out the configuration wasn't actually added by this PR, but just extracted/moved from the existing ApplyTLSToCommand() function, so it could be re-used.

In this case we need to load/set the config early to be used by the healthz/metrics so the PR is attempting to re-use what was already there, but called later on.

If the configmap is empty or not set, then it should work as currently, inheriting the default TLS settings from library-go.

PR #1648, opened 2026-08-05
:bug: Restart klusterlet operator when TLS ConfigMap changes

rokej:
## Summary

- Add `StartTLSConfigMapWatcher` to `RunKlusterletOperator`, matching the cluster-manager operator pattern.
- When `ocm-tls-profile` changes, the process exits with `os.Exit(0)` so Kubernetes restarts the pod and the operator can re-apply TLS settings for its own serving endpoint (port 8443).
- Leaves `AddonTLSConfigController` unchanged (it still copies the ConfigMap into addon namespaces).

This complements #1639, which wires TLS into the serving endpoint at startup via ConfigMap/flags. Without a watcher on the klusterlet operator, a ConfigMap update (or a race where the sidecar creates the ConfigMap after process start) would leave the operator's `:8443` endpoint on stale TLS settings.

Related: #1638, #1639

## Test plan

- [x] `go test ./pkg/operator/operators/klusterlet/...`
- [ ] Confirm CI unit/integration/e2e pass
- [ ] On an OpenShift spoke: change `APIServer` TLS profile, verify `ocm-tls-profile` updates and the klusterlet operator pod restarts
- [ ] After restart (and with #1639 applied), verify `:8443` enforces the new min TLS version

## Summary by CodeRabbit

* **New Features**
  * The operator now detects serving TLS configuration changes and restarts automatically so updated security settings take effect.
  * TLS settings, including supported versions and cipher suites, are recorded during startup.

* **Bug Fixes**
  * Improved handling when TLS monitoring is canceled or unavailable during startup.
  * Deployment reconciliation now waits reliably for expected containers, replicas, and startup arguments before completing.

### DIFF
--- a/pkg/operator/operators/klusterlet/options.go
+++ b/pkg/operator/operators/klusterlet/options.go
@@ -18,6 +18,7 @@ import (
 	operatorclient "open-cluster-management.io/api/client/operator/clientset/versioned"
 	operatorinformer "open-cluster-management.io/api/client/operator/informers/externalversions"
 	workclientset "open-cluster-management.io/api/client/work/clientset/versioned"
+	tlslib "open-cluster-management.io/sdk-go/pkg/tls"
 
 	"open-cluster-management.io/ocm/pkg/operator/helpers"
 	"open-cluster-management.io/ocm/pkg/operator/operators/klusterlet/controllers/addonsecretcontroller"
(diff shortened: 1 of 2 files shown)


pkg/operator/operators/klusterlet/options.go
```diff
@@ -109,6 +110,26 @@ func (o *Options) RunKlusterletOperator(ctx context.Context, controllerContext *
...
 	}
 
+	// Load TLS config and start watching for changes. The watcher restarts the process
+	// when the ConfigMap changes so the operator's own serving endpoint (port 8443) can
+	// pick up the updated TLS settings on the next startup. Agent deployments already
+	// re-read the ConfigMap during reconcile via populateTLSConfig.
+	currentTLSConfig, err := tlslib.StartTLSConfigMapWatcher(ctx, kubeClient, helpers.GetOperatorNamespace(),
+		func() {
+			logger.Info("TLS ConfigMap changed, restarting", "component", "klusterlet operator")
+			os.Exit(0)
```
qiujian16, line 120: should we set a cancel for root ctx, and cancel here? It should be a graceful exit.

mikeshng: @rokej if you are still interested in this PR, could you see Qiu Jian's feedback. I think this PR is ready to go in.

PR #1667, opened 2026-08-27
🐛Bump addon-framework to v1.3.1 and surface JobFailed status feedback

stevekuznetsov:
addon-framework v1.3.1 makes addon pre-delete hooks retry when their Pod/Job terminally fails (e.g. pod evicted under MemoryPressure/ DiskPressure, node lost, or Job backoffLimit exhausted). Previously an evicted hook pod stayed Failed forever, the pre-delete hook finalizer was never removed, and the ManagedClusterAddOn (and its owning ManagedCluster) hung on deletion. See ACM-37228.

Add a JobFailed WellKnownStatus feedback rule to the work-agent so the Job 'Failed' condition is reported back on the ManifestWork. The v1.3.1 hook completion logic consumes this to detect a terminally failed hook Job and recreate it.

cc @mikeshng @rokej 

## Summary by CodeRabbit

- **New Features**
  - Added status feedback for detecting failed Kubernetes Jobs.
  - Job failure conditions are now surfaced as a clear failed status.

- **Chores**
  - Updated the add-on framework dependency to version 1.3.1.
  - Updated end-to-end test environments to use Kubernetes version 1.29.2 for more consistent validation.


mikeshng (approved):
/approve

/lgtm

I would prefer we don't add this to wellknownstatus to minimize backward compatibility risk but I don't think it's that big of a deal.

Thank you for the fix!

mikeshng (approved):
/lgtm

Attempting to fix the e2e

PR #1677, opened 2026-09-04
:bug: fix(registration): infer hosted addon mode from install namespace

rokej:
## Summary

Registration-agent decides whether a ManagedClusterAddOn runs outside the managed cluster (hosted mode) primarily from the `addon.open-cluster-management.io/hosting-cluster-name` annotation on the MCA resource. When that annotation is temporarily absent—such as during concurrent updates while many clusters are being imported—the agent treats the addon as running on the spoke. It then uses the spoke API client to create hub-kubeconfig secrets and manage leases on the wrong cluster, which breaks addons that are actually deployed in the hosting cluster's `klusterlet-{cluster}` namespace.

This change adds a per-addon fallback: if the MCA reports an install namespace of `klusterlet-{managedClusterName}`, treat the addon as hosted even when the hosting-cluster-name annotation is missing. ManagedCluster-level annotations alone are not used as a gate, because some addons on a hosted cluster still run on the spoke (for example, addons installed in `open-cluster-management-agent-addon`).

Also wires the ManagedCluster lister into the addon lease controller so both the registration and lease controllers evaluate hosted mode consistently, and adds structured logging when fallback logic is used.

## Test plan

- [x] Unit tests in `pkg/registration/spoke/addon/` pass
- [x] Verify hosted addon registration and lease renewal when MCA `hosting-cluster-name` is absent but install namespace is `klusterlet-{cluster}`
- [x] Verify spoke-deployed addons on hosted clusters are still classified correctly

## Summary by CodeRabbit

- **New Features**
  - Added hosted-mode add-on detection based on installation location and managed-cluster configuration.
  - Add-ons can resolve their hosting cluster from managed-cluster settings.
  - Lease and registration handling now supports hosted add-ons across the appropriate cluster context.

- **Bug Fixes**
  - Improved hosted add-on availability and lease handling when hosting annotations are absent but hosted-mode configuration is present.

### DIFF
--- a/pkg/registration/spoke/addon/configuration.go
+++ b/pkg/registration/spoke/addon/configuration.go
@@ -10,6 +10,8 @@ import (
 	"k8s.io/klog/v2"
 
 	addonv1beta1 "open-cluster-management.io/api/addon/v1beta1"
+	clusterv1listers "open-cluster-management.io/api/client/cluster/listers/cluster/v1"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
 
 	addonwebhook "open-cluster-management.io/ocm/pkg/addon/webhook/v1beta1"
 	"open-cluster-management.io/ocm/pkg/registration/register/csr"
@@ -18,6 +20,13 @@ import (
 
 const (
 	defaultAddOnInstallationNamespace = "open-cluster-management-agent-addon"
+
+	// ManagedCluster annotations that gate hosted-mode addon deployment (aligned with
+	// multicloud-operators-foundation HostedClusterInfo and klusterlet-addon-controller).
+	annotationEnableHostedModeAddons       = "addon.open-cluster-management.io/enable-hosted-mode-addons"
+	annotationKlusterletDeployMode         = "import.open-cluster-management.io/klusterlet-deploy-mode"
+	annotationKlusterletHostingClusterName = "import.open-cluster-management.io/hosting-cluster-name"
+	klusterletDeployModeHosted             = "Hosted"
 )
 
 // registrationConfig contains necessary information for addon registration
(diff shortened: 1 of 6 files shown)


mikeshng (commented):
@rokej could you take a look at the enhancement we just completed and see if there is some overlap or something we can use from the new feature? https://github.com/open-cluster-management-io/enhancements/issues/188 the KEP is here: https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/188-hosted-addon-follow-klusterlet the actual implementation into api, addon-framework and ocm might have slight difference but the general idea is the same. 

I didn't look at the code in details yet but this PR description doesn't make much sense:

> if the MCA reports an install namespace of klusterlet-{managedClusterName}, treat the addon as hosted even when the hosting-cluster-name annotation is missing

How can we make that the default without potentially breaking some users with that name format for the klusterlet? It doesn't feel that unique and it doesn't feel like it implies hosted mode?

kahirokunn:
Wouldn’t removing the annotation normally indicate an intention to move the add-on from hosted mode back to running on the spoke? I understand that switching deployment modes after installation is not currently supported, though.

kahirokunn:
In other words, I don't think this is a bug.

kahirokunn:
https://github.com/open-cluster-management-io/ocm/pull/1677#pullrequestreview-5115552043
It would be great if we could use this feature, wouldn't it?

rokej:
It looks like the feature is about aligning the initial installation of hosted-mode addons with the hosted klusterlet. This is about fixing the following race condition which causes the lease controller of the hosted-mode work-manager addon to talk to incorrect API server during its startup after the installation.

## The Race Condition

When a ManagedCluster is imported in **hosted mode**, within a **700ms window**, **5 concurrent writes** hit the work-manager `ManagedClusterAddOn` object:

| Writer | Operation |
|--------|-----------|
| `addon-manager-controller` | Set `hosting-cluster-name` annotation (the critical write) |
| `ocm-controller` | Patch status (supported configs) |
| `ocm-controller` | Add finalizers |
| `ocm-controller` | Update addon status |
| `ocm-controller` | Permission configuration |

Each write increments `resourceVersion`, so the next writer gets **"the object has been modified"**. The addon-manager-controller's annotation write fails **3 times** and backs off. The `hosting-cluster-name` annotation is **never set** on the work-manager addon.

## The Impact

When the registration agent's `ManagedClusterAddOnLeaseController` starts and reads the work-manager addon:

1. `isAddonRunningOutsideManagedCluster()` (`pkg/registration/spoke/addon/configuration.go#L81`) checks for the `hosting-cluster-name` annotation → **missing** → returns `false`
2. The controller defaults to `spokeLeaseClient` (`pkg/registration/spoke/addon/lease_controller.go#L127-L129`) instead of `managementLeaseClient`
3. It calls `spokeLeaseClient.Leases("klusterlet-<id>").Get()` on the **hosted cluster's** kube-apiserver
4. The `klusterlet-<id>` namespace only exists on the **management cluster** → `namespaces 'klusterlet-...' not found`
5. Exponential backoff (1s → 2s → 5s → 10s → 20s → 40s → 80s...) → **never recovers**
6. `work-manager-hub-kubeconfig` secret is never saved → `klusterlet-addon-workmgr` pod stuck on `FailedMount` → no cluster claims synced → **cluster install times out**

rokej:
This PR implements a fallback mechanism of checking whether the addon is running in the hosted mode or not when it starts up during the race condition. An addon is installed in klusterlet-{abc} only when the managedclusteraddon is configured to run in hosted-mode so I think this fallback is safe assumption.

mikeshng (commented):
@rokej alot of assumptions in this PR is made using downstream ACM code. In upstream OCM-io there is no work manager addon (work manager is not manifestwork), there is no ocm-controller, there is no multicloud-operators-foundation and your assumption of 

> An addon is installed in klusterlet-{abc} only when the managedclusteraddon is configured to run in hosted-mode

is only true in downstream ACM. Please double check.

rokej:
Closing in favor of #1679.

#1679 addresses the root cause by deriving `addon.open-cluster-management.io/hosting-cluster-name` on ManagedClusterAddOn create from import annotations, so the registration agent does not depend on a later annotation sync that can lose optimistic-lock races with addon framework controllers.

The spoke-side install-namespace fallback in this PR is no longer needed for that scenario.

PR #1679, opened 2026-09-04
:bug: derive hosting-cluster-name on MCA create for hosted clusters

rokej:
## Summary

- Derive `addon.open-cluster-management.io/hosting-cluster-name` on **ManagedClusterAddOn create** from import annotations when hosted-mode addons are enabled, instead of depending on a later annotation sync.
- Add `pkg/addon/hosted` with shared derivation logic and unit tests; wire it into the addon-management install reconciler.

## Problem

When a ManagedCluster is imported in hosted mode, several controllers update the same `ManagedClusterAddOn` object almost immediately after it is created. In a tight window (~700ms), concurrent writes can hit a single addon (for example `work-manager`):

| Writer | Operation |
|--------|-----------|
| addon-annotation-controller | Set `hosting-cluster-name` annotation (critical for hosted mode) |
| addon deploy controller | Patch status (supported configs) |
| addon deploy controller | Add finalizers |
| addon deploy controller | Update addon status |
| addon deploy controller | Permission configuration |

Each write bumps `resourceVersion`, so the next writer often gets `the object has been modified`. The annotation controller’s `hosting-cluster-name` write can lose after repeated conflicts and back off, leaving the annotation unset on the MCA.

### Impact

When the registration agent’s `ManagedClusterAddOnLeaseController` starts and reads that addon:

1. `isAddonRunningOutsideManagedCluster()` checks for `addon.open-cluster-management.io/hosting-cluster-name` → missing → returns `false`
2. The lease controller uses `spokeLeaseClient` instead of `managementLeaseClient`
3. It calls `spokeLeaseClient.Leases("klusterlet-<id>").Get()` on the **hosted cluster** API server
4. The `klusterlet-<id>` namespace exists only on the **management** cluster → `namespaces "klusterlet-..." not found`
5. Exponential backoff never recovers; hub kubeconfig secrets are not written, addon pods stay in `FailedMount`, and cluster bring-up can stall

Today, hosted mode is inferred from an annotation that may arrive **after** create via `addon-annotation-controller`, which races with addon framework controllers that patch the same MCA.

## Fix

At **MCA create** time, the addon-management install reconciler already copies `addon.open-cluster-management.io/*` annotations from the ManagedCluster. This change extends that path to **derive** `hosting-cluster-name` from import annotations when:

- `addon.open-cluster-management.io/enable-hosted-mode-addons` is `"true"`
- `import.open-cluster-management.io/klusterlet-deploy-mode` is `Hosted` or `SingletonHosted`
- `import.open-cluster-management.io/hosting-cluster-name` is set

Those import annotations are typically present on the ManagedCluster before placement-installed MCAs are created, so the hosting annotation is on the MCA **from the first write** and the lease controller does not need to wait for a follow-up annotation sync.

Scope: placement-installed MCAs on clusters that meet the hosted-addon criteria at create time (not all MCAs fleet-wide). Existing MCAs are unchanged unless recreated or updated by other controllers.

## Test plan

- [x] `go test ./pkg/addon/hosted/... ./pkg/addon/controllers/addonmanagement/...`
- [x] `make lint`
- [x] Manual: import hosted cluster, confirm new MCAs have `hosting-cluster-name` at create and addon lease controller uses management-cluster client

## Summary by CodeRabbit

* **Bug Fixes**
  * Add-ons now correctly inherit hosted-mode settings from their managed cluster.
  * Hosted add-ons now receive the appropriate hosting cluster information when created.
  * Add-on metadata is more accurately propagated, including hosted manifest location and hosted-mode enablement settings.
* **Tests**
  * Added coverage for hosted and singleton-hosted deployment modes, missing configuration, and unrelated annotations to ensure consistent behavior.


pkg/addon/controllers/addonmanagement/addon_install_reconciler.go
```diff
@@ -99,22 +99,16 @@ func (d *managedClusterAddonInstallReconciler) reconcile(
...
 	}
 
-	addonAnnotations := map[string]string{}
-	for k, v := range cluster.Annotations {
-		if strings.HasPrefix(k, addonv1beta1.GroupName) {
-			addonAnnotations[k] = v
-		}
-	}
-	return addonAnnotations, nil
+	return hosted.AddonAnnotationsFromManagedCluster(cluster), nil
```
mikeshng: nit: why `hosted.` ? addon annotations should also be copied when its not hosted. I think the function does it properly but just all packed under the hosted package for some reason. How about you leave the original code and add the hosted one then return like before?

mikeshng (commented):
This fix looks a lot better and much easier to understand. Could you please check if there is existing hosted addon integration or e2e tests that you can easily modify to check the new code is working properly? If there is no existing tests then just leave it. Thanks.

pkg/addon/controllers/addonmanagement/addon_install_reconciler.go
rokej: Yes, you are right. We don't really need a separate package for hosted.

rokej:
> Could you please check if there is existing hosted addon integration or e2e tests that you can easily modify to check the new code is working properly? If there is no existing tests then just leave it.

A good suggestion. It would be nice to have the full hosted-addon stack scenario test but the deviation tests are already covered by `annotations_test.go` and `controller_test.go` and the existing integration/e2e tests cover addon-annotation sync when the MC already has the addon key. A dedicated integration case isn’t a trivial edit of current tests so I would just leave things as is.

mikeshng (approved):
/approve

/lgtm

Thanks for double checking!

PR #1684, opened 2026-09-08
:bug: feat: use hash-based owner label for ManifestWorkReplicaSet

### LINKED ISSUE #1596: ManifestWorkReplicaSet silently fails when `namespace.name` exceeds the 63-char label limit

payjain:
## Summary

Closes #1596 (Request 1).

The MWRS controller stamps every generated `ManifestWork` with a label whose value is `namespace.name`. When that exceeds the 63-char Kubernetes label-value limit, the ManifestWork create is silently rejected.

- Introduce a new label `work.open-cluster-management.io/ownerkey-hash` with a deterministic SHA-256 hash (32 hex chars, always a valid label value)
- Add an annotation `work.open-cluster-management.io/manifestworkreplicaset-owner` with the readable `namespace/name` for traceability
- Deprecate the old `work.open-cluster-management.io/manifestworkreplicaset` label (still set when value fits ≤63 chars, for backward compatibility)
- Selection and event filtering check both labels so pre-upgrade ManifestWorks are found via a fallback selector and get the new labels on their next reconcile cycle
- GC controller checks both label keys when deciding to skip MWRS-managed ManifestWorks

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./pkg/work/...` passes
- [x] Unit tests pass for `manifestworkreplicasetcontroller` (34 tests including new ones)
- [x] Unit tests pass for `manifestworkgarbagecollection` (8 tests including new hash-label skip test)
- [x] `TestDeployReconcileLongOwnerNameUsesHashLabel` — verifies long namespace.name now succeeds
- [x] `TestBuildManifestWorkLabelsAndAnnotations` — verifies hash label, annotation, and conditional old label
- [ ] Integration tests (existing selectors still work since old label is set for short names)
- [ ] E2E tests

## Summary by CodeRabbit

* **New Features**
  * ManifestWork ReplicaSets support owner names longer than label limits through hashed ownership metadata.
  * Owner annotations preserve readable namespace and name information.

* **Bug Fixes**
  * Existing ManifestWorks continue to be recognized during ownership transitions, including legacy metadata.
  * Ownership detection correctly handles names containing dots.
  * Reconciliation no longer fails for long owner names.

* **Behavior Updates**
  * Garbage collection continues to preserve ReplicaSet-managed ManifestWorks.
  * Ownership and listing issues are logged for improved visibility.


payjain:
The `integration` check failure is unrelated to this PR — it's the same flaky klusterlet operator test (`Klusterlet — Deployment should be updated when klusterlet is changed` at `klusterlet_test.go:624`) that expects 8 args but gets 7. This has failed consistently across retries and also appears on other PRs (e.g. `fix-mca-hosted-annotation-at-create` branch).

All checks relevant to this PR's MWRS label changes pass: `build`, `unit`, `verify`, `e2e`, `e2e-hosted`, `e2e-singleton`, `DCO`, `verify PR contents`.

Could a maintainer re-run or bypass the flaky integration check? cc @qiujian16

payjain:
## Manual validation on kind clusters

**Setup:** Two kind clusters (hub + 1 managed cluster) with OCM installed via `clusteradm`. ManifestWorkReplicaSet feature gate enabled. Work controller image built from `feat/mwrs-hash-label` branch (commit `a717a3c`).

**Reproducing the bug (upstream image):**
- Created MWRS `my-long-replicaset-name` in namespace `this-is-a-very-long-namespace-name-for-testing` — combined label value is 70 chars (exceeds K8s 63-char limit)
- Result: No ManifestWork created, MWRS status shows `InvalidManifestWorkName`

**Verifying the fix (PR image):**

| Test | Result |
|------|--------|
| Long-name MWRS (70-char label value) creates ManifestWork | **PASS** |
| ManifestWork has `ownerkey-hash` label (32-char SHA-256 hex) | **PASS** |
| Old `manifestworkreplicaset` label is omitted (value > 63 chars) | **PASS** |
| Readable annotation `manifestworkreplicaset-owner: namespace/name` present | **PASS** |
| Short-name MWRS (`short-ns/short`) has both old and new labels | **PASS** |
| MWRS status correctly reports `total: 1, applied: 1, available: 1` | **PASS** |
| Deleting MWRS cleans up its ManifestWorks | **PASS** |

All scenarios pass. The hash-based label resolves the >63-char failure while maintaining backward compatibility for short names.  

During the validation, I did reproduce the original bug. When the work controller was still running the upstream v1.3.1 image (before we fixed the image loading), created the same long-name MWRS and got exactly the failure described in issue #1596:

status:
  conditions:
  - message: "ManifestWork owner reference \"this-is-a-very-long-namespace-name-for-testing.my-long-replicaset-name\"
      (namespace.name) is not a valid label value: must be no more than 63 bytes..."
    reason: InvalidManifestWorkName
    status: "False"
    type: ManifestworkApplied

And kubectl get manifestworks -n cluster1 returned No resources found — no ManifestWork was created.

After switching to our PR image (a717a3c), the same MWRS successfully created the ManifestWork with the hash label. So both sides of the before/after were validated on the same cluster.

mikeshng (commented):
@payjain this is a great PR and thank you for your contribution!

However, there is a significant code review feedback action item from coderabbitai `Authorization Bypass (CWE-284) Reachability: External · Exploitability: Moderate`. Anytime we see security/exploit/vulnerability feedback like this, it needs to be carefully handled. PTAL thank you!

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_deploy_reconcile.go
```diff
@@ -380,22 +363,31 @@ func buildManifestWork(
 	clusterNS string,
 	placementRefName string,
 ) *workv1.ManifestWork {
-	// Get ManifestWorkReplicaSet labels
-	labels := mwrSet.Labels
-
 	// Merge mwrSet.Labels with the required labels
 	mergedLabels := make(map[string]string)
-	for k, v := range labels {
+	for k, v := range mwrSet.Labels {
```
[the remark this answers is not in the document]
payjain: Fixed in the latest push (commit `e10b120`).

**What was done:**
1. `buildManifestWork` now explicitly `delete()`s the deprecated `ManifestWorkReplicaSetControllerNameLabelKey` from merged labels when the owner key exceeds 63 chars — even if it was injected via user-supplied MWRS labels. This prevents a short-name MWRS from claiming ownership through the deprecated-label fallback.
2. Test helpers (`CreateTestManifestWorks` / `CreateTestManifestWork`) now conditionally set the deprecated label only when `namespace.name` <= 63 chars, matching production behavior.
3. Added regression test `TestBuildManifestWorkStripsForgedDeprecatedLabel` that sets a forged deprecated label on the MWRS and asserts it's stripped from the generated ManifestWork.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
```diff
@@ -152,13 +186,27 @@ func NewManifestWorkReplicaSetController(
...
 		WithInformersQueueKeysFunc(controller.placementDecisionQueueKeysFunc, placeDecisionInformer.Informer()).
 		WithInformersQueueKeysFunc(controller.placementQueueKeysFunc, placementInformer.Informer()).
 		WithSync(controller.sync).ToController("ManifestWorkReplicaSetController")
 }
 
+// filterByMWRSOwnership returns true if the object has either the new hash
+// label or the deprecated old label, indicating it is owned by an MWRS.
+func filterByMWRSOwnership(obj interface{}) bool {
+	accessor, _ := meta.Accessor(obj)
+	lbls := accessor.GetLabels()
```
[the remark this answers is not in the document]
payjain: Fixed in the latest push (commit `e10b120`).

`filterByMWRSOwnership` now unwraps `cache.DeletedFinalStateUnknown` tombstones before calling `meta.Accessor`. Added two regression tests: one with a tombstone wrapping a hash-labeled ManifestWork (should return true), and one with a tombstone wrapping a non-MWRS object (should return false).

payjain:
@mikeshng Thank you for the review and flagging the security finding!

All three CodeRabbit findings have been addressed in the latest push (commit `e10b120`):

1. **CWE-284 Authorization Bypass (Security)**: `buildManifestWork` now explicitly strips the deprecated `ManifestWorkReplicaSetControllerNameLabelKey` from merged labels when the owner key exceeds 63 chars — even if it was injected via user-supplied MWRS labels. This prevents a short-name MWRS from claiming ownership through the deprecated-label fallback. Added regression test `TestBuildManifestWorkStripsForgedDeprecatedLabel`.

2. **Tombstone panic (Stability)**: `filterByMWRSOwnership` now unwraps `cache.DeletedFinalStateUnknown` tombstones before calling `meta.Accessor`. Added two tombstone regression tests.

3. **gofmt formatting (Maintainability)**: Already fixed in the previous push.

Test helpers (`CreateTestManifestWorks` / `CreateTestManifestWork`) also updated to conditionally set the deprecated label only when `namespace.name` <= 63 chars, matching production behavior.

All unit tests pass locally.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
```diff
@@ -59,19 +61,47 @@ type manifestWorkWithPlacements struct {
...
+				}
+				for _, mw := range oldMWs {
+					if !seen.Has(mw.UID) {
+						mws = append(mws, mw)
+					}
+				}
+			}
+		}
+	}
+
```
weirdwiz, line 108: instead of fetching old mwrs, wouldn't it be better to add the new label during reconcile so the get function remains straightforward? 
wdyt?

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
```diff
@@ -142,23 +172,52 @@ func NewManifestWorkReplicaSetController(
...
+			// Fall back to the deprecated label for pre-upgrade ManifestWorks.
 			labelValue, ok := accessor.GetLabels()[workapiv1alpha1.ManifestWorkReplicaSetControllerNameLabelKey]
 			if !ok {
 				return []string{}
 			}
-			keys := strings.Split(labelValue, ".")
-			if len(keys) != 2 {
+			parts := strings.SplitN(labelValue, ".", 2)
+			if len(parts) != 2 {
```
weirdwiz, line 193: nit, original version should work fine because we're splitting into 2 parts anyway

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
```diff
@@ -142,23 +172,52 @@ func NewManifestWorkReplicaSetController(
...
+	lbls := accessor.GetLabels()
+	if len(lbls) == 0 {
+		return false
+	}
+	if len(lbls[workhelper.ManifestWorkReplicaSetOwnerKeyHashLabelKey]) > 0 {
+		return true
+	}
+	return len(lbls[workapiv1alpha1.ManifestWorkReplicaSetControllerNameLabelKey]) > 0
+}
+
```
weirdwiz, line 224: imo we don't need this function, we can probably use `queue.FilterByLabel` but with the new label

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
```diff
@@ -59,19 +61,47 @@ type manifestWorkWithPlacements struct {
...
+			if listErr != nil {
+				klog.V(4).Infof("failed to list ManifestWorks by deprecated label for MWRS %s/%s: %v", mwrs.Namespace, mwrs.Name, listErr)
+			} else {
+				seen := sets.New[types.UID]()
+				for _, mw := range mws {
+					seen.Insert(mw.UID)
+				}
+				for _, mw := range oldMWs {
+					if !seen.Has(mw.UID) {
+						mws = append(mws, mw)
```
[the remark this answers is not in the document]
payjain: Fixed in commit `c475c28`. The deprecated-label fallback now skips ManifestWorks that already have `ManifestWorkReplicaSetOwnerKeyHashLabelKey`. Only truly legacy ManifestWorks (pre-upgrade, no hash label) are merged into the result set. This prevents cross-ownership where a MW with hash pointing to MWRS-B could be claimed by MWRS-A through the deprecated-label fallback.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
payjain: Good suggestion! The dual-query approach was chosen deliberately over backfilling during reconcile for a few reasons:

1. **Read-only migration path** — the deprecated-label query is read-only and self-healing. Backfilling would require a patch/update on every pre-upgrade ManifestWork, adding write load during the upgrade window.
2. **Atomicity** — if the controller restarts mid-backfill, some MWs would have the new label and some would not. The dual-query approach handles partial states gracefully.
3. **Scope** — this function only runs when the old label value fits 63 chars (most cases). For long names, there are no pre-upgrade MWs to find since they were never created (the original bug).

The deprecated-label path will naturally stop matching once all pre-upgrade MWs are reconciled and stamped with the hash label. It could be removed in a future release.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
payjain: The change from `strings.Split` to `strings.SplitN` is intentional. With `strings.Split("my.ns.mwrs", ".")` you get `["my", "ns", "mwrs"]` (3 parts) — the `len(parts) != 2` check rejects it, so dotted namespaces silently fail to queue. With `strings.SplitN("my.ns.mwrs", ".", 2)` you get `["my", "ns.mwrs"]` (2 parts) — correctly splitting on the first dot only, matching the `namespace.name` convention.

This is a pre-existing bug fix (dotted namespaces like `my.ns` would never trigger reconciliation with the old `Split`). There is a test case `"old label with dotted namespace handled by SplitN"` in `TestManifestWorkQueueKeyFuncWithDottedNames` that validates this.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
payjain: `queue.FileterByLabel` (the existing utility) only checks for the existence of a single label key. Our `filterByMWRSOwnership` needs to check for **either** the new hash label **or** the deprecated old label — a logical OR that `FileterByLabel` cannot express. Additionally, `filterByMWRSOwnership` handles `cache.DeletedFinalStateUnknown` tombstone unwrapping, which `FileterByLabel` does not (the existing `FileterByLabel` would panic on tombstones).

For backward compatibility during the migration window, we need both label checks in a single predicate.

pkg/work/hub/controllers/manifestworkreplicasetcontroller/manifestworkreplicaset_controller.go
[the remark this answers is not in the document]
payjain: One additional note: this scenario is **theoretical only**. In practice, `buildManifestWork` always stamps both the deprecated label and the hash label from the same MWRS source — so a ManifestWork can never organically end up with a deprecated label pointing to MWRS-A and a hash label pointing to MWRS-B. That conflicting state would require manual label tampering via kubectl. The `seen` UID dedup already handles the normal case (same MW found by both queries). This hash-owner guard is purely defense-in-depth — nothing would break without it, but it makes ownership intent explicit for 3 extra lines.

mikeshng (commented):
I think the PR looks ok overall. 

Just a question maybe I missed it. The old objects are not being migrated right now. This is something you decided explicitly and not a miss right?

payjain:
:)

payjain:
> I think the PR looks ok overall.
> 
> Just a question maybe I missed it. The old objects are not being migrated right now. This is something you decided explicitly and not a miss right?

@mikeshng Great question — yes, this is deliberate.

Rather than actively backfilling the new label onto existing ManifestWorks, the controller queries by both labels (dual-query). Pre-upgrade MWs get the new hash label naturally on their next reconcile cycle, so migration happens organically with no extra write load or partial-state risk on restart.

The deprecated-label fallback also excludes MWs that already have the hash label (commit `c475c28`) to prevent cross-ownership during the transition. Once all MWs have been reconciled, the deprecated path can be removed in a future release.

mikeshng (commented):
@payjain could you please test again with your latest changes this specific regression test case: that the old objects with the old label will migrate over to the new label (keeping the old one) after a reconciliation? 

@annelaucg could you or your team please sign off on this change? Thank you!

pkg/work/hub/test/helper.go
```diff
@@ -14,6 +14,7 @@ import (
 	workapiv1alpha1 "open-cluster-management.io/api/work/v1alpha1"
 
 	testingcommon "open-cluster-management.io/ocm/pkg/common/testing"
+	workhelper "open-cluster-management.io/ocm/pkg/work/helper"
```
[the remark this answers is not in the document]
payjain: The import block follows the repo's 4-group gci convention (stdlib, k8s.io, open-cluster-management.io/api, open-cluster-management.io/ocm). The workhelper import is in the correct local group alongside existing testingcommon and spoketesting imports. No formatting change needed.

PR #1701, opened 2026-09-18
build: switch registration/work image base to ubi-minimal + install zlib

### LINKED ISSUE #1698: registration image: ubi9/ubi-micro base lacks libz.so.1 — bundled AWS CLI crashes on startup
---

## Summary

The OCM `registration` container image is built on
`registry.access.redhat.com/ubi9/ubi-micro`. The `awsirsa` registration
driver's startup path invokes the bundled AWS CLI to fetch IRSA
credentials, and that CLI dynamically links against `libz.so.1`. `ubi-micro`
does not ship `libz.so.1`. The container crashes before any registration
logic runs, with a linker error like:

```
error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

## Impact

Any Klusterlet configured for `authType: awsirsa` cannot bootstrap. Symptom
looks like the pod restarting in a tight loop with `CrashLoopBackOff`;
`kubectl logs --previous` shows the linker error.

## Suggested fix

Switch the registration image's base to `ubi9/ubi-minimal` and explicitly
install `zlib`:

```diff
# Dockerfile
- FROM registry.access.redhat.com/ubi9/ubi-micro
+ FROM registry.access.redhat.com/ubi9/ubi-minimal
+ RUN microdnf install -y zlib && microdnf clean all
```

`ubi-minimal` is only marginally larger than `ubi-micro` (~35 MB vs ~15 MB
compressed) and comes with `microdnf` for on-the-fly package install.

## PoC

Verified in the `ocm-bcfips/registration-operator:v0.16.0-govcloud-patch2`
build ( internal) — the base image was switched to `ubi-minimal`
with `zlib` installed. `awsirsa` bootstrap completed successfully on

## Reproduction

```yaml
apiVersion: operator.open-cluster-management.io/v1
kind: Klusterlet
metadata:
  name: klusterlet
spec:
  registrationConfiguration:
    registrationDriver:
      authType: awsirsa
```

Deploy on any EKS cluster; the klusterlet-agent pod immediately
CrashLoopBackOffs. `kubectl logs -p` on the pod shows the linker error.

## Alternative fix

If `ubi-micro` is preferred (smallest base), an alternative is to statically
link the AWS CLI dependency chain in the OCM build. That's a larger change
and requires vendoring the AWS CLI at build time.

## Reporter contact

Daryl Smith

## Related

- Companion issues: CRD regex (api), Go partition ARN (ocm), clusteradm resourceVersion (clusteradm)

darylsmit-rgb:
## Problem

The `awsirsa` registration driver in `pkg/registration/register/aws_irsa/aws_irsa.go` builds a `clientcmdapi.ExecConfig` that spawns the AWS CLI as a Kubernetes credential plugin:

```go
Exec: &clientcmdapi.ExecConfig{
    APIVersion: "client.authentication.k8s.io/v1beta1",
    Command:    "/awscli/dist/aws",
    Args:       []string{"--region", awsRegion, "eks", "get-token", ...},
}
```

The AWS CLI is delivered by an initContainer that copies the binaries out of `amazon/aws-cli:latest` into a shared volume mounted at `/awscli` — see `manifests/klusterlet/management/klusterlet-registration-deployment.yaml`, `klusterlet-work-deployment.yaml`, `klusterlet-agent-deployment.yaml`.

The CLI then runs from the **main** container's chroot, so it needs the main container's dynamic libraries.

`registry.access.redhat.com/ubi9/ubi-micro:latest` ships without `libz.so.1`, which the AWS CLI dynamically links against. Result:

```
error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

The klusterlet's registration-agent surfaces this as `CrashLoopBackOff` before any awsirsa logic runs. Any Klusterlet with `authType: awsirsa` cannot bootstrap on the current image.

## Fix

Switch the two Dockerfiles that mount `/awscli` to `ubi9/ubi-minimal` and install `zlib`:

- `build/Dockerfile.registration` — main + singleton agent
- `build/Dockerfile.work`         — manifestwork agent

`ubi-minimal` is only marginally larger than `ubi-micro` (~35 MB vs ~15 MB compressed) and comes with `microdnf` for package install.

The three Dockerfiles whose main containers do NOT mount `/awscli` (`Dockerfile.addon`, `Dockerfile.placement`, `Dockerfile.registration-operator`) are left on `ubi-micro`.

## Alternative considered

If the smallest possible base is preferred, the AWS CLI could be statically linked or the shared library could be baked into the initContainer's payload. Both are larger changes and out of scope for this fix.

## Test plan

- Manual verification of a klusterlet with `authType: awsirsa` bootstrapping successfully on an EKS cluster in the `aws-us-gov` partition. The rebuilt `registration` and `work` images ran the AWS CLI exec plugin without the `libz.so.1` crash.
- Existing image build CI still passes (only the base image and one `RUN` line changed).

## AI disclosure

Per [CONTRIBUTING.md § AI Policy and Guidance](CONTRIBUTING.md#ai-policy-and-guidance): AI-assisted tooling helped draft the initial commit message and this PR description. The technical analysis (libz dependency, which Dockerfiles mount /awscli, the ubi-minimal + microdnf choice) was authored and verified by me on live `aws-us-gov` EKS, and I have reviewed every line of the diff. I will respond to review feedback directly (no AI-drafted responses).

Fixes #1698

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed registration runtime startup for AWS IAM Roles for Service Accounts by including the required compression library.
  * Updated the runtime environment to ensure the AWS CLI credential plugin can launch successfully.


darylsmit-rgb:
@coderabbitai review

mikeshng (commented):
@rokej @dhaiducek @tesshuflower does this have any downstream impact? It's looks ok upstream side as long as all the CI runs.

tesshuflower:
> @rokej @dhaiducek @tesshuflower does this have any downstream impact? It's looks ok upstream side as long as all the CI runs.

I've checked and we're actually using a ubi-minimal image already for our downstream copy.

However, I would prefer not to add more packages on top of the base - doing this may require downstream changes for us - and potentially future churn with updates (we need to do specific rpm packaging for anything outside the base image and then would need to maintain them going forward).

@darylsmit-rgb could you check if this part is really necessary after switching to the ubi-minimal image?

https://github.com/open-cluster-management-io/ocm/pull/1701/changes#diff-5031f6764c8b53661125985a754948977ed7bdde9e2eeb95be5d75a1e89241c5R23

From my quick check it looks to me like zlib should be available by default.

darylsmit-rgb:
@tesshuflower — you're right, thanks for the catch. Verified empirically:

```
$ crane export registry.access.redhat.com/ubi9/ubi-minimal:latest - | tar -tf - | grep -E "libz\.so"
usr/lib64/libz.so.1
usr/lib64/libz.so.1.2.11
```

tesshuflower:
@mikeshng given the latest updates, I'm good with the changes in this PR (moving to ubi-minimal from ubi-micro for the 2 images).

mikeshng (approved):
/approve

/lgtm

Thank you for the fix and Tesshu for the great suggestion!

FYI @suvaanshkumar

